PR #29, opened 2020-07-08
Remove standalone suffix from the documentation.

ph:
need to go in 7.9 7.x

### DIFF
--- a/docs/en/ingest-management/getting-started.asciidoc
+++ b/docs/en/ingest-management/getting-started.asciidoc
@@ -219,7 +219,7 @@ To configure {agent} manually:
 Agent configurations list, click the default config.
 
 . Select the **YAML** tab to see the configuration for {agent}. Copy the
-content and put it into a file named `elastic-agent-standalone.yml` on the
+content and put it into a file named `elastic-agent.yml` on the
 system where {agent} is installed.
 +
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/getting-started.asciidoc
```diff
@@ -255,7 +255,7 @@ datasources:
 +
 [source,shell]
 ----
-./elastic-agent -c elastic-agent-standalone.yml run
+./elastic-agent -c elastic-agent.yml run
```
skh, line 258: In this case I believe `./elastic-agent run` is sufficient, as it uses `elastic-agent.yml` by default?

I'm fine with the change. That said, the reason to use a different filename here was to prevent the configuration used for standalone mode to be overwritten when the user later switches to fleet mode.

I know there's a warning that the config will be overwritten, so either way is fine with me.

ph: Yes, we also want to reduce the friction and concepts.

ph:
updated from review

ph:
@bmorelli25 how backport are handled in this repository?

bmorelli25:
@ph The backport tool is the easiest way. I'll take care of it for this PR.

PR #49, opened 2020-07-16
Move content about how to stop Elastic Agent

dedemorton:
Moves content about stopping Elastic Agent to a more prominent location that's easier to discover through a search.

Requires https://github.com/elastic/beats/pull/19971.

(Build won't pass until after 19971 is merged.)


dedemorton:
Content changes were approved in elastic/beats#19971 so I'm merging without review.

PR #54, opened 2020-07-17
Template for tutorial-type guides

EamonnTP:
This PR adds a template for guidance when creating tutorial-type guides within the Observability documentation. 

For the time being, I have placed the file outside of the docs directory. 

### Related issue

https://github.com/elastic/observability-dev/issues/664

### DIFF
--- a/tutorial-template.asciidoc
+++ b/tutorial-template.asciidoc
@@ -4,8 +4,10 @@ your tutorial. You can use this asciidoc template as a framework, replacing the 
 
 // The title of your tutorial should focus on what the reader will achieve.
 // For example, How to ingest custom data into Elasticsearch.
-// The anchor id name, in this case [[tutorial-title]], must be unique and can only be used once in a doc. You can link to any ID within a document. For example,
-// if you need to link to this section from elsewhere, you would use <<tutorial-title>> as the reference. 
+// The anchor id name, in this case [[tutorial-title]], must be unique and can only be used once in a doc. 
+// The anchor id name should be the same as the file name.
+// You can link to any ID within a document. For example, if you need to link to this section from elsewhere,
+// you would use <<tutorial-title>> as the reference. 
 [[tutorial-title]]
 = Tutorial title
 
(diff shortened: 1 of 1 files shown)


tutorial-template.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+
+[source,shell]
+----
+./elastic-agent -c elastic-agent-standalone.yml run
+----
+
+To include a sample file, specify the source format. For example:
+
+[source,json]
+{
```
dedemorton, line 54: did you mean to include a file here?

tutorial-template.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+    "sample_id": ENT,
+    "sample_name": "String",
+    "sample_number": INT,
+    "line_number": "String",
+    "speaker": "String",
+    "text_entry": "String",
+}
+
+To include an image, save it in a folder in the Observability-docs repo, and
```
dedemorton, line 63: I think we should proscribe a folder called `images`.

dedemorton: I'd keep this a bit more generic, too, and leave out the repo name because stuff tends to bounce around sometimes.

tutorial-template.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+include it using a path relative to the document where the `image::` statement appears. For example,
```
dedemorton, line 64: I would say "by using an `image::` statement" (then show the example).

We need to revisit how we specify the image path to make it possible to move things around more easily, but that's somewhat more complicated in asdciidoc than it probably should be. :-) Probably best right now to be less precise.

tutorial-template.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+include it using a path relative to the document where the `image::` statement appears. For example, 
+this image is saved in the `images` directory:
+
+[role="screenshot"]
+image::images/my-image.png[Alt text]
+
+
+[float]
+[[the-next-step]]
+== Step 2: The next step
```
dedemorton, line 73: @bmorelli25 It would be neat to have some special formatting for these step headings to make them pop in the tutorial docs.

bmorelli25: Interesting. It would indeed. I can't think of an easy way to do this off of the top of my head. I'll add it to my asciidoctor extension project list and dig into it later.

tutorial-template.asciidoc
EamonnTP: Apologies. I meant `To display a sample file`...

tutorial-template.asciidoc
dedemorton: cool, I'll mark it as resolved for now then!

PR #68, opened 2020-07-28
[docs] Add troubleshooting topic about running under SYSTEM for Endpoint

dedemorton:
Adds a troubleshooting topic for users who try to run endpoint under a user other than system.

This is a follow-up to https://github.com/elastic/beats/pull/20172

### DIFF
--- a/docs/en/ingest-management/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting.asciidoc
@@ -22,6 +22,7 @@ Contact us in the {im-forum}[discuss forum]. Your feedback is very valuable to u
 * <<ingest-manager-app-crashes>>
 * <<agent-enrollment-timeout>>
 * <<es-apikey-failed>>
+* <<process-not-root>>
 * <<agent-hangs-while-unenrolling>>
 
 **Frequently asked questions:**
@@ -155,6 +156,33 @@ property in the `kibana.yml` configuration file. For example:
 ----
 xpack.encryptedSavedObjects.encryptionKey: "something_at_least_32_characters"
 ----
+
+[discrete]
+[[process-not-root]]
+== {agent} fails on Windows with `Agent process is not root/admin or validation failed` message
+
+Make sure the user has administrator-level privileges.
+
+If you're using the {elastic-endpoint} integration, also make sure you're
+running {agent} under the SYSTEM account.
+
+To run {agent} under the SYSTEM account, you can:
+
+. Download https://docs.microsoft.com/en-us/sysinternals/downloads/psexec[PsExec]
+and extract the contents to a folder, for example, `d:\tools`.
+. Open a command prompt as an Administrator (right-click the Command Prompt
+icon and select *Run As Administrator*).
+. From the command prompt, run {agent} under the SYSTEM account:
++
+[source,sh]
+----
+d:\tools\psexec.exe -sid "C:\Program Files\Elastic-Agent\elastic-agent.exe" run
+----
+
+
+TIP: If you install {agent} as a service as described in
+<<elastic-agent-installation>>, the Agent runs under the SYSTEM account by
+default.
 
 [discrete]
 [[agent-hangs-while-unenrolling]]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -161,6 +162,33 @@ property in the `kibana.yml` configuration file. For example:
 xpack.encryptedSavedObjects.encryptionKey: "something_at_least_32_characters"
 ----
 
+[discrete]
+[[process-not-root]]
+== {agent} fails on Windows with `Agent process is not root/admin or validation failed` message
+
+Make sure the user has administrator-level privileges.
```
dedemorton, line 169: I'm not sure how generic the error is, but if it only shows up when Endpoint is used, then I think I can remove this line. WDYT?

docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -161,6 +162,33 @@ property in the `kibana.yml` configuration file. For example:
...
+
+If you're using the {elastic-endpoint} integration, also make sure you're
+running {agent} under the SYSTEM account.
+
+To run {agent} under the SYSTEM account, you can:
```
dedemorton, line 174: I figured it might be useful to show this but my windows skills are kind of rusty. Someone will need to confirm that this is correct. I did some light testing on a VM, and it seemed OK.

docs/en/ingest-management/troubleshooting.asciidoc
EricDavisX: its not windows specific, macOS and Linux will require Administrator to run Endpoint, too

docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -161,6 +162,33 @@ property in the `kibana.yml` configuration file. For example:
...
+icon and select *Run As Administrator*).
+. From the command prompt, run {agent} under the SYSTEM account:
++
+[source,sh]
+----
+d:\tools\psexec.exe -sid "C:\Program Files\Elastic-Agent\elastic-agent.exe" run
+----
+
+
+TIP: If you install {agent} as a service as described in
```
EricDavisX, line 188: The requirement isn't just for Windows, so, we could add another line or just a 'tip' if you wanted, as:
If you are using Linux or macOS and are not installing {agent} as a service as described in <<elastic-agent-installation>>, then you will need to run the Agent with 'sudo' user.

I specifically wanted to keep that juju word sudo out of the in-product instructions, but here in a troubleshooting doc, it seems ok and appropriate to cite it

docs/en/ingest-management/troubleshooting.asciidoc
EricDavisX: I like what you have

docs/en/ingest-management/troubleshooting.asciidoc
pzl: for a more advanced audience, the correct terminology would be 

> ... needs to be run as the root user.

`sudo` being an instrument for a normal user to run something with root privileges. But that's one of those "technically correct but gives less helpful information" situations. Less advanced users, would then need to go figure out how to run something as root, and probably end up on stackoverflow telling them about `su` or `sudo`. 

Depends on the tone and audience balance being struck for this document. `sudo` being more helpful but not showing the whole picture. `root` being accurate, but would still leave questions for inexperienced unix users

docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -156,6 +157,33 @@ property in the `kibana.yml` configuration file. For example:
 xpack.encryptedSavedObjects.encryptionKey: "something_at_least_32_characters"
 ----
 
+[discrete]
+[[process-not-root]]
+== {agent} fails on Windows with `Agent process is not root/admin or validation failed` message
```
dedemorton, line 162:
```suggestion
== {agent} fails with `Agent process is not root/admin or validation failed` message
```

PR #70, opened 2020-07-29
[docs] Promote ingest management to beta

dedemorton:
Promote ingest management docs to beta in preparation in 7.9

### DIFF
--- a/docs/en/ingest-management/getting-started.asciidoc
+++ b/docs/en/ingest-management/getting-started.asciidoc
@@ -6,17 +6,10 @@
 <titleabbrev>Get started</titleabbrev>
 ++++
 
-experimental[]
+beta[]
 
 This guide describes how to get started with the new {ingest-management}
 capabilities available in this release.
-
-// tag::experimental-warning[]
-This early release of {ingest-management} is experimental. We invite you to
-install and test these capabilities **in a test environment**. You might run
-into problems and need to modify your setup to get this feature running. Please
-do not enable or use {ingest-management} in a production environment.
-// end::experimental-warning[]
 
 For feedback and questions, please contact us in the {im-forum}[discuss forum].
 
@@ -26,7 +19,7 @@ For feedback and questions, please contact us in the {im-forum}[discuss forum].
 
 Before you begin, please read <<ingest-management-limitations>>.
 
-To use this **experimental** release of {ingest-management}, you need:
+To use this **beta** release of {ingest-management}, you need:
 
 * An {es} cluster and {kib} (version 7.8) with a basic license.
 You can use our https://www.elastic.co/cloud/elasticsearch-service[hosted {ess}]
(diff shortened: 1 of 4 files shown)


dedemorton:
@mostlyjason I've removed some content that we added when the release was experimental. Let me know if I've been too aggressive. 

Also please review my comments flagged for REVIEWERS. Thanks!

docs/en/ingest-management/getting-started.asciidoc
```diff
@@ -6,18 +6,11 @@
...
 
 This guide describes how to get started with the new {ingest-management}
 capabilities available in this release.
 
-// tag::experimental-warning[]
-This early release of {ingest-management} is experimental. We invite you to
-install and test these capabilities **in a test environment**. You might run
-into problems and need to modify your setup to get this feature running. Please
-do not enable or use {ingest-management} in a production environment.
-// end::experimental-warning[]
```
mostlyjason, line 19: It'd be nice to include some kind of warning still. We are planning some big changes for the next release. What about this?

This early release of {ingest-management} is beta. We invite you to install and test these capabilities **in a test environment**. There may be bugs or incompatible changes in future beta releases as we get feedback and refine the system.

dedemorton: @mostlyjason I'm conflicted about this because I always want to do what's best for the user, but I think we need to be careful about consistent messaging here. For the alpha, it sounded like our alpha offering was exceptionally fragile even for an alpha, so the extra language seemed OK. But for beta, I'd like to be consistent with how we handle beta releases across other products/docs unless there is something different about ingest management that necessitates an exception to our usual approach. Here's the language that we get by default when we use the `beta[]` tag:

"This functionality is in beta and is subject to change. The design and code is less mature than official GA features and is being provided as-is with no warranties. Beta features are not subject to the support SLA of official GA features."

This language was carefully crafted, reviewed, and then approved by legal. I get a little antsy about the idea of adding our own twist.

Also, we have a script that we can run at GA to clean up beta admonitions. That script won't catch any hand-crafted text. Anyhow, let me know what you think.

mostlyjason: Ahh ok thanks I wasn't aware that's what the beta tag does. I think the standard text is fine. 

FYI there is an open question about whether this will be a beta release. Should be able to confirm in the next few days.

dedemorton: Ok, I'll hold off merging this until we know for sure. Thanks!

docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -35,12 +33,15 @@ Contact us in the {im-forum}[discuss forum]. Your feedback is very valuable to u
 * <<i-rebooted-my-host>>
 * <<what-is-the-endpoint-package>>
 
+//REVIEWERS: Do we want to keep the following topic or just remove it now that
+//Ingest Manager is enabled by default?
+
```
dedemorton, line 38:
```
```

dedemorton, line 38:
```
```

dedemorton:
LGTM received from mostlyjason over slack, so I'm merging.

PR #71, opened 2020-07-30
[docs] Update docs to reflect support for Endpoint integration

dedemorton:
Removes content about Endpoint that is no longer required in the troubleshooting docs because the Endpoint integration will function in 7.9.


EricDavisX:
@bradenlpreston what do you think?  @hmnichols ?

dedemorton:
It made sense to mention Endpoint in the troubleshooting section for 7.8, but since it works in 7.9, the info doesn't belong under troubleshooting. Something like that belongs in an overview, but still might be out of place because we don't highlight other integrations.

IMO, until we build out more overview content and put some thought into how we want to present this info, it's probably best to leave the "teasers" to the marketing team so that we are on-message.

dedemorton:
Replaced by #101

PR #98, opened 2020-08-13
document limits on Cloud for 7.9

roncohen:
Initial attempt at documenting the limits we arrived at for 7.9 on Cloud. Would love to get your input on how to arrange it nicely @dedemorton


roncohen:
looks great, thank you @dedemorton !

PR #99, opened 2020-08-13
[7.x] document limits on Cloud for 7.9 (#98)

roncohen:
Backports the following commits to 7.x:
 - document limits on Cloud for 7.9  (#98)


roncohen:
"Merge & squash" button was green so I assumed both checks had passed, but they hadn't. Apologies.

PR #111, opened 2020-08-21
Update troubleshooting.asciidoc

mostlyjason:
Updating path to the logs. @ph can you confirm this is correct? This is what I saw in my linux tar.gz install.

### DIFF
--- a/docs/en/ingest-management/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting.asciidoc
@@ -211,7 +211,7 @@ it might not be started. You need to <<run-elastic-agent,start {agent}>>.
 [[where-are-the-agent-logs]]
 == Where does {agent} store logs after startup?
 
-When started successfully, {agent} logs are stored in the the folder where {agent} 
+When started successfully, {agent} logs are stored in the folder where {agent} 
 was started. Also, {metricbeat} logs are stored in `data/logs/default`. If that log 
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -211,10 +211,10 @@ it might not be started. You need to <<run-elastic-agent,start {agent}>>.
 [[where-are-the-agent-logs]]
 == Where does {agent} store logs after startup?
 
-When started successfully, {metricbeat} logs are stored in
-`data/logs/metricbeat` under the folder where {agent} was started. If that log
-path does not exist, {agent} was unable to start {metricbeat}, which is a
-higher level problem to triage.
+When started successfully, {agent} logs are stored in the the folder where {agent}
```
dedemorton, line 214:
```suggestion
When started successfully, {agent} logs are stored in the folder where {agent} 
```

dedemorton:
@ph Should I merge this when the CI passes (even though I don't think it's accurate for all platforms (as noted in my previous comment). Not sure if Jason is planning to add info about the other platforms.

ph:
@dedemorton good catch, theses path are only valid for the tarball and not the installed agent on windows, linux.

dedemorton:
Closing this issue because we need to document the log locations for all platforms, not just macOS.

Here's the related issue: elastic/observability-docs#119

PR #120, opened 2020-08-26
Make troubleshooting and FAQ separate topics

dedemorton:
The troubleshooting topic was getting quite long. I think it's worth having an FAQ right now. Eventually all the topics in the FAQ should be covered in the main docs or no longer necessary when the software is GA. At that time, we might want to remove the FAQ.

### DIFF
--- a/docs/en/ingest-management/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting.asciidoc
@@ -10,6 +10,7 @@ described here, please review open issues in the following GitHub repositories:
 * https://github.com/elastic/kibana/issues[{kib}]
 * https://github.com/elastic/beats/issues[{beats}]
 * https://github.com/elastic/package-registry/issues[{package-registry}]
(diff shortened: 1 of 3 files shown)


docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -1,40 +1,18 @@
...
 
-We have collected the most common known problems and frequently asked questions
-here. If your question isn't answered here, please review open issues in the
-following GitHub repositories:
+We have collected the most common known problems here. If your problem isn't
+described here, please review open issues in the following GitHub repositories:
 
 * https://github.com/elastic/kibana/issues[{kib}]
 * https://github.com/elastic/beats/issues[{beats}]
 * https://github.com/elastic/package-registry/issues[{package-registry}]
```
EricDavisX, line 13: do we want to do some cross linking here?  ... we could put in the new public endpoint repo:
https://github.com/elastic/endpoint/issues

PR #143, opened 2020-09-24
Add FAQ topics about how agent accesses integrations

### LINKED ISSUE #118: [docs] Ingest Manager & Agent (and Integrations / Beats / Endpoint) data flow citation is needed in docs
Hi, internal and external questions are arising about what Ingest Manger and Agent are doing in terms of where data is coming from and how it gets to where it goes for setup, etc.  Users familiar with Beats may have an incorrect inherited assumption, and so we can document it all together, in one place, and in a way to clarify and inform new-to-Beats users, too.

Content is below, tho  you can frame it all as desired in the Docs, I don't think it needs to be in the FAQ or listed as Q&A necessarily:   

Does the Elastic Agent download or does Kibana download the Integration packages needed from the internet?

answer:  Kibana connects to the Elastic package registry at epr.elastic.co and downloads the Packages and stores their assets in Elasticsearch.  This done for the user now in Ingest Manager as opposed to configuring and executing it manually as in stand-alone Elastic Beats usage.

Does Elastic Agent download anything at all from the Internet?

answer: No, with one special exception.  Since Metricbeat and Filebeat are bundled with the Elastic Agent they are on the host when Agent is installed.  The special case is the Elastic Endpoint, which is downloaded by the Agent based OS and architecture from the Elastic download site, when Elastic Agent Policy is set to include it.  Bundling Elastic Endpoint together with Elastic Agent is a known feature request scoped for a future release (link to issue?)

Do the Beats that Agent configures on the host send assets into Elasticsearch like stand-alone beats do?

answer: No, this is handled by Ingest Manager now.  For the example of the stand-alone Elastic Beat Nginx module, the Integrations packages on epr.elastic.co contain what used to be in the Filebeat or Metricbeat module.  That is, the Elasticsearch and Kibana assets. As the assets are now managed by Ingest Manager, there is no need for the Elastic Agent configured Beats to run their setup step and send assets to Elasticsearch and/or Kibana.

dedemorton:
Closes #118.

I still plan to add an architecture overview, but hopefully this will help Beats users who are trying to understand how agent is different from standalone Beats.

@EricDavisX Let me know if you want me to backport this to 7.9, too.

### DIFF
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -106,6 +106,48 @@ Support for installing {agent} as a service on all supported systems will be
 available in a future release.
 
 [discrete]
+[[does-agent-download-packages]]
+== Does {agent} or {kib} download integration packages?
+
+{agent} does not download integration packages. When you add an integration in
+{ingest-manager}, {kib} connects to the {package-registry} at `epr.elastic.co`,
+downloads the integration package, and stores its assets in {es}. This means
+that you no longer have to run a manual setup command to load integrations as
+you did previously with {beats} modules.
+
+[discrete]
+[[does-agent-download-anything-from-internet]]
+== Does {agent} download anything from the Internet?
+
+// REVIEWERS: I'm being intentionally non-specific here and not mentioning Beats
+// because I think that's the general direction we're heading. I will only
+// mention Beats in the ingest management docs when necessary. Does that sound
+// right? 
+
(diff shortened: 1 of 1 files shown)


EricDavisX (approved):
lgtm; note the bundling of Agent to have Endpoint is hoped for a near term .minor release, so we can tag that section to be removed.  thanks so much.

EricDavisX:
yeah, a 7.9 backport would be great.

PR #161, opened 2020-10-02
Fix URL for Elastic Package Registry

dedemorton:
Changes the URL of the Elastic Package Registry to include version info for 7.9 only.

@ruflin From what I understand, we only want to document this for 7.9, not other releases. Please let me know if that's not correct.


kaiyan-sheng (approved):
lgmt

PR #183, opened 2020-10-14
Update Readme

EamonnTP:
To provide a little more guidance to contributors, and that the Logs, Metrics, and Uptime content will reside in the Observability Guide from v7.10.0 onwards, it might be best to call out the directory structure in the readme.


bmorelli25 (approved):
This is great. Thanks!

PR #225, opened 2020-11-05
Update limits based on effort for 7.10

ph:
This updates the limits 8000 for 8gb, 4k for 4gb and 2k for 1gb .

### DIFF
--- a/docs/en/ingest-management/fleet-limitations.asciidoc
+++ b/docs/en/ingest-management/fleet-limitations.asciidoc
@@ -20,16 +20,15 @@ This feature has additional limitations at the current time:
 *   Requires internet access for {kib} to download integration packages
 *   No support for advanced {beats} settings like multiline, processors, and so
 on
-*   No support for Kubernetes or Docker autodiscovery
 *   On {ecloud}, there's a limit to the number of enrolled {agent}s that a
 single {kib} instance can handle:
 +
 [%header]
 |===
 |{kib} instance size {ess-icon} |Max. number of {agent}s
-|8gb| 1000
-|4gb| 750
-|1gb| 500
+|8gb| 8000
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet-limitations.asciidoc
```diff
@@ -20,16 +20,15 @@ This feature has additional limitations at the current time:
 *   Requires internet access for {kib} to download integration packages
```
dedemorton, line 20: For line 19, I thought that we added proxy support. The What's new for 7.10 says: "Configure a proxy in Fleet to access the internet." We haven't documented that yet, either.

docs/en/ingest-management/fleet-limitations.asciidoc
```diff
@@ -20,16 +20,15 @@ This feature has additional limitations at the current time:
...
 *   No support for advanced {beats} settings like multiline, processors, and so
```
dedemorton, line 21: Should you remove processors as an example and say something else here? I'm assuming processors are supported since they are shown in all the variable/condition examples. :-)

dedemorton (commented):
The changes look good, but there are a few other limitations we list here that I'm not sure are exactly true.

docs/en/ingest-management/fleet-limitations.asciidoc
ph: I will do a followup for the proxy support with @jfsiii

docs/en/ingest-management/fleet-limitations.asciidoc
ph: is still not expose in fleet so I'll keep it.

dedemorton (approved):
LGTM

PR #227, opened 2020-11-06
[7.x] Update limits based on effort for 7.10 (#225)

ph:
Backports the following commits to 7.x:
 - Update limits based on effort for 7.10 (#225)


dedemorton (approved):
LGTM

PR #279, opened 2020-12-01
Use un-authenticated git URL for synthetics examples

andrewvc:
The SSH URL only works if you have valid SSH creds w/ push pull which applies to us Elastic employees but no one else


bmorelli25 (approved):
Ah, thanks!

PR #289, opened 2020-12-07
Clarify where to run examples

andrewvc:
This can be confusing for users, since running npx commands outside of the `todos` folder, but within the synthetics repo causes weird issues since that repo defines the synthetics package itself.
  
In the future we should find a better method than asking users to clone the repo, if they could just get the examples folder that would reduce the number of places errors creep in.

### DIFF
--- a/docs/en/observability/synthetics-create-test.asciidoc
+++ b/docs/en/observability/synthetics-create-test.asciidoc
@@ -250,7 +250,7 @@ npm install -g @elastic/synthetics
 
 Now it's time to write your tests:
 
-. Create a new Node.js project.
+. Create a new NPM/Node.js project.
 . Create a `javascript` or `typescript` file that imports your tests.
 All synthetic test files must use the `.journey.ts` or `.journey.js` file extension.
 . Compile everything together.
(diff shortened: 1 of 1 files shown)


docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -261,17 +261,28 @@ If you'd like to test it locally, clone the repo, and install the example:
...
+The `todos` folder has its own `package.json`, which makes it its own NPM project. If you
+run these commands outside the `todos` folder you can run into confusing situations with
+NPM's module resolution that can cause this command to not work.
 
 [source,sh]
 ----
-npx @elastic/synthetics examples/todos
+# Run tests on the current directory. Please note the dot `.` which indicates 
+# that we want to run tests on the current directory
+# current directory!
```
bmorelli25, line 284: Repeated
```suggestion
```

bmorelli25 (approved):
One recommendation, otherwise LGTM.

bmorelli25:
> In the future we should find a better method than asking users to clone the repo, if they could just get the examples folder that would reduce the number of places errors creep in.

We have an Observability repo for project tutorial files. If it makes sense at some point in the future to move the `todos` example out of elastic/synthetics, we can do that.

PR #313, opened 2020-12-23
docs: Create shared directory and add APM Kubernetes info

bmorelli25:
## Summary

* Adds APM agent Kubernetes installation instructions
* Creates a shared content folder in the `observability-docs` repo.

I grouped each widget and its content together in individual folders. IMO, this makes it easier to find the content you're looking for. Open to feedback!

Screen Shot 2020-12-23 at 11 09 26 AM


docs/en/observability/add-observability-data.asciidoc
```diff
@@ -12,30 +12,30 @@ You can use our
 https://www.elastic.co/cloud/elasticsearch-service[hosted {ess}]
 on {ecloud} (recommended), or self-manage the {stack} on your own hardware.
 
-include::tab-widgets/deploy-ess-stack-widget.asciidoc[]
+include::./../shared/spin-up-the-stack/widget.asciidoc[]
```
dedemorton, line 15: I've started using an attribute called `{tab-widgets}` to resolve the path (so you can move things around without having to update relative links). Would also let us add a shared attributes file to define the location for all projects. Just build the path starting with a built-in attribute (or an attribute based on one that's built in), like `{observability-docs-root}`, so that it works from anywhere.

dedemorton (approved):

docs/en/observability/add-observability-data.asciidoc
bmorelli25: Brilliant

bmorelli25:
Doc build still 404s. I'm going to take a risk and merge this.

Issue #323, opened 2021-01-05
Add tutorials and quick start videos to the observability guide


dedemorton:
The observability docs team will partner with the education team to publish tutorials in the observability guide. This issue is to track cross-team tasks and communication.

Note: This repository is public, so please do not include links to internal documents.

**Participants and stake holders:**

Writing team: @bmorelli25, @dedemorton, @EamonnTP 
Education team: @melvynator 

**Goals**

Deliver tutorial and guides to help customers solve real-world use cases.

**Tasks**

- [x] Embed videos in the current version of the observability guide. Issue: https://github.com/elastic/observability-docs/pull/324 [@bmorelli25] 
- [x] Provide education team with asciidoc tutorial template and work to refine it. PR: https://github.com/elastic/observability-docs/pull/54

**Open issues and questions**

* Does the education team need help ramping up on how to contribute to the docs?

bmorelli25:
Closing this as all tasks are completed. Additional information on the cloud monitoring tutorials created by the education team can be found in https://github.com/elastic/observability-docs/issues/299.

PR #324, opened 2021-01-11
docs: Add quickstart videos

bmorelli25:
## Summary

This PR adds quick start videos to the Observability Guide:
* [Logs](https://observability-docs_324.docs-preview.app.elstc.co/guide/en/observability/master/monitor-logs.html)
* [Metrics](https://observability-docs_324.docs-preview.app.elstc.co/guide/en/observability/master/analyze-metrics.html)
* [UE](https://observability-docs_324.docs-preview.app.elstc.co/guide/en/observability/master/user-experience.html)
* [APM](https://observability-docs_324.docs-preview.app.elstc.co/guide/en/observability/master/apm.html)

Screen Shot 2021-01-11 at 9 01 19 AM

*Notes*

* The first video offers an overview of all three videos, and the subsequent videos start with "Step x", so it doesn't seem like a good user experience to split the videos up (unless they were all split onto the same page).
* I think the videos can replace the screenshots we have on each of these pages as you don't need to hit play to see the thumbnails. I wonder if we might be able to improve the APM thumbnail though – it's a bit... boring? See what I mean [here](https://www.elastic.co/training/apm-quick-start), then compare with a screenshot we have [here](https://www.elastic.co/guide/en/observability/7.10/apm.html) or [here](https://www.elastic.co/guide/en/kibana/7.10/transactions.html).

## Related issues

For https://github.com/elastic/observability-docs/issues/323.


dedemorton (approved):
LGTM!

This works really well!

One question I have is around versioning. I assume that we will maintain old versions of the videos to support older docs?

dedemorton:
Another thought: To address the problem you describe in your notes, would it make sense to show "related video" links below the video? I think having the quick start on the individual pages is a lot better for folks who skip the overview. I wonder if the videos can be tweaked at some point to be more standalone.

melvynator:
I really like it! It looks great!

We can change the thumbnail for APM with something more exciting in the 7.11 release. Will that be ok? 

> One question I have is around versioning. I assume that we will maintain old versions of the videos to support older docs?

We don't maintain old videos. The newer version erases the older version. At the same time, the videos are really focusing on getting started with the cloud. As the cloud UI changes it will make older videos irrelevant/deprecated.

melvynator:
I have one thought with regards to your 2 points above @dedemorton. Maybe we should make a clearer distinction between the video and the text by adding some small title like: "Learn how to get get started with log monitoring in the cloud"

dedemorton:
>We don't maintain old videos. The newer version erases the older version. 

Then we'll need to figure out how to handle the video links when we release a new version. I might be overthinking this a little, but if at sometime in the future users are on an older version (maybe their org is slow to upgrade), it's not going to be a great experience if the videos show functionality that doesn't exist in that version.

@melvynator At the very least, we should add something to indicate that the video applies to the latest released version. Otherwise, users might see something in the video and get really frustrated when they can't figure out how to do it from reading the docs.

melvynator:
If you think a note can do it I will be happy with that, we can also display the video only on master and the latest version if you think that would be giving a better experience.

bmorelli25:
> We can change the thumbnail for APM with something more exciting in the 7.11 release. Will that be ok?

Yes, that works I think.

> it's not going to be a great experience if the videos show functionality that doesn't exist in that version.

> If you think a note can do it I will be happy with that, we can also display the video only on master and the latest version if you think that would be giving a better experience.

Hmm. I had assumed old videos would not be overwritten. I wonder if we can work around this with attributes in the docs? Something like `if (branch == current) show video`? Let me play around with it.

> Maybe we should make a clearer distinction between the video and the text by adding some small title like: "Learn how to get get started with log monitoring in the cloud"

I initially had added a header similar to this, but it felt too disconnected from the rest of the content.

EamonnTP:
Loving these updates!

bmorelli25:
Blocked while we figure out a way to conditionally show videos in `current`. Thread in #obs-docs.

bmorelli25:
Going to merge and backport this as-is so we can get the videos up. I'll follow-up with a new PR that conditionally shows a screenshot or a video depending on the doc version.

PR #343, opened 2021-01-26
Update FAQ info about agent policies

### LINKED ISSUE #282: Verify accuracy of FAQ topic about agent policy
The security writers pointed out that this information in the docs might be out-of-date:

https://www.elastic.co/guide/en/fleet/current/fleet-faq.html#what-is-my-agent-config

dedemorton:
Closes #282 

This FAQ topic was added in an earlier release. Looks like the config has changed quite a bit so that this topic is no longer relevant. 

The Agents tab clearly lists the policy that is assigned to each enrolled agent, so I don't think we need an FAQ topic for this.

### DIFF
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -35,21 +35,41 @@ higher level problem to triage.
 == What policy is the {agent} running?
 
 To find the policy file, inspect the `elastic-agent.yml` file in the
-folder where you ran {agent}. If you're running the agent in {fleet} mode, this
-file contains the following citation:
+directory where {agent} is running. Not sure where the agent is running? See 
+<<installation-layout>>.
+
+If the agent is running in {fleet} mode, this file contains the following
+citation:
 
 [source,yaml]
 ----
-Management: mode: "fleet"
+fleet:
+  enabled: true
 ----
 
-The `action_store.yml` contains the entire, unencrypted policy:
+The `action_store.yml` file (located under `data/elastic-agent-*`) contains the
+entire, unencrypted policy.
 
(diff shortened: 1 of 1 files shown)


EricDavisX (changes requested):
I suggest we close this and leave the info.  the problem we are solving with leaving this in is that support / engineers may want to talk users thru how to find things on the host, as they are used by the running Agent.  Looking at things in the Fleet UI is easier (usually) but is not reflective of the state on the host necessarily.

dedemorton:
@EricDavisX Oh I see where I got off into the weeds. I forgot to look in the directory where the agent's running (rather than extract path) for the files described in this topic. I will clarify this because users are likely to get confused too.

docs/en/ingest-management/faq.asciidoc
```diff
@@ -35,21 +35,41 @@ higher level problem to triage.
...
+    hosts:
+    - https://3m63533c175a4036b3d8bbe7bd462fa3.us-east-1.aws.found.io:443
+    type: elasticsearch
+----
 
 This file also shows the version of all packages used by the current
 policy.
+--
+
+* To see the {agent} version, look at the download folder and zip filenames.
```
EricDavisX, line 71: this is ok to leave as is, but we could update it now or later if we want to cite using the running agent to execute the 'version' command to get better details.  it would need to be run differently on linux/mac and windows of course, so it gets more verbose quickly.

EricDavisX (approved):
looks good - we could add more info on agent version now or later as you desire.

docs/en/ingest-management/faq.asciidoc
dedemorton: That's a good idea. Added in the latest commit.

PR #344, opened 2021-01-26
Add privileges required to run Elastic Agent standalone

### LINKED ISSUE #287: Document permissions required to run Elastic Agent standalone
We tell users that superuser is currently required to use Fleet, but the docs don't mention the permissions required to run standalone.

Currently we use the following permissions when we create the API key for the agent. These permissions need to be mentioned in the docs that describe how to configure standalone agents:

```
{
          names: ['logs-*', 'metrics-*', 'events-*', '.ds-logs-*', '.ds-metrics-*', '.ds-events-*'],
          privileges: ['write', 'create_index', 'indices:admin/auto_create'],
        },
```

dedemorton:
Closes #287 

See the preview here: https://observability-docs_344.docs-preview.app.elstc.co/guide/en/fleet/master/run-elastic-agent-standalone.html

I think it's a little weird to show the API call inside of the steps, but I think it's Ok since we will hopefully replace the example with the role name when it's available.

TODO:
- [x] Test privileges to confirm that they work as expected.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/run-elastic-agent-standalone.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/run-elastic-agent-standalone.asciidoc
@@ -48,6 +48,7 @@ deployment.
 . Modify `ES_USERNAME` and `ES_PASSWORD` in the outputs section of
 `elastic-agent.yml` to use your {es} credentials. For example:
 +
+--
 [source,yaml]
 ----
 [...]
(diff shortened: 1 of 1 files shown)


dedemorton:
The Kibana UI for defining roles does not accept `indices:admin/auto_create` as a valid privilege. Waiting to hear back for SMEs about what to provide here, so I'll mark this as draft until I've heard back.

dedemorton:
I want to point out here that @legrego made the following comment recently. I agree with him, but AFAIK there is no role that has the right privileges (except `fleet_enroll`, which is not appropriate for users who are running standalone agents). So I propose that we document the privileges now and someone on the dev team should follow up with the security team to get a role added for the standalone use case.

Comment from @legrego:
"...we really don’t want to encourage the use of these “granular” privileges, so I’d love to see them removed from our public docs. They are implementation details of Elasticsearch, and are not subject to the same backwards compatibility guarantees. We strongly recommend using the built-in supported privileges instead. If the built-in privileges aren’t granular enough, then that’s something we can raise to the ES security team for evaluation."

docs/en/ingest-management/elastic-agent/run-elastic-agent-standalone.asciidoc
```diff
@@ -56,10 +57,31 @@ outputs:
...
     hosts:
       - 'http://localhost:9200'
-    username: ES_USERNAME
+    username: ES_USERNAME <1>
     password: ES_PASSWORD
 [...]
 ----
+<1> This user must have the privileges shown in the following example.
+
+To create a role that has the privileges need by the {agent} user, go to
```
EamonnTP, line 66:
```suggestion
To create a role that has the privileges needed by the {agent} user, go to
```

EamonnTP (approved):
LGTM. Just one minor suggestion.

docs/en/ingest-management/elastic-agent/run-elastic-agent-standalone.asciidoc
dedemorton: Whoops! Thanks for catching this.

EricDavisX (approved):
looks ok to me

bmorelli25 (approved):

PR #350, opened 2021-01-26
Conditionally display videos in current version

bmorelli25:
## Summary

This PR conditionally displays videos in the observability guide when `is-current-version == true`. When `false`, backup screenshots are displayed instead. I've looked at this a million different ways, and this is by far the easiest way to accomplish this. This PR requires two lines in the `elastic/docs` repo to be changed on each minor release.

## Related

* This PR is a follow-up to https://github.com/elastic/observability-docs/pull/324. 
* For #323

This PR requires an additional change in the `elastic/docs` repo before merging: https://github.com/elastic/docs/pull/2044

PR #353, opened 2021-01-27
[7.10] docs: conditionally display videos in obs-docs (#350)

bmorelli25:
Backports the following commits to 7.10:
 - docs: conditionally display videos in obs-docs (#350)


bmorelli25:
HTML diff is empty as expected!

PR #373, opened 2021-02-05
fix Filebeat config step 5 labels

daverick:
the proposed config for Filebeat with auto discover and labels is not working as the `fileset` settings is only working (and [documented](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-autodiscover-hints.html) ) on annotations with auto discover and hints but not with simple auto discover  on labels.

the commit propose a config that is working but a better solution might be available.


ChrsMark (approved):
lgtm, thanks!

PR #374, opened 2021-02-06
Add troubleshooting problem about telemetry issue in 7.11

### LINKED ISSUE #318: Quick 7.11 Docs note to support Endpoint Security Telemetry feature Agent side requirements
Hi.  @kevinlog mentions this to the Ingest Team, and we wanted to document it.  We will need to quickly confirm specifics to type up the docs.

https://github.com/elastic/kibana/pull/86277

This ticket merges a small change, about the Protections Team Telemetry feature, but I don't know off hand the larger meta ticket, sorry.  Kevin can you post and update?

Assigning Janeen and Dede to coordinate on which parts may go into one or both of Security side and Fleet / Agent side docs.

We know this:

if you upgrade an Agent from pre-7.11 the Telemetry feature will be on by default in the 7.11 Agent and it will attempt to write to the new index per the above issue, but it will fail (because the API key, as upgraded, will not have permissions).

Nothing else bad will happen, apart from errors in the logs, mind you.  but we expect users to notice and ask about it so we can be proactive.

The work arounds we have so far are this:
 - re-enroll the Agent to pick up the API key changes (which should be available automatically after upgrading, right Kevin?)
 - change the Kibana side policy for this Telemetry feature to turn it off so it won't keep logging error messages on Elastic Endpoint side as well as Elastic index side (likely 1 error per Elastic Endpoint per some time frame).  We don't know for sure what the back-off policy is on attempting this, so it could be a lot of noise in the Elastic and Endpoint side logs.  Kevin was going to log a ticket for improving that back-off strategy for this specific case (as needed).

dedemorton:
Closes #318

### DIFF
--- a/docs/en/ingest-management/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting.asciidoc
@@ -178,4 +178,37 @@ If this happens, select **Force unenroll** from the *Actions* menu in the
 Agents list.
 
 This will invalidate all API keys related to the agent and change the status to
-`inactive` so that the agent no longer appears in {fleet}. 
+`inactive` so that the agent no longer appears in {fleet}.
+
+[discrete]
+[[endpoint-unauthorized]]
+== API key is unauthorized to send telemetry to `.logs-endpoint.diagnostic.collection-*` indices
+
+Telemetry is turned on by default in {kib}. This helps us learn about the
+features that our users are most interested in, so we can focus our efforts on
+making them even better.
+
+If you've recently upgraded from version 7.10 to 7.11, you might see the
+following message when you view {endpoint-sec} logs:
+
+----
+action [indices:admin/auto_create] is unauthorized for API key id [KbvCi3YB96EBa6C9k2Cm]
+of user [fleet_enroll] on indices [.logs-endpoint.diagnostic.collection-default]
+----
+
+This message indicates that {elastic-endpoint} does not have the correct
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -178,4 +178,37 @@ If this happens, select **Force unenroll** from the *Actions* menu in the
...
 
 This will invalidate all API keys related to the agent and change the status to
-`inactive` so that the agent no longer appears in {fleet}. 
+`inactive` so that the agent no longer appears in {fleet}.
+
+[discrete]
+[[endpoint-unauthorized]]
+== API key is unauthorized to send telemetry to `.logs-endpoint.diagnostic.collection-*` indices
+
+Telemetry is turned on by default in {kib}. This helps us learn about the
```
dedemorton, line 187: Not sure if this should say "in Kibana" or "in the Elastic Stack.

kevinlog: Say Elasticstack

kevinlog (approved):

docs/en/ingest-management/troubleshooting.asciidoc
kevinlog: *Elastic Stack

dedemorton, line 187:
```suggestion
Telemetry is turned on by default in the {stack}. This helps us learn about the
```

PR #403, opened 2021-02-23
Small FAQ updates for Fleet / Agent based on recent feature work

EricDavisX:
Reading through the docs and I find a few items I had meant to help update after some recent feature work was updated.

### DIFF
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -157,7 +157,7 @@ As of 7.10 and beyond, the endpoint security capabilities are bundled with {agen
 so a fully capable artifact can be installed with no connection to the Elastic
 download site.  In most cases, however, the {elastic-endpoint-integration} process
 is instructed to download newer released versions of the malware model it uses, as
-well as other artifacts.  For more information, see the {security-guide}/index.html[Elastic Security solution documentation]
+well as other artifacts.  For more information, see the {security-guide}/index.html[Elastic Security solution documentation].
 
 [discrete]
 [[do-i-need-to-setup-elastic-agent]]
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -17,8 +17,8 @@ Also read <<fleet-troubleshooting>>.
 If {agent} was successfully enrolled, but doesn't show up in the *Agents* list,
 it might not be started. Make sure the `elastic-agent` process is running on
 the host. If it's not running, use the <<elastic-agent-run-command,`run`>>
-command to start it.  The most common usage to deploy an Agent is to use the
-"install" command which takes care of starting the {agent} for you.
+command to start it.  The most common way to deploy an {agent} is by using
+the `install` command. This command starts the {agent} for you.
 
 [discrete]
 [[where-are-the-agent-logs]]
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -17,9 +17,8 @@ Also read <<fleet-troubleshooting>>.
 If {agent} was successfully enrolled, but doesn't show up in the *Agents* list,
 it might not be started. Make sure the `elastic-agent` process is running on
 the host. If it's not running, use the <<elastic-agent-run-command,`run`>>
-command to start it.
-
-//TODO: Verify that this is correct. Or would you use the restart command?
+command to start it.  The most common usage to deploy an Agent is to use the
+"install" command which takes care of starting the {agent} for you.
 
 [discrete]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/faq.asciidoc
```diff
@@ -152,13 +153,11 @@ you did previously with {beats} modules.
 [[does-agent-download-anything-from-internet]]
 == Does {agent} download anything from the Internet?
```
EricDavisX, line 155: @paul-tavares actually, can I get your comment on what you think we could say here?  I'd like our message now to be future compliant or we can just add a separate note now about Fleet Server usage.  I see from a recent PR that the artifacts like Trusted Apps and 'Exceptions' are actually sent from Kibana currently (which downloads them) and that they will be served via Fleet Server in the future, which will need to download them from somewhere (from Kibana or elsewhere?) before passing it on.

paul-tavares: Hi @EricDavisX - The artifacts will be stored in elasticsearch, so fleet-server will have access to them and does not have the need to download them. Currently, these artifacts are Endpoint specific and the agent (or fleet) has no awareness of them. That may change with the post you saw - as we move from kibana to fleet-server, these will have to also move, but I don't yet see a need for the agent to download anything. The downloading will be done by the Endpoint if installed.

docs/en/ingest-management/faq.asciidoc
```diff
@@ -28,7 +27,9 @@ command to start it.
 When started successfully, {metricbeat} logs are stored in
 `data/logs/metricbeat` under the folder where {agent} was started. If that log
 path does not exist, {agent} was unable to start {metricbeat}, which is a
-higher level problem to triage.
+higher level problem to triage.  Usually you can see these logs in the Fleet
+UI, unless there are problems severe enough that the processes cannot send 
+data to Elasticsearch.
```
dedemorton, line 32:
```suggestion
higher level problem to triage. Usually you can see these logs in the {fleet}
UI, unless there are problems severe enough that the processes cannot send 
data to {es}.
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -152,13 +153,11 @@ you did previously with {beats} modules.
...
-{agent} policy is set to include the {elastic-endpoint-integration} integration,
-{agent} must download software from the Elastic download site.
-
-NOTE: Bundling endpoint security capabilities with {agent} is a known
-feature request scoped for a future release. 
+As of 7.10 and beyond, the endpoint security capabilities are bundled with {agent},
+so a fully capable artifact can be installed with no connection to the Elastic
+download site.  In most cases, however, the {elastic-endpoint-integration} process
+is instructed to download newer released versions of the malware model it uses, as
+well as other artifacts.  For more information, see the {security-guide}/index.html[Elastic Security solution documentation]
```
dedemorton, line 160:
```suggestion
well as other artifacts.  For more information, see the {security-guide}/index.html[Elastic Security solution documentation].
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -17,9 +17,8 @@ Also read <<fleet-troubleshooting>>.
 If {agent} was successfully enrolled, but doesn't show up in the *Agents* list,
 it might not be started. Make sure the `elastic-agent` process is running on
 the host. If it's not running, use the <<elastic-agent-run-command,`run`>>
-command to start it.
-
-//TODO: Verify that this is correct. Or would you use the restart command?
+command to start it.  The most common usage to deploy an Agent is to use the
+"install" command which takes care of starting the {agent} for you.
```
dedemorton, line 21:
```suggestion
command to start it.  The most common way to deploy an {agent} is by using
the `install` command. This command starts the {agent} for you.
```

docs/en/ingest-management/faq.asciidoc
dedemorton: "the processes" seems a little vague. Can we just say "the agent cannot send"? I think maybe you are talking about the underlying Beats shippers, but I wonder if we can generalize here. Or maybe we can say something like "severe enough that the agent or its related processes cannot send..."

dedemorton (commented):
Just a few minor changes. Thanks for updating this!!

docs/en/ingest-management/faq.asciidoc
EricDavisX: I like your latter note, I'll do that.

EricDavisX: done

EricDavisX:
Ready for re-review.

EricDavisX:
I don't know fully if the 7.11 docs backport is helpful at this point, but i included it.

PR #410, opened 2021-02-25
7.12-update for enroll command

EricDavisX:
Relates to https://github.com/elastic/beats/pull/24155
Fixes Docs need in: https://github.com/elastic/beats/issues/24199

we updated the Agent enroll usage, and it is no longer working without the params passed, so we are updating the docs to reflect.


dedemorton (commented):
Nice. Do you mind updating the synopsis and descriptions in `docs/en/ingest-management/elastic-agent/elastic-agent-command-line.asciidoc` too?

EricDavisX:
happily. i can get it done.  thanks for pointing it out!

EricDavisX:
done and ready for re-review.  I started to put in an addition for 'upgrade' only being available in stand-alone mode but realized that Jason had done a much better text in a separate pr so I undid mine.  woo.

PR #413, opened 2021-02-25
[DOCS] Updates Kibana path for Filebeat

KOTungseth:
Since `Kibana` has been replaced with `Analytics` in the main menu, I removed `Kibana`, and updated to text to match what we use in the Kibana docs.

### DIFF
--- a/docs/en/observability/ingest-logs.asciidoc
+++ b/docs/en/observability/ingest-logs.asciidoc
@@ -180,7 +180,7 @@ Let's confirm your data is correctly streaming to your cloud instance.
 include::{beats-repo-dir}/tab-widgets/open-kibana-widget.asciidoc[]
 --
 
-. In the side navigation, click *{kib} > Discover*.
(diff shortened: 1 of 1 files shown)


EamonnTP:
Thanks @KOTungseth for the update.

Elsewhere in the Observability Guide, we call out the main UI heading and the app. For example, `To view the Logs app, go to Observability > Logs`. Should we update these steps to be consistent with your update? If so, that might be something like this, `To view the Logs app, open the main menu, then click Logs`. 

Including @elastic/obs-docs for visibility.

KOTungseth:
Before the switch to `Analytics`, I removed all instances of `Kibana` since it serves as the title and is not actionable. I would use `To view the Logs app, open the main menu, then click Logs.`

EamonnTP (approved):

dedemorton:
@KOTungseth I think this need to be backported to 7.x and master, no?

KOTungseth:
@dedemorton in the works!

PR #414, opened 2021-02-25
[DOCS] Updates Kibana path for Metricbeat

KOTungseth:
Since `Kibana` has been replaced with `Analytics` in the main menu, I removed `Kibana`, and updated to text to match what we use in the Kibana docs.


EamonnTP (approved):
LGTM!

PR #499, opened 2021-04-08
Add leaderelection docs

ChrsMark:
This PR adds docs for `kubernetes_leaderelection` provider introduced at https://github.com/elastic/beats/pull/24913

### DIFF
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
@@ -195,6 +195,33 @@ access Kubernetes Secrets API and get a specific secret no matter which namespac
 In this, this option should be carefully considered.
 
 [discrete]
+[[kubernetes_leaderelection-provider]]
+==== Kubernetes LeaderElection Provider
+
+Provides the option to enable leaderelection between a set of Agents
+running on Kubernetes. Only one Agent at a time will be the holder of the leader
+lock and based on this, configurations can be enabled with the condition
+that the Agent holds the leadership. This can be useful in cases where we want
+only Agent between a set of Agents to collect cluster wide metrics for the
+Kubernetes cluster like from `kube-state-metrics` endpoint.
+
+Provider needs a `kubeconfig` file so as to establish connection to Kubernetes API,
+or it can automatically reach the API if it runs in an inCluster environment (Agent runs as Pod).
+
+[source,yaml]
+----
+providers.kubernetes_leaderelection:
+  #kube_config: /Users/elastic-agent/.kube/config
+  #leader_lease: agent-k8s-leader-lock
+----
+
+`kube_config`:: (Optional) Use given config file as configuration for Kubernetes
+client. If kube_config is not set, KUBECONFIG environment variable will be
+checked and if not present it will fall back to InCluster.
+`leader_lease`:: (Optional) Specify the name of the leader lease.
+By default it is `elastic-agent-cluster-leader`.
+
+[discrete]
 [[dynamic-providers]]
 === Dynamic Providers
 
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
```diff
@@ -194,6 +194,33 @@ This means that anyone who have access to Agent Pod (`kubectl exec` for example)
...
+providers.kubernetes_leaderelection:
+  #kube_config: /Users/elastic-agent/.kube/config
+  #leader_lease: agent-k8s-leader-lock
+----
+
+`kube_config`:: (Optional) Use given config file as configuration for Kubernetes
+client. If kube_config is not set, KUBECONFIG environment variable will be
+checked and if not present it will fall back to InCluster.
+`leader_lease`:: (Optional) Specify the name of the leader lease.
+By default it is `elastic-agent-cluster-leader`.
```
exekias, line 222: Should we add a table with the available key from this provider?

docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
```diff
@@ -194,6 +194,33 @@ This means that anyone who have access to Agent Pod (`kubectl exec` for example)
...
+Kubernetes cluster like from `kube-state-metrics` endpoint.
+
+Provider needs a `kubeconfig` file so as to establish connection to Kubernetes API,
+or it can automatically reach the API if it runs in an inCluster environment (Agent runs as Pod).
+
+[source,yaml]
+----
+providers.kubernetes_leaderelection:
+  #kube_config: /Users/elastic-agent/.kube/config
+  #leader_lease: agent-k8s-leader-lock
```
exekias, line 215: This is not related to this PR but I have a question myself, do we have a parameter to pass the node name for `inCluster` discovery? Do we still need it in this context? In the past we got many issues related to the hostname of the Pod being different from the Pod name / node name

docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
ChrsMark: Nice catch! Thanks!

docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
ChrsMark: I don't think that we need because `inCluster` here has only to do with the k8s client initialisation, see https://github.com/elastic/beats/blob/2494ae509b1036d39bd1bbb4ec751c50347c9aff/x-pack/elastic-agent/pkg/composable/providers/kubernetesleaderelection/kubernetes_leaderelection.go#L52.

Usually `Node` setting is used by another function called `DiscoverKubernetesNode`  (see https://github.com/elastic/beats/blob/8b2cebb6229fa93aa580c6fb93dc65f988f3226c/x-pack/elastic-agent/pkg/composable/providers/kubernetes/kubernetes.go#L67) which is not used in this provider.

So I think that we are covered here?

exekias: Good point, I think we are fine then!

docs/en/ingest-management/elastic-agent/elastic-agent-providers.asciidoc
```diff
@@ -194,6 +194,67 @@ This means that anyone who have access to Agent Pod (`kubectl exec` for example)
...
+By default it is `elastic-agent-cluster-leader`.
+
+The available key is:
+
+|===
+|Key |Type |Description
+
+|`kubernetes_leaderelection.leader`
+|`bool`
+|The value of the leadership flags. True when the Agents holds the leader lock.
```
exekias, line 231: I would rephrase a little bit:

```suggestion
|The value of the leadership flag. It is `true` when the Agent is the current leader, `false` otherwise.
```

exekias (approved):

PR #508, opened 2021-04-13
Mention command for generating encryption keys

dedemorton:
Adds pointer to docs about generating encryption keys.

Related issue: https://github.com/elastic/kibana/pull/82838. I [said we would add this info](https://github.com/elastic/kibana/pull/82838#issuecomment-730027156), but lost track of the change.

(writers: try to ignore the passive voice. :-) it works better with the current bulleted list structure)


bmorelli25 (approved):
> (writers: try to ignore the passive voice. :-) it works better with the current bulleted list structure)

giphy (1)

😉

PR #558, opened 2021-04-21
Update tail-logs.asciidoc

leahleahy:


docs/en/observability/tail-logs.asciidoc
```diff
@@ -44,15 +44,15 @@ view when a large number of events occurred.
 When you have searched and filtered your logs for a specific log event, you may want to examine the
 metadata and the structured fields associated with that event. To view the *Log event document details* fly-out,
```
leahleahy, line 45: Can we just call this a menu or is fly-out your terminology?

PR #562, opened 2021-04-21
Update analyze-metrics.asciidoc

leahleahy:


EamonnTP (commented):
I would keep the original as it stresses the importance. I would say it's safe to close this PR.

PR #565, opened 2021-04-21
Update monitor-kubernetes-pods.asciidoc

leahleahy:


EamonnTP (commented):
I would keep the original as it stresses how important it is to install and configure Metricbeat. I would say it's safe to close this PR.

PR #567, opened 2021-04-21
Update monitor-aws-services.asciidoc

leahleahy:


docs/en/observability/monitor-aws-services.asciidoc
```diff
@@ -3,13 +3,12 @@
 
 [NOTE]
 =====
-If you haven't already, you need to install and configure {metricbeat} to populate
+If you haven't already, please install and configure {metricbeat} to populate
```
EamonnTP, line 6: I would keep the original. It stresses how important it is to install and configure Metricbeat; otherwise you will have no data to analyze.

EamonnTP (changes requested):
Looks good. Minor suggestion.

PR #569, opened 2021-04-21
Update configure-metrics-sources.asciidoc

leahleahy:


docs/en/observability/configure-metrics-sources.asciidoc
```diff
@@ -9,7 +9,7 @@ and container names.
 [[metrics-config-settings]]
 == Override configuration settings
 
-. To access this page, go to *Observability > Metrics*.
+. Go to *Observability > Metrics*.
```
EamonnTP, line 12: I think the original gives the user more context.

PR #573, opened 2021-04-21
Update inspect-uptime-duration-anomalies.asciidoc

leahleahy:


docs/en/observability/inspect-uptime-duration-anomalies.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[inspect-uptime-duration-anomalies]]
 = Inspect uptime duration anomalies
 
-Each monitor location is modelled, and when a monitor runs
+Each monitor location is modeled, and when a monitor runs
```
leahleahy, line 4: This should be modeled, right?

EamonnTP: Good catch. Yes.

PR #574, opened 2021-04-21
Update configure-uptime-settings.asciidoc

leahleahy:


docs/en/observability/configure-uptime-settings.asciidoc
```diff
@@ -6,9 +6,9 @@ by the {uptime-app}, configure alert connectors, and set expiration/age threshol
 for TLS certificates.
 
 Uptime settings apply to the current space only. To segment
-different uptime use cases and domains, use different settings in other spaces.
+different Uptime use cases and domains, use different settings in other spaces.
```
leahleahy, line 9: Same comment about capitalization as the other articles.

EamonnTP: No need to capitalize as we are not talking specifically about the Uptime app.

PR #599, opened 2021-04-29
Improve k8s docs

ChrsMark:
This PR changes documentation of how to run Agent on k8s so as to match the leaderelection approach we have adopted. This is quite similar to what we have for Beats (https://www.elastic.co/guide/en/beats/metricbeat/current/running-on-kubernetes.html)


jsoriano (approved):
:+1:

PR #612, opened 2021-05-03
Add note on how to to build the docs

ruflin:
This should help everyone which is looking for a place to copy the command to build the docs.


README.md
```diff
@@ -24,3 +24,13 @@ In general, we only backport documentation changes to [live stack versions](http
 
 After your PR is merged, we recommend using the [backport tool](https://github.com/sqren/backport) to easily open backport PRs:
 `npx backport`. Alternatively, ping **[@obs-docs](https://github.com/orgs/elastic/teams/obs-docs)** and we'd be happy to handle the backport process for you.
+
+## Build
+
+To build the docs, the `elastic/docs` repository has to be checked out. To build the observabilty docs and open it in the browser, run the following command:
+
+```
+../docs/build_docs --doc ./docs/en/observability/index.asciidoc --chunk 1 --resource ../beats/libbeat/docs --resource ../apm-server/docs/guide --open
```
dedemorton, line 33: Since we have multiple books in the repository now, I wonder if we should consider providing a shell script or something that builds all the docs?

dedemorton: I would list all the books you can build from this repo, the repos that are required for each book, and then show the build commands (and maybe point to the alias file).

ruflin: My goal with this change is mainly to have an example I can copy / paste and adjust for the different books. I took it from my local "notes" as I thought everyone could use this. I like the idea of having all commands listed but I would rather let someone else do this in a follow up PR that knows more about all the books.

README.md
```diff
@@ -24,3 +24,13 @@ In general, we only backport documentation changes to [live stack versions](http
 
 After your PR is merged, we recommend using the [backport tool](https://github.com/sqren/backport) to easily open backport PRs:
 `npx backport`. Alternatively, ping **[@obs-docs](https://github.com/orgs/elastic/teams/obs-docs)** and we'd be happy to handle the backport process for you.
+
+## Build
+
+To build the docs, the `elastic/docs` repository has to be checked out. To build the observabilty docs and open it in the browser, run the following command:
```
dedemorton, line 30:
```suggestion
To build the docs:

. Check out the `elastic/docs` repository, along with any repositories that contain source files.
. Run the `build_docs` script, passing in the path to the `index.asciidoc` and resource paths to other repos that contain source files. For example, to build the Observability Guide and open it in the browser, run:
```

README.md
dedemorton: Fair enough. I think we need to add a bit more guidance, though, so I'll suggest an edit.

ruflin: Thx, merged it in and also resolved conflicts with more recent changes.

dedemorton (approved):
LGTM.

Issue #623, opened 2021-05-05
Document Fleet Server environment variables


dedemorton:
copied over from #449:

Document the environment variables that exist for running Fleet Server.

PR #645, opened 2021-05-11
Get K8s documentation building

dedemorton:
Changes to get the K8s documentation building.

These docs still need more work (for example, there are inconsistencies across the topics and they should be tested for usability).

TODO: 
- [x] Add a note that points to https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-elastic-agent.html (can be done when we clean up inconsistencies in a later PR)


bmorelli25:
I'll test these while working on https://github.com/elastic/observability-docs/issues/623

PR #682, opened 2021-05-20
Clarify no import dependencies can be used with inline scripts

paulb-elastic:
Clarify that dependencies cannot be added via imports with inline scripts.

### DIFF
--- a/docs/en/observability/synthetics-create-test.asciidoc
+++ b/docs/en/observability/synthetics-create-test.asciidoc
@@ -147,7 +147,7 @@ Which option is right for you? That depends.
 If you want to invoke a single journey with few steps, an inline journey might be the easiest workflow --
 you only need to write your tests into your `heartbeat.yml` file.
 
-Note that you can only paste the steps into an inline script, so these cannot be dependent on any components that you may want to `import`. If you do have any dependencies, these would need to be defined with an `import` outside of the `journey` object and run via a test suite.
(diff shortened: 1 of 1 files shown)


docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -140,21 +140,23 @@ journey("Ensure placeholder is correct", ({page}) => {
...
+* Run as an inline journey -- embed into `heartbeat.yml`
 * Run a test suite -- import your tests into Heartbeat
 
 Which option is right for you? That depends.
 If you want to invoke a single journey with few steps, an inline journey might be the easiest workflow --
-you only need paste your tests into your `heartbeat.yml` file.
-if you have a complex test suite, or your tests need to live with your application code,
+you only need to write your tests into your `heartbeat.yml` file.
+
+Note that you can only paste the steps into an inline script, so these cannot be dependent on any components that you may want to `import`. If you do have any dependencies, these would need to be defined with an `import` outside of the `journey` object and run via a test suite.
```
vigneshshanmugam, line 150:
```suggestion
Note that you can only paste the steps into an inline script, so these cannot be dependent on any packages that you may want to `import`. If you do have any dependencies, these would need to be defined with an `import` outside of the `journey` object and run via a test suite.
```

vigneshshanmugam (approved):
LGTM

bmorelli25:
Thanks, Paul! I'll get this backported for you.

PR #693, opened 2021-05-24
Updating Fleet limitations for 7.13

mostlyjason:
Providing feature updates here. We still need to provide updated limit on the number of agents, tracked separately.


mostlyjason:
Wrong branch, closing

PR #712, opened 2021-05-25
Clarify breaking changes for 7.13

dedemorton:
The breaking change about removing agent routes needs to be qualified so it's clear in the title that we're talking about Kibana.

### DIFF
--- a/docs/en/ingest-management/release-notes/release-notes-7.13.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-7.13.asciidoc
@@ -33,7 +33,7 @@ impact to your application.
 
 [discrete]
 [[breaking-97206]]
-.Remove Elastic Agent routes and related services from {kib}
+.Remove Elastic Agent routes and related services from {kib} and implement them in {fleet-server}
 [%collapsible]
 ====
 *Details* +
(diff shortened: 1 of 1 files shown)


EricDavisX:
Dede, should we highlight here where they have moved to?  or is it note the right spot to start talking details about Fleet Server API routes?

docs/en/ingest-management/release-notes/release-notes-7.13.asciidoc
```diff
@@ -33,7 +33,7 @@ impact to your application.
 
 [discrete]
 [[breaking-97206]]
-.Remove Elastic Agent routes and related services
+.Remove Elastic Agent routes and related services from {kib}
```
EricDavisX, line 36:
```suggestion
.Remove Elastic Agent routes and related services from {kib} and implement them in {fleet-server}
```

EricDavisX: I love the clarification - if you think it isn't quite right to extend it a bit further it is ok.  I am just putting the suggestion out there in case the extension feels right.  what do you think? I could go either way, but in this case the slight more detail may help even a bit more.  up to you

dedemorton: Yah, I like your change because it's more descriptive.

EricDavisX (approved):
lgtm;

PR #725, opened 2021-05-26
Logs app: Add index pattern content

EamonnTP:
### Summary

This PR adds content regarding Kibana index patterns that are available for selection in the Logs app settings. 

### Docs preview 

https://observability-docs_725.docs-preview.app.elstc.co/guide/en/observability/master/configure-data-sources.html#edit-config-settings

### DIFF
--- a/docs/en/observability/configure-logs-sources.asciidoc
+++ b/docs/en/observability/configure-logs-sources.asciidoc
@@ -26,8 +26,10 @@ default configuration settings.
 | *Index pattern* | Index pattern or patterns in the {es} indices to read log data from.
 
 Each log source now integrates with {kib} index patterns which support creating and
-querying {ref}/runtime.html[runtime fields]. Instead of entering an index pattern name,
-click *Use {kib} index patterns* and select from a list of existing index patterns.
+querying {kibana-ref}/managing-index-patterns.html[runtime fields].
+
+Instead of entering an index pattern name,
+click *Use {kib} index patterns* and select the `filebeat-*` log index pattern.
 
 | *Fields* | Configuring fields input has been deprecated. You should adjust your indexing using the
 <<logs-app-fields,Logs app fields>>, which use the {ecs-ref}/index.html[Elastic Common Schema (ECS) specification].
--- a/docs/en/observability/configure-logs-sources.asciidoc
+++ b/docs/en/observability/configure-logs-sources.asciidoc
@@ -23,7 +23,11 @@ default configuration settings.
 
 | *Name* | Name of the source configuration. 
 
-| *Indices* | Index pattern or patterns in the {es} indices to read log data from.
+| *Index pattern* | Index pattern or patterns in the {es} indices to read log data from.
+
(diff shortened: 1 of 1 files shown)


docs/en/observability/configure-logs-sources.asciidoc
```diff
@@ -23,7 +23,11 @@ default configuration settings.
 
 | *Name* | Name of the source configuration. 
 
-| *Indices* | Index pattern or patterns in the {es} indices to read log data from.
+| *Index pattern* | Index pattern or patterns in the {es} indices to read log data from.
+
+Each log source now integrates with {kib} index patterns which support creating and
```
bmorelli25, line 28: What are *Kibana index patterns*? This term was added to our "What's New" doc as well, but I can't find any mention of it in the [runtime fields docs](https://www.elastic.co/guide/en/kibana/7.13/managing-index-patterns.html), or anywhere else in the Kibana guide.

EamonnTP:
@bmorelli25 I think the term _Kibana Index Patterns_ relates to the standard index patterns described [here](https://www.elastic.co/guide/en/kibana/current/index-patterns.html). Previously, the Logs UI used Elasticsearch index name patterns. Now the user has a choice to configure a Kibana index pattern instead to incorporate [runtime fields](https://www.elastic.co/guide/en/kibana/current/managing-index-patterns.html#runtime-fields), which is not available with Elasticsearch index name patterns.

Hi @weltenwort Could you help clarify this is little more? Thanks!

weltenwort:
@EamonnTP thanks for taking care of the docs :heart: Your explanation is absolutely on point. We're using the prefix "Kibana" to "index patterns" to differentiate them from the index name patterns (strings such as `filebeat-*`) which the ES docs also call index patterns in some places. The Kibana team is aiming to avoid that ambiguity by renaming Kibana index patterns to "data views" in the next major version: elastic/kibana#94244.

As for the Logs UI, using Kibana index patterns (KIP) will become the preferred style of configuration soon. The user will be prompted to choose a KIP when they visit the Logs UI for the first time. And we'll incorporate a flyout that allows for creation of new KIPs right there without leaving the app.

EamonnTP:
Thanks @weltenwort! Is this doc update sufficient enough or do you feel we need to add more clarification?

bmorelli25 (approved):
Fantastic. Appreciate the explanation @EamonnTP and @weltenwort!

weltenwort:
@EamonnTP how about adding something like

> You can continue to use log sources configured to use an index name pattern such as `filebeat-*` instead of a Kibana index pattern, but some features like those depending on runtime fields might not be available.

Would that be useful info for the user?

EamonnTP:
> @EamonnTP how about adding something like
> 
> > You can continue to use log sources configured to use an index name pattern such as `filebeat-*` instead of a Kibana index pattern, but some features like those depending on runtime fields might not be available.
> 
> Would that be useful info for the user?

Thanks @weltenwort I added the additional context.

PR #742, opened 2021-05-27
Add note about removing system extension during uninstall

dedemorton:
@EricDavisX  It sounds like this note would be useful to macOS users, so I've added it back in.

Do we want to say where the extension lives?

### DIFF
--- a/docs/en/ingest-management/elastic-agent/uninstall-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/uninstall-elastic-agent.asciidoc
@@ -51,10 +51,10 @@ files are installed? See <<installation-layout>>.
 installed for endpoint protection. The directory structure is similar to {agent},
 for example, `/Library/Elastic/Endpoint/*`.
 +
-//TODO: Add to docs when this is confirmed.
-//NOTE: When you remove the Elastic Endpoint integration from a macOS host, the
-//Endpoint System Extension is left on disk intentionally. If you want to remove
-//the extension, refer to the documentation for your operating sytem.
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/elastic-agent/uninstall-elastic-agent.asciidoc
```diff
@@ -51,10 +51,10 @@ files are installed? See <<installation-layout>>.
 installed for endpoint protection. The directory structure is similar to {agent},
 for example, `/Library/Elastic/Endpoint/*`.
 +
-//TODO: Add to docs when this is confirmed.
-//NOTE: When you remove the Elastic Endpoint integration from a macOS host, the
-//Endpoint System Extension is left on disk intentionally. If you want to remove
-//the extension, refer to the documentation for your operating sytem.
+NOTE: When you remove the Elastic Endpoint integration from a macOS host
```
EamonnTP, line 54:
```suggestion
NOTE: When you remove the {elastic-endpoint} integration from a macOS host
```

EamonnTP (approved):
LGTM. It looks like there is an attribute for Elastic Endpoint.

docs/en/ingest-management/elastic-agent/uninstall-elastic-agent.asciidoc
dedemorton: thanks, I noticed the security docs weren't using the attribute, so I wondered if I should. Probably doesn't hurt, tho!

EricDavisX (approved):
i like it as is, let's leave macOS system admin alone links + file locations alone for now.  lgtm;

PR #797, opened 2021-06-17
Fix formatting error caused by addition of tagged region

dedemorton:
Fixes tagged region that is causing page formatting to break.


EamonnTP (approved):
LGTM!

PR #799, opened 2021-06-18
Remove docs for managed agent on k8s

ChrsMark:
This PR removes section for managed elastic-agent on k8s since we don't currently officially support it.
K8s manifests are also being removed: https://github.com/elastic/beats/pull/26368

### DIFF
--- a/docs/en/ingest-management/elastic-agent/elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent.asciidoc
@@ -16,7 +16,7 @@ To learn how to install, configure, and run your {agent}s, see:
 * <<elastic-agent-installation>>
 * <<uninstall-elastic-agent>>
 * <<run-elastic-agent-standalone>>
-* <<running-on-kubernetes>>
+//* <<running-on-kubernetes>>
 * <<running-on-kubernetes-standalone>>
 * <<upgrade-elastic-agent>>
 * <<start-elastic-agent>>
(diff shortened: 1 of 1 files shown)


ruflin:
I wonder if we should fully remove it or instead mainly remove the links to the manifest files and mention that it is not fully supported yet. But we have already questions on discuss and users will run it on k8s, so some guidance would be good.

ChrsMark:
@ruflin I think that providing guidance without an actual manifest would be incomplete. Also I have not run the case yet so I'm not sure what kind of guidance we could provide here. 

We can prioritise the work for https://github.com/elastic/beats/issues/25237 even in a primitive stage so as to have something complete soon. We also need this for ourselves when we test k8s integration/package manually in managed mode.

ruflin:
SGTM

dedemorton:
Can we just remove this from the documentation build rather than deleting the file completely? 

To do that, you just need to comment out lines 19 and 34

```
* <<running-on-kubernetes>>
...

include::running-on-kubernetes.asciidoc[leveloffset=+1]
```

If you think it's better to remove the file completely,  then make sure you also remove these lines ^^ to fix the doc build.

ChrsMark:
@dedemorton what you proposed sounds better so I applied your approach, thanks!

ChrsMark:
@dedemorton can we have this in soon since the manifests have already been removed?

dedemorton (approved):
LGTM! Sorry about the delay. I fell down a TLS rabbit hole.

dedemorton:
I'm going to take the liberty of merging and backporting this.

PR #816, opened 2021-06-28
[Uptime] Update TLS alert documentation to reflect current action variable image

### LINKED ISSUE #815: [Uptime] Update TLS certificate alert documentation to reflect changes to TLS alert logic
In 7.14, TLS Alerts will be moved from a batched model, where one multiple expiring or aging certificates only triggered 1 action, to an individual model where each aging or expired certificate triggers its own action.

The current document is general enough that it does not appear to need a significant amount of updates. However, the action variables section is now out of date. The current action variables screenshot should be updated to reflect the below message.
Screen Shot 2021-06-28 at 12 20 15 PM

dominiqueclarke:
Resolves #815 

Following elastic/kibana#102138, this PR updates the [action variables image](https://github.com/elastic/observability-docs/blob/master/docs/en/observability/uptime-tls-alert.asciidoc#action-variables) used in the TLS Alert documentation.


dedemorton (approved):
LGTM! Thanks for doing this.

PR #822, opened 2021-06-30
Fix uninstall commands

### LINKED ISSUE #724: Agent uninstall command is wrong on macOS and Linux
This page https://www.elastic.co/guide/en/fleet/current/uninstall-elastic-agent.html says to uninstall Agent with the command `sudo elastic-agent uninstall`. That won't work for users because the directory Agent is installed into isn't in (most) people's `PATH`. The command should be a full path command, e.g. `sudo /Library/Elastic/Agent/elastic-agent uninstall` (macOS) or `sudo /opt/Elastic/Agent/elastic-agent uninstall` (Linux).

cc @ph @gogochan

dedemorton:
Closes #724 

This should be backported to 7.13.


EamonnTP (approved):
Looks great.

PR #890, opened 2021-07-26
Add note about avoid using leaderelection in large clusters

ChrsMark:
This PR adds a note about avoid using leaderelection in large clusters and instead consider using a Deployment in addition to the Daemonset so as to specify different resources according to the cluster's needs.

We have a similar note for Beats too: https://www.elastic.co/guide/en/beats/metricbeat/7.13/running-on-kubernetes.html#_deploying_metricbeat_to_collect_cluster_level_metrics_in_large_clusters


MichaelKatsoulis (approved):

ChrsMark:
@Mergifyio backport 7.16

PR #898, opened 2021-07-27
example for resetting a package to bring back the dashboard assets

EricDavisX:
Just an idea we decided might be nice to have doc'ed after a brief conversation in slack:

https://elastic.slack.com/archives/CNMNXV4RG/p1627384694155700

### DIFF
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -247,5 +247,4 @@ POST /api/fleet/epm/packages/system-1.0.0
 { "force": true }
 ----
 
-The package version is shown in the Integrations view in {kib} or can be ascertained 
-using the https://epr.elastic.co/search?package=system[registry] 
+The package version is shown in the Integrations view in {kib}.
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -231,7 +231,8 @@ are both accessible; by default these are `5601` and `9200` respectively.
 [[how-do-i-reset-a-missing-dashboard-asset]]
 == If I delete an integration dashboard asset from {kib}, how do I get it back?
 
-To reinstall the assets for a specific integration, you can use the API.
+To reinstall the assets for a specific integration, you can use the UI for most cases
+and the {fleet} API for those not allowed in the UI.
 You'll need to know the package name and version you want to reinstall.
 To reinstall package assets, execute the following call with the `force` parameter in the body:
 
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -240,8 +240,11 @@ POST /api/fleet/epm/packages/[package name]-[package version]
 { "force": true }
 
 So, for example, to reset the system v1.0.0 package you could POST:
+[source,console]
+----
 POST /api/fleet/epm/packages/system-1.0.0
 { "force": true }
+----
 
 The package version is shown in the Integrations view in {kib} or can be ascertained 
 using the https://epr.elastic.co/search?package=system[registry] 
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -232,7 +232,7 @@ are both accessible; by default these are `5601` and `9200` respectively.
 == If I delete an integration dashboard asset from {kib}, how do I get it back?
 
 The aseets for a given integration can be re-loaded into {es} with the API.
-You'll need to know the name of the package and the version you want to reinstall.
+You'll need to know the package name and version you want to reinstall.
 To reinstall package assets, execute the following call with the `force` parameter in the body:
 
 [source,console]
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -239,7 +239,7 @@ in the body:
 POST /api/fleet/epm/packages/[package name]-[package version]
 { "force": true }
 
-So, for example, to reset the system v1.0.0 package you could POST:
+So, for example, to reinstall the system v1.0.0 package, POST:
 [source,console]
 ----
 POST /api/fleet/epm/packages/system-1.0.0
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -231,7 +231,7 @@ are both accessible; by default these are `5601` and `9200` respectively.
 [[how-do-i-reset-a-missing-dashboard-asset]]
 == If I delete an integration dashboard asset from {kib}, how do I get it back?
 
-The aseets for a given integration can be re-loaded into {es} with the API.
+To reinstall the assets for a specific integration, you can use the API.
 You'll need to know the package name and version you want to reinstall.
 To reinstall package assets, execute the following call with the `force` parameter in the body:
 
--- a/docs/en/ingest-management/faq.asciidoc
+++ b/docs/en/ingest-management/faq.asciidoc
@@ -207,7 +207,6 @@ same way as other integrations. Try it out! For more information, see the
 //REVIEWERS: Unsure here whether we want to say Elastic Endpoint or Elastic
 //Security. ^^
 
-
 [discrete]
 [[how-are-agent-kibana-communications-secured]]
 == How are communications secured between {agent} and {kib}?
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
 you can copy the {es} endpoint URL from the overview page of your deployment.
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete a {kib} dashboard asset, how do I get it back?
+
+The aseets for a given integraion can be re-loaded into {es} with the API.
+You'll need to know the name of the package and the version you want to reinstall.
+Use cURL or your favorite API UI tool to run the following with the noted 'force' param
+in the body:
+
+POST /api/fleet/epm/packages/[package nmae]-[package version]
+{ "force": true }
+
+So, for example, to reset the system v1.0.0 package you could POST:
+POST /api/fleet/epm/packages/system-1.0.0
+{ "force": true }
+
+The package version is shown in the Integrations view in {kib} or can be ascertained 
+using the https://epr.elastic.co/search?package=system[registry] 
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
 you can copy the {es} endpoint URL from the overview page of your deployment.
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete a {kib} dashboard asset, how do I get it back?
+
+The aseets for a given integraion can be re-loaded into {es} with the API.
```
EricDavisX, line 234:
```suggestion
The aseets for a given integration can be re-loaded into {es} with the API.
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
...
+You'll need to know the name of the package and the version you want to reinstall.
+Use cURL or your favorite API UI tool to run the following with the noted 'force' param
```
EricDavisX, line 236:
```suggestion
Use cURL or your favorite API UI tool to execute the following call with the noted "force" parameter
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
...
+in the body:
+
+POST /api/fleet/epm/packages/[package nmae]-[package version]
```
EricDavisX, line 239:
```suggestion
POST /api/fleet/epm/packages/[package name]-[package version]
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
 you can copy the {es} endpoint URL from the overview page of your deployment.
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete a {kib} dashboard asset, how do I get it back?
```
EricDavisX, line 232:
```suggestion
== If I delete an integration dashboard asset from {kib}, how do I get it back?
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
 you can copy the {es} endpoint URL from the overview page of your deployment.
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete an integration dashboard asset from {kib}, how do I get it back?
+
+The aseets for a given integration can be re-loaded into {es} with the API.
```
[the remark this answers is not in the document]
dedemorton: Note we use different verbs here to mean the same thing in this topic: reload, reinstall, and reset. We should use one.

dedemorton (commented):
some editorial feedback.

docs/en/ingest-management/faq.asciidoc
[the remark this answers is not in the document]
EricDavisX: you have a good eye for that nuance @dedemorton thank you.  And YES - you can do this from the UI for any Integration that is allowed to be removed.  Only a few special ones are NOT allowed, I think System, Endpoint, and Fleet Server (when it is on a hosted Agent policy).  Should we add a note here that simply adding and removing it from the UI can do the same thing, if it is available to do so?

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
...
+
+The aseets for a given integraion can be re-loaded into {es} with the API.
+You'll need to know the name of the package and the version you want to reinstall.
+Use cURL or your favorite API UI tool to run the following with the noted 'force' param
+in the body:
+
+POST /api/fleet/epm/packages/[package nmae]-[package version]
+{ "force": true }
+
+So, for example, to reset the system v1.0.0 package you could POST:
```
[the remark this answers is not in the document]
EricDavisX: I did like using the other words in case someone was searching for them - and Jen had used the term 'reset' in slack so I wanted it to be here.  I'm ok with whatever you feel is best.

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,22 @@ and port for your setup. If you run everything locally, the address is
...
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete an integration dashboard asset from {kib}, how do I get it back?
+
+The aseets for a given integration can be re-loaded into {es} with the API.
+You'll need to know the name of the package and the version you want to reinstall.
+Use cURL or your favorite API UI tool to execute the following call with the noted "force" parameter
```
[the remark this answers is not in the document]
EricDavisX: I don't know if the Kibana Dev console can send API commands to Kibana itself?  I'm not as versed with it so it didn't occur to me to suggest it, if it is a thing.

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,27 @@ and port for your setup. If you run everything locally, the address is
 you can copy the {es} endpoint URL from the overview page of your deployment.
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete an integration dashboard asset from {kib}, how do I get it back?
+
+To reinstall the assets for a specific integration, you can use the API.
```
EricDavisX, line 234:
```suggestion
To reinstall the assets for a specific integration, you can use the UI for most cases
and the {fleet} API for those not allowed in the UI.
```

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,27 @@ and port for your setup. If you run everything locally, the address is
...
+You'll need to know the package name and version you want to reinstall.
```
EricDavisX, line 235:
```suggestion
In order to use the API, you'll need to know the package name and version
you want to reinstall.
```

docs/en/ingest-management/faq.asciidoc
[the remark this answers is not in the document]
dedemorton: I see what you're saying. Inconsistencies add ambiguity, though, because users wonder, "Do all these words mean the same thing?"

Let's try to use the verb that is the most descriptive of what the user is trying to do. Hopefully that is the same term that users are likely to search on and the same term we use throughout the docs!

EricDavisX:
@dedemorton let me know what you think about those last 2 suggestions I put in, thanks!

docs/en/ingest-management/faq.asciidoc
dedemorton: I think the introductory phrase is already implied, but feel free to commit if you think it makes things more clear.

docs/en/ingest-management/faq.asciidoc
dedemorton: I think that's OK, but I wonder if we should mention where in the UI. We haven't really documented the Integrations UI very well yet. Maybe that's just something we need to add later?

EricDavisX: I'll submit that the move to make Integrations a top level (left hand) Kibana navigation item means that everyone in the world should find it.  We can leave it out for now and if we are proven wrong, then we can add more pointers in for it.

docs/en/ingest-management/faq.asciidoc
EricDavisX: it only flows right to me, based on the addition of the other.  I like it. Will do, thanks.

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,29 @@ and port for your setup. If you run everything locally, the address is
...
+
+So, for example, to reinstall the system v1.0.0 package, POST:
+[source,console]
+----
+POST /api/fleet/epm/packages/system-1.0.0
+{ "force": true }
+----
+
+The package version is shown in the Integrations view in {kib} or can be ascertained 
+using the https://epr.elastic.co/search?package=system[registry]
```
jen-huang, line 254:
```suggestion
The package version is shown in the Integrations view in {kib}.
```

do we reference epr.elastic.co anywhere else in docs? if not, I think we should hide this complexity from the user

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,29 @@ and port for your setup. If you run everything locally, the address is
...
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
+== If I delete an integration dashboard asset from {kib}, how do I get it back?
+
+To reinstall the assets for a specific integration, you can use the UI for most cases
+and the {fleet} API for those not allowed in the UI.
+In order to use the API, you'll need to know the package name and version
+you want to reinstall.
```
jen-huang, line 237:
```suggestion
To reinstall the assets for a specific integration, you can use the {fleet} API using the package name and version.
```
we don't offer a "reinstall" UI, the closest is uninstall + install, but we don't allow uninstall when the package is in use in a policy so this probably won't work for most cases. what do you think of removing the part about UI and just reference the API?

EricDavisX: ugh, you're right.

docs/en/ingest-management/faq.asciidoc
EricDavisX: I was thinking we need to start exposing it somewhere, if it isn't already.  I was trying to sneak it in, but it isn't really needed for this topic.  I could start a new one, but first I can review if we have it exposed or not, and we can figure out if we want to start with this (we may not, as you note).

docs/en/ingest-management/faq.asciidoc
```diff
@@ -227,3 +226,25 @@ and port for your setup. If you run everything locally, the address is
 you can copy the {es} endpoint URL from the overview page of your deployment.
 If you're not running in {ecloud}, make sure the {kib} and {es} HTTPS ports
 are both accessible; by default these are `5601` and `9200` respectively.
+
+[discrete]
+[[how-do-i-reset-a-missing-dashboard-asset]]
```
EricDavisX, line 231:
```suggestion
[[how-do-i-reinstall-a-missing-dashboard-asset]]
```

EricDavisX:
A'ight.  changes in, its ready for re-re-review

jen-huang (approved):
Thanks for adding this!

PR #902, opened 2021-07-28
Add info about --insecure flag to agent install docs

### LINKED ISSUE #802: Include --insecure flag in the Agent docs
We've had several users run into certificate problems and get stuck. 
When a user follows the in-product instructions to add a Fleet Server in 7.13/14 in a self-managed cluster, they must use the `--insecure` flag when enrolling agents or they will get an error message https://github.com/elastic/beats/issues/26365. 

The current documentation does not describe adding this flag https://www.elastic.co/guide/en/fleet/current/elastic-agent-installation.html#install-fleet-managed-agent.

Can we document how to add this flag and what implications it has for their security?

Can we also include that error message in our troubleshooting guide along with guidance on how to resolve it?

dedemorton:
Closes #802

Please respond to any questions embedded as comments in the source files.

### DIFF
--- a/docs/en/ingest-management/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting.asciidoc
@@ -79,6 +79,41 @@ One of the requests to the {fleet} API will most likely have returned an error. 
 message doesn't give you enough information to fix the problem, please contact us in the {forum}[discuss forum].
 
 [discrete]
+[[agent-enrollment-certs]]
+== {agent} enrollment fails on the host with `x509: certificate signed by unknown authority` message
+
+To ensure that communication with {fleet-server} is encrypted,
+{fleet-server} requires {agent}s to present a signed certificate. In a
+self-managed cluster, if you don't specify certificates when you set up
+{fleet-server}, self-signed certificates are generated automatically.
+
+If you attempt to enroll an {agent} in a {fleet-server} with a self-signed
+certificate, you will encounter the following error:
+
+[source,sh]
+----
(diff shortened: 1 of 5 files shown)


docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -308,10 +343,13 @@ agent you want to unenroll.
 [[ca-cert-testing]]
 == On {fleet-server} startup, ERROR seen with `State changed to CRASHED: exited with code: 1`
 
+//QUESTION: Is this section still true? It does appear that we get a better error now,
+//plus it seems that users are overlooking this section. Should I keep this section or remove it?
```
mostlyjason, line 347: What makes you think they are overlooking it and why? If that is the only error its not useful without more information, so having this could still be helpful. I'm not sure if its still true or not.

mostlyjason (approved):

docs/en/ingest-management/troubleshooting.asciidoc
dedemorton: OK, I just wasn't sure if this problem would still result in a crash. Sounds like it does, so I will keep it. I think users who are looking for the certs-related error do not find this section, but it's OK to cover the details in both. I will leave as-is.

docs/en/ingest-management/troubleshooting.asciidoc
```diff
@@ -308,10 +343,13 @@ agent you want to unenroll.
...
+
```
dedemorton, line 348:
```suggestion
```

Issue #856, opened 2021-07-13
Update limitations for GA release


mostlyjason:
We should retire our limitations page for the GA release https://www.elastic.co/guide/en/fleet/current/fleet-limitations.html. This was originally created for beta, and we don't want to use negative language in the GA release. 

We'll replace it a new page called "Beats vs Elastic Agent comparison". This will allow us to clearly define what is supported with more detail so we don't unintentionally create breaking changes down the road. It will help net new users select which shipper to use, and help users migrating from Beats to Agent. We'll redirect from the limitations page to this one.

- Overview
  - Compare features between Beats/Agent standalone/Agents enrolled in Fleet
  - Clarify its comparing Beats as a standalone shipper, only include Beats that Agent supports
- Inputs
  - Do we plan to deprecate any?
  - What about ones that are not used in any integration?
- Outputs
  - not redis, file, console
- Top level configuration settings
  - Allowlist of ones we definitely support
  - Processors
  - Autodiscover
  - General settings <- deprecated?
  - Don’t define individual input / output settings right now
- Stack feature compatibility (CCS, spaces)
- Other commonly expected features

Next Steps: 
- [x] Write content in a spreadsheet @nimarezainia & @mostlyjason 
- [x] Convert to a doc page @dedemorton 
- [x] Identify a new home for limitations on existing pages, if they don't fit on the comparison page
  - [x] Requires internet access for Kibana to download integration packages from the Elastic Package Registry 
  - [x] Elastic Agent requires internet access to perform binary upgrades from Fleet 
  - [ ] ~On Elastic Cloud, there’s a limit to the number of enrolled Elastic Agents that a single Kibana instance can handle~ removing from this list because scaling guidelines should be covered by https://github.com/elastic/observability-docs/issues/523

mostlyjason:
@nimarezainia can you help collect a list of inputs, outputs and config settings that are supported in beats but not supported in agent? I believe you are already tracking that internally?

nimarezainia:
Document out for review.

dedemorton:
Adding this as a reminder. 

Assuming we want to the page to have this URL:
https://www.elastic.co/guide/en/fleet/7.14/fleet-agent-comparison.html

The page ID needs to be `[[fleet-agent-comparison]]`

Issue #523, opened 2021-04-15
Fleet Server scaling and deployment guidelines


mostlyjason:
We need a guide for users who are responsible for operating Fleet Server. It should describe:

- [x] how to observe Fleet Server and know when to scale it https://github.com/elastic/fleet-server/issues/173
- [x] how to scale it https://github.com/elastic/fleet-server/issues/185
- [x] scaling limits https://github.com/elastic/fleet-server/pull/240
- [x] scaling guidance for ESS instance types https://github.com/elastic/fleet-server/issues/134

dedemorton:
Another related issue: https://github.com/elastic/fleet-server/issues/416

EricDavisX:
We decided to close out the above issue and document the nuance that we expect all Fleet Servers in production to use full TLS and that we don't support a mixed environment.

dedemorton:
Closed by https://github.com/elastic/observability-docs/pull/1061

PR #947, opened 2021-08-02
Remove limitations topic from Fleet and Agent docs

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/856

Removes limitations topic. All relevant info will be moved to the appropriate section in the docs.

TODO: 
- [x] Force the PR build to break to identify broken links across the library.
- [x] ~Open PRs in relevant repos to fix broken links~
- [x] Create redirect page that redirects users to the beats/agent comparison topic
- [x] Move relevant limitations to other locations in the documentation
- [x] Figure out what to do with these limitations. Still valid? If so, where should they go:
  - [x]  In order to use Fleet, the Elastic Agents must have a direct network connection to Kibana and Elasticsearch. It is also possible to run the Elastic Agents in standalone mode in cases where a network connection is not available or not needed.
  - [x] "Fleet is currently only available to users with the superuser role. This role is necessary to create indices, install integration assets, and update Elastic Agent policies."

TODO (after merging):
- [ ] Create redirect request with the web team that goes live on release day
- [ ] Remove redirect page after web team sets up redirects

NOTE: I have not added scaling guidelines to this PR. They should be part of the work Nima is doing on this issue: https://github.com/elastic/observability-docs/issues/523


bmorelli25 (approved):
Looks good.

Both of these might fit into the prereq section on this page https://www.elastic.co/guide/en/fleet/current/elastic-agent-installation.html#install-fleet-managed-agent?

* In order to use Fleet, the Elastic Agents must have a direct network connection to Kibana and Elasticsearch. It is also possible to run the Elastic Agents in standalone mode in cases where a network connection is not available or not needed.
* "Fleet is currently only available to users with the superuser role. This role is necessary to create indices, install integration assets, and update Elastic Agent policies."

dedemorton:
> Both of these might fit into the prereq section on this page

Agreed. I think the requirements have changed though. Agent connects to Fleet Server now and can use a proxy server (or will be able to do so after a bug is fixed). Also I can't remember if superuser role is still required for Fleet. I'll wait for @mostlyjason to confirm.

mostlyjason:
@dedemorton looks good. I noticed it doesn't actually remove the limitations page, just adds a redirect. Is that intentional? Also, yes a superuser role is still required.

dedemorton:
@mostlyjason Yes, usually we set up a redirect when possible for moved or deleted pages to avoid 404s. If you think a redirect is confusing, I can remove it. Just let me know.

EricDavisX:
we are pulling out support for Proxy for 7.14 - does that make any impact?  we aren't sure it will be fixed in 7.14.1 or 7.15.

And, about the 'direct connection' statment:
 In order to use Fleet, the Elastic Agents must have a direct network connection to Kibana and Elasticsearch.
shouldn't it rather just be the below?
 In order to use Fleet, the Elastic Agents must have a direct network connection to Fleet Server and Elasticsearch.

dedemorton:
@EricDavisX 

I've added this to the isntall topic: In order to use Fleet, the Elastic Agents must have a direct network connection to Fleet Server and Elasticsearch.

Do you think it also needs to be added to the getting started topics, or are the setup instructions clear enough? My concern is that the list of prereqs is getting quite long, and new users won't know what elastic agents and fleet server are until they run through the steps.

EricDavisX:
I think that is great - and I agree with your thinking about pre-reqs

mostlyjason:
Thanks for adding that to the install topic. Redirect should be fine. Can we publish this now?

Side note: Another option is to mention the requirement for connectivity to Fleet and ES later in the getting started instructions, like when the user is setting up the agent?

mostlyjason (approved):

PR #1020, opened 2021-09-02
Update Fleet Server docs with link to related ECE info

### LINKED ISSUE #825: Document certs and DNS requirements for using ECE with Fleet Server
In 2.10, ECE users will have to manually update their TLS certificates before they can use Fleet Server. The same will be true for APM in ECE 3.0. A wildcard certificate is required to create multiple deployments on demand in the format $id.$host_root. However, adding fleet as an extra wildcard like *.*.$host_root is not possible. So users have to manually and statically add the $domain part like *.fleet.$host_root. 

They might also need to add DNS infrastructure similarly to what they need to do for aliases now.

dedemorton:
Closes #825 

The cloud updates were merged today, so I've added a link to the docs that will provide more info for ECE users.


EamonnTP (approved):
LGTM!

Issue #878, opened 2021-07-20
Create Integrations docs page


mostlyjason:
Create a top-level documentation page for Fleet Integrations. Move integration policy content and data streams under it. Document that access to EPR is a requirement.

Related:
- https://github.com/elastic/observability-docs/issues/513

dedemorton:
This issue sounds mostly organizational and would be part of the effort described here: https://github.com/elastic/observability-docs/issues/784

dedemorton:
This would also be a good place to our hosted integrations documentation.

bmorelli25:
Closed in #1029.

PR #1029, opened 2021-09-04
Add integrations page and document how to upgrade integrations

### LINKED ISSUE #900: [Request] Document how users can upgrade integrations and integration policies
## Description
We'd like to document how integrations are upgraded, both automatically and manually by users on the integrations settings page. (Note: this requirement may be related to https://github.com/elastic/observability-docs/issues/878)

Also, we'd like to document the options to upgrade integration policies. Describe how this works in the case where the integration is automatically upgraded and manually. Also, explain how it works when the upgrade can be completed automatically, or when the user needs to manually resolve conflicts. See more information here: https://github.com/elastic/kibana/issues/106048

## Collaboration

* [x] The docs team will lead producing the content
* [ ] The product team will provide the initial content and the docs team will edit / review
* [ ] The docs team will define with product team the structure and location, and the product team will provide the initial content
* [ ] Other (please describe)

## Suggested Target Release

7.15 since we'll be releasing a new integration policy upgrade feature 

## Stakeholders
@jen-huang @mostlyjason @kpollich

dedemorton:
Closes #900 and #878.

Preview link: https://observability-docs_1029.docs-preview.app.elstc.co/guide/en/fleet/master/integrations.html 

There are a few questions inline in the review, plus I have a few more here. Please respond by adding a comment to the review  or discussion:

- [x] @bmorelli25 I’ve duplicated some of the info that’s currently in https://docs.elastic.co/en/integrations. Do you think it works OK? Should the exact same content live in both places? 

- [x] Is this statement still true? “When Elastic releases a new Integration, it shows up in the Integrations listing in {kib} after a restart (like during a stack upgrade).” If so, do we want to document the API call used to trigger a refresh in Kibana?

- [x] Was I right to remove all the old/manual steps about updating integrations and rolling out the updates. Seems like the new functionality pretty much covers that. 

TODO after merging:
- [ ] Fix typos in the policies documentation where we say "policy integration" instead of "integration policy"


docs/en/ingest-management/integrations.asciidoc
```diff
@@ -0,0 +1,208 @@
...
+[[view-integrations]]
+== View integration policies
+
+An integration policy is created when you add an integration to an {agent}
+policy.
+
+To view details about all the integration policies for a specific integration:
+
+. In {kib}, go to *Management > Integrations > Manage*. Search for and select
+the integration you want view.
```
bmorelli25, line 89:
```suggestion
the integration you want to view.
```

docs/en/ingest-management/integrations.asciidoc
```diff
@@ -0,0 +1,208 @@
...
+[[view-integration-assets]]
+== View integration assets
+
+Integrations come with a number of assets, such as dashboards, saved searches,
+and visualizations for analyzing data.
+
+To view integration assets:
+
+. In {kib}, go to *Management > Integrations > Manage*. Search for and select
+the integration you want view.
```
bmorelli25, line 126:
```suggestion
the integration you want to view.
```

bmorelli25 (commented):
> @bmorelli25 I’ve duplicated some of the info that’s currently in https://docs.elastic.co/en/integrations. Do you think it works OK? Should the exact same content live in both places?

Yes, I think that's fine. There will be a fair amount of duplication between the two for at least the foreseeable future.

docs/en/ingest-management/integrations.asciidoc
```diff
@@ -0,0 +1,208 @@
...
+
+[discrete]
+[[add-integration-under-integrations]]
+== Add an integration to a policy
+
+Policies consist of one or more integrations. To add a new integration to a
+policy:
+
+//QUESTION: A couple of these tasks are also covered under "Policies" but I
+//think it makes sense to cover them here, too, since the steps are different. WDYT?
```
bmorelli25, line 57: ++ to include them here too

docs/en/ingest-management/integrations.asciidoc
```diff
@@ -0,0 +1,208 @@
...
+
```
dedemorton, line 58:
```suggestion
```

EricDavisX:
I added @jen-huang as I don't know the feature actually very well.  We could ask the test team to review it?  they know this stuff better than me for sure.  :)

dedemorton:
@joshdover Can you review this PR and respond to questions (or delegate)? This PR is blocked from merging until we get at least one developer to review it.

kpollich (approved):
The new documentation around updating integrations and upgrading policies looks good to me. Thanks for your work here, DeDe!

>  Was I right to remove all the old/manual steps about updating integrations and rolling out the updates. Seems like the new functionality pretty much covers that.

I think this is a bit of a gray area. It might still be useful to include the previous documentation about deleting and re-rolling out an integration as a troubleshooting method. For example, when a user attempts an upgrade of a policy that is many versions old, there may be substantial conflicts or configuration issues that would warrant "delete -> rollout" workflow as we suggested previously.

joshdover:
> I think this is a bit of a gray area. It might still be useful to include the previous documentation about deleting and re-rolling out an integration as a troubleshooting method. For example, when a user attempts an upgrade of a policy that is many versions old, there may be substantial conflicts or configuration issues that would warrant the "delete -> rollout" workflow as we suggested previously.

+1 on keeping the delete + rollout workaround, but we probably shouldn't highlight it in the happy path docs. Maybe we should move this to the "Troubleshooting common problems" section with a link from this new updating section?

dedemorton:
@joshdover OK, I think this is ready to be merged. I move the old upgrade steps to the troubleshooting topic as you requested. 

I do have one question left that nobody answered:

* Is this statement still true? “When Elastic releases a new Integration, it shows up in the Integrations listing in {kib} after a restart (like during a stack upgrade).” If so, do we want to document the API call used to trigger a refresh in Kibana?

We need to tell users how and when the upgraded integrations show up in Kibana. Is it after restarting Kibana? Upgrading to a new stack version? Calling the API?

kpollich:
> * Is this statement still true? “When Elastic releases a new Integration, it shows up in the Integrations listing in {kib} after a restart (like during a stack upgrade).” If so, do we want to document the API call used to trigger a refresh in Kibana?
> 
> We need to tell users how and when the new integration versions show up in Kibana. Is it after restarting Kibana? Upgrading to a new stack version? Calling the API?

I'm not sure about the specifics here. In my testing, it seems like we will always fetch the latest version of an integration when checking for updates, regardless of the cache status. We do cache integration data, so restarting Kibana will always trigger a "cache bust" and ensure all integration data is fresh. 

@nchaulet - Do you have any insight here? Do we need to document that a Kibana restart is required to ensure up-to-date integration data?

nchaulet:
> @nchaulet - Do you have any insight here? Do we need to document that a Kibana restart is required to ensure up-to-date integration data? 

I think we do not cache the calls to EPR so there is no need for a restart, if a new package version is available it will be displayed.

joshdover (approved):
LGTM, thanks for your help here!

PR #1031, opened 2021-09-07
Update beats-agent-comparison.asciidoc

mostlyjason:
Updating description for managed by Fleet mode on the Beats/Agent capabilities page.


dedemorton (approved):
LGTM, but I wonder if we can just drop the "Once installed" phrase (in both instances).

The phrase is grammatically ambiguous/repetitive and doesn't add much (it's technically a misplaced modifier). The main difference we want to highlight is between fleet managing policies vs users managing them manually. 

I'll approve because I think only word nerds will care about this. You can change or leave as-is.

mostlyjason:
@dedemorton thanks I removed the once installed phrases

Issue #1018, opened 2021-09-02
[Uptime] Add obs-docs page on mapping errors


andrewvc:
The kibana feature in https://github.com/elastic/kibana/pull/110857 needs a page in the docs explaining mapping errors and how to fix them.

PR #1046, opened 2021-09-08
[Uptime] Add troubleshooting page for Uptime guide

justinkambic:
Resolves https://github.com/elastic/observability-docs/issues/1018.

The goal of this PR is to introduce a troubleshooting page we can link to from the Uptime app in certain cases. Today, the only issue we have is this mapping problem.

Our goal is to link to this docs page from the Uptime UI when this issue occurs.

### DIFF
--- a/docs/en/observability/troubleshoot-uptime.asciidoc
+++ b/docs/en/observability/troubleshoot-uptime.asciidoc
@@ -11,7 +11,7 @@ through an intermediary such as {ls} without the `setup` command being run.
 To fix this problem, you typically need to remove your Heartbeat indices, and create
 new ones, with the appropriate mappings installed. To achieve this, follow the steps below.
 
-=== Step 1: Delete your Heartbeat indices
+=== Step 1: Delete your {heartbeat} indices
 
 Because we need to ensure the mapping is appropriately applied to all {heartbeat} data going forward,
 we must delete all {heartbeat} indicies that match the pattern the {uptime-app} will use.
--- a/docs/en/observability/troubleshoot-uptime.asciidoc
+++ b/docs/en/observability/troubleshoot-uptime.asciidoc
@@ -4,7 +4,7 @@
 == Mapping issues
 
 There are situations in which {heartbeat} data can be indexed without the correct mappings applied.
-These situations cannot occur with the {elastic-agent} configured via fleet, only with standalone {heartbeat}.
+These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat}.
 This can occur when the underlying `heartbeat-VERSION` ILM alias is deleted manually or when {heartbeat} writes data
 through an intermediary such as {ls} without the `setup` command being run.
 
--- a/docs/en/observability/troubleshoot-uptime.asciidoc
+++ b/docs/en/observability/troubleshoot-uptime.asciidoc
@@ -35,8 +35,7 @@ NOTE: For more information on how to use this command, you can reference the
 ./heartbeat setup -e 
 ----
 
-This will perform the necessary startup tasks and ensure that your indices will have the appropriate
-mapping going forward.
+This command performs the necessary startup tasks and ensures that your indicies have the appropriate mapping going forward.
 
 === Step 3: Run {heartbeat} again
 
--- a/docs/en/observability/troubleshoot-uptime.asciidoc
+++ b/docs/en/observability/troubleshoot-uptime.asciidoc
@@ -1,0 +1,33 @@
+[[troubleshoot-uptime]]
+= Uptime troubleshooting
+
+== Mapping issues
+
+It is possible for Heartbeat to create an index without the appropriate mappings.
+When this happens, the {uptime-app} will be unable to get the data it requires; in this
+state, the UI cannot function.
+
+To fix this problem, you typically need to remove your Heartbeat indices, and create
+new ones, with the appropriate mappings installed. To achieve this, follow the steps below.
+
+=== Step 1: Delete your Heartbeat indices
+
+["source","sh"]
+----
+curl -u {USERNAME}:{PASSWORD} -X DELETE {INDEX_URL}
+----
+
+=== Step 2: Run the Heartbeat setup command
+
+["source","sh"]
+----
+./heartbeat setup -e 
+----
+
+This will perform the necessary startup tasks and ensure that your indices will have the appropriate
+mapping going forward.
+
+=== Step 3: Run Heartbeat again
+
+Now when you run heartbeat, your data will be indexed with the appropriate mappings. When
+the {uptime-app} attempts to fetch the data, it should be able to render without issues.
--- a/docs/en/observability/troubleshoot-uptime.asciidoc
+++ b/docs/en/observability/troubleshoot-uptime.asciidoc
@@ -18,7 +18,7 @@ delete all the {heartbeat} indicies that match the pattern the {uptime-app} will
 
 
 
-=== Step 2: Run the Heartbeat setup command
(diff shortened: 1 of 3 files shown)


docs/en/observability/troubleshoot-uptime.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+state, the UI cannot function.
+
+To fix this problem, you typically need to remove your Heartbeat indices, and create
+new ones, with the appropriate mappings installed. To achieve this, follow the steps below.
+
+=== Step 1: Delete your Heartbeat indices
+
+["source","sh"]
+----
+curl -u {USERNAME}:{PASSWORD} -X DELETE {INDEX_URL}
```
justinkambic, line 17: I was unsure what was appropriate to use in this case; my thinking was by including these missing values the user would be less likely to simply copy/paste and delete something before they were ready.

docs/en/observability/troubleshoot-uptime.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+----
+
+=== Step 2: Run the Heartbeat setup command
+
+["source","sh"]
+----
+./heartbeat setup -e
```
justinkambic, line 24: Also unsure if it's appropriate to supply this command here, or if we should wholesale link to the page of the Beats book that deals with this command, to avoid this text becoming outdated.

docs/en/observability/troubleshoot-uptime.asciidoc
andrewvc: I would avoid cURL here and make the advice more generic. People can do this in a number of ways, including the kibana management UI. I don't think we need a code example at all. I'd maybe link to the kibana mgmt docs and the ES index delete api docs.

docs/en/observability/troubleshoot-uptime.asciidoc
```diff
@@ -0,0 +1,33 @@
+[[troubleshoot-uptime]]
+= Uptime troubleshooting
+
+== Mapping issues
+
+It is possible for Heartbeat to create an index without the appropriate mappings.
```
andrewvc, line 6: Could we phrase this differently? This makes it seem like a heartbeat bug. Maybe:

"There are situations in which Heartbeat data can be indexed without the correct mappings applied. These situations cannot occur with the elastic-agent configured via fleet, only with standalone heartbeat. This can occur when the underlying `heartbeat-VERSION` ILM alias is deleted manually or when heartbeat writes data through an intermediary such as Logstash without the `setup` command being run."

docs/en/observability/troubleshoot-uptime.asciidoc
andrewvc: Actually, it may make sense to link to https://www.elastic.co/guide/en/beats/heartbeat/current/heartbeat-installation-configuration.html somewhere

docs/en/observability/troubleshoot-uptime.asciidoc
andrewvc: We may also want to recommend agent / fleet as an alternative here

andrewvc: Though they still need to delete the bad indices in that case

justinkambic: > Could we phrase this differently? This makes it seem like a heartbeat bug.

Added in https://github.com/elastic/observability-docs/pull/1046/commits/5e439ac02744484eff67ea9612e817cb09c02f71.

Need to research where in Fleet docs it makes the most sense for us to link.

docs/en/observability/troubleshoot-uptime.asciidoc
justinkambic: Link added in https://github.com/elastic/observability-docs/pull/1046/commits/5e439ac02744484eff67ea9612e817cb09c02f71.

docs/en/observability/troubleshoot-uptime.asciidoc
justinkambic: Removed the comment, linked to Index Management and Delete Index API docs https://github.com/elastic/observability-docs/pull/1046/commits/5e439ac02744484eff67ea9612e817cb09c02f71.

docs/en/observability/troubleshoot-uptime.asciidoc
```diff
@@ -0,0 +1,44 @@
+[[troubleshoot-uptime]]
+= Uptime troubleshooting
+
+== Mapping issues
+
+There are situations in which {heartbeat} data can be indexed without the correct mappings applied.
+These situations cannot occur with the {elastic-agent} configured via fleet, only with standalone {heartbeat}.
```
hendry-lim, line 7:
```suggestion
These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat}.
```

EamonnTP (commented):
Thanks @justinkambic for adding this to the docs. It looks great! Just a few minor suggestions from me.

docs/en/observability/troubleshoot-uptime.asciidoc
```diff
@@ -0,0 +1,44 @@
...
+The below command will cause {heartbeat} to perform its setup processes.
+
+NOTE: For more information on how to use this command, you can reference the
+{heartbeat-ref}/heartbeat-installation-configuration.html[Heartbeat quickstart guide].
+
+["source","sh"]
+----
+./heartbeat setup -e 
+----
+
```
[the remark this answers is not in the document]
justinkambic: This seems to render text like `{beatname_lc} setup -e` - is this desired? Sorry, I'm a somewhat ignorant of asciidoc conventions. Is there some argument I need to supply to make it show `heartbeat`?

docs/en/observability/troubleshoot-uptime.asciidoc
```diff
@@ -0,0 +1,44 @@
+[[troubleshoot-uptime]]
+= Uptime troubleshooting
```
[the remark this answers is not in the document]
justinkambic: That sounds fine to me - the original title was just intended as a draft anyway. Modified in https://github.com/elastic/observability-docs/pull/1046/commits/ca1338aaa195229a8d040dff941585c120096e08.

docs/en/observability/troubleshoot-uptime.asciidoc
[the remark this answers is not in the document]
EamonnTP: Apologies! Yes, you can the `:beatname_lc: heartbeat` under the title name and `:!beatname_lc:` at the bottom on the asciidoc file. For an example, you can see it in this [file](https://github.com/elastic/observability-docs/edit/7.14/docs/en/observability/ingest-uptime.asciidoc).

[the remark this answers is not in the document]
justinkambic: Yep, I actually just figured this out by looking at the file. Sorry, should've investigated first.

I've just pushed https://github.com/elastic/observability-docs/pull/1046/commits/13caad02cf4e73b0a27018567f5b248d9e0c7300.

justinkambic:
I think all feedback has been addressed, please LMK if any more revisions are required. Thanks for the speedy reviews!

EamonnTP (approved):

andrewvc (changes requested):
There's one critical step missing here, you must stop all heartbeat instances before deleting the indices!

justinkambic:
> There's one critical step missing here, you must stop all heartbeat instances before deleting the indices!

Somehow when transferring my notes to here I missed adding this! Well-spotted.

andrewvc (approved):
LGTM

PR #1058, opened 2021-09-13
Add Endpoint security and Osquery to beats/agent comparison

mostlyjason:
I noticed that we are missing Endpoint security and Osquery in the beats/agent comparison docs https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html.  These are key advantages and reasons to adopt Elastic Agent so we should include them in the comparison.

CC @melissaburpo @caitlinbetz

### DIFF
--- a/docs/en/ingest-management/beats-agent-comparison.asciidoc
+++ b/docs/en/ingest-management/beats-agent-comparison.asciidoc
@@ -73,6 +73,11 @@ The following table shows the inputs supported by the {agent} in {version}:
 |{n}
 |{y}
 
+|Endpoint security
+|{n}
+|{n}
+|{y}
+
 |GCP Pub/Sub
 |{y}
 |{y}
@@ -110,6 +115,11 @@ The following table shows the inputs supported by the {agent} in {version}:
 
 |O365 Mgmt Act API
 |{y} (beta)
+|{y} (beta)
+|{y} (beta)
+
+|Osquery 
+|{n}
 |{y} (beta)
 |{y} (beta)
 
--- a/docs/en/ingest-management/beats-agent-comparison.asciidoc
+++ b/docs/en/ingest-management/beats-agent-comparison.asciidoc
@@ -75,8 +75,8 @@ The following table shows the inputs supported by the {agent} in {version}:
 
 |Endpoint security
 |{n}
-|{n}
 |{y}
+|{n}
 
 |GCP Pub/Sub
 |{y}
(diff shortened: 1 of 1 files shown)


nimarezainia (approved):

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -113,6 +118,11 @@ The following table shows the inputs supported by the {agent} in {version}:
 |{y} (beta)
 |{y} (beta)
 
+|Osquery 
+|{n}
+|{y} (beta)
+|{y} (beta)
```
melissaburpo, line 124: I checked with the team, and for now, I think we should only list Osquery Manager support for `Fleet-managed Elastic Agent`. While technically aspects can work for standalone agents (e.g. you can manually schedule queries), currently they would not be able to do live queries. Similar to other items on this page, instead of listing it as `{n}`, we could note that it is `Under consideration`. What do you think, @mostlyjason?

melissaburpo: After thinking about it more, I suspect we should just list it as not supported for standalone mode, similar to Endpoint Security.

For standalone mode, people could use the Osquery Log Collection integration. But the newer Osquery Manager integration is made specifically for centrally managing osquery deployments in Fleet.

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -73,6 +73,11 @@ The following table shows the inputs supported by the {agent} in {version}:
 |{n}
 |{y}
 
+|Endpoint security
```
dedemorton, line 76: Wait this says that Endpoint is not supported on Fleet-managed Elastic Agent?

image

dedemorton (commented):
Can you look at the preview and confirm that everything renders as you expect?

https://observability-docs_1058.docs-preview.app.elstc.co/guide/en/fleet/master/beats-agent-comparison.html

docs/en/ingest-management/beats-agent-comparison.asciidoc
mostlyjason: Ok thanks I updated it so that it is only shown as support for Fleet-managed agents

docs/en/ingest-management/beats-agent-comparison.asciidoc
mostlyjason: Thanks I got the order of the columns wrong. I updated it to only supported for Fleet-managed agents.

melissaburpo (approved):

dedemorton (approved):
LGTM

Issue #960, opened 2021-08-04
RAC documentation


paulb-elastic:
## Description

The RAC implementation going in with 7.15, will need to be documented.

(placeholder issue https://github.com/elastic/kibana/issues/107614 raised, although it's just a stub at the moment)

It may be best for the docs team to liaise with the stakeholders after FF, to determine what has made it into 7.15, that needs documenting.

## Collaboration

* [ ] ~~The docs team will lead producing the content~~
* [ ] ~~The product team will provide the initial content and the docs team will edit / review~~
* [X] The docs team will define with product team the structure and location, and the product team will provide the initial content
* [ ] ~~Other (please describe)~~

## Suggested Target Release

7.15

## Stakeholders

@jasonrhodes / @weltenwort / @Kerry350 / @mgiota

PR #1062, opened 2021-09-15
Add alerts and cases content

EamonnTP:
### Summary

This PR adds content for the Alerts page and the Cases page within the Observability app.

### Related issue

Closes https://github.com/elastic/observability-docs/issues/960


mgiota:
@EamonnTP Thanks a lot for adding content for Alerts and Cases pages!

LGTM! I leave a few comments below:

`Alerts`
I liked how you structured the content under subsections. It gives a clear view to the user on how he can manage the alerts.

> The Alerts page lists all the alerts you have created using the Logs, Metrics, Uptime, or APM apps.

We don’t actually create alerts through Logs, Metrics etc but `rules` that eventually can trigger alerts if the specified conditions are met. Could we possibly rephrase this sentence? 

> Alerting can also be centrally managed from the Kibana Management UI that provides a set of built-in rule types and connectors for you to use. Click Manage Rules

I would say Alerting and Rule Management are not exactly the same. At least code wise we differentiated these two concepts. Shall we better say `Rules` can be centrally managed from the Kibana Management UI instead? We could possibly add to this that alerts created from Kibana Management UI rules will also show up in the observability alerts table.

PR #1072, opened 2021-09-15
Update air-gapped.asciidoc

akshay-saraswat:
* Changed Important Admon to Warning Admon mentioning that this is a beta feature

### DIFF
--- a/docs/en/integrations/air-gapped.asciidoc
+++ b/docs/en/integrations/air-gapped.asciidoc
@@ -26,7 +26,7 @@ For more information, see the {fleet-guide}/fleet-overview.html#package-registry
 [[air-gapped-diy-epr]]
 == Host your own Elastic Package Registry
 
-IMPORTANT: This procedure is not officially supported.
+beta::[]
 
 If routing traffic through a proxy server is not an option, you can host your own Elastic Package Registry.
 The Package Storage instance must be deployed and hosted on-site as a Docker container.


docs/en/integrations/air-gapped.asciidoc
```diff
@@ -26,7 +26,7 @@ For more information, see the {fleet-guide}/fleet-overview.html#package-registry
 [[air-gapped-diy-epr]]
 == Host your own Elastic Package Registry
 
-IMPORTANT: This procedure is not officially supported.
+WARNING: This functionality is in beta and is subject to change. The design and code is less mature than official GA features and is being provided as-is with no warranties. Beta features are not subject to the support SLA of official GA features.
```
bmorelli25, line 29: Thanks! We actually have beta admonitions built into our doc system. You can use them like this:
```suggestion
beta::[]
```

mtojek: Not sure if in this case it's important to put this statement.

akshay-saraswat: @mtojek We discussed this during one of our team meetings.  "IMPORTANT: This procedure is not officially supported."  looked weird, so, we decided to change the wording to the beta statement. Please let me know if you still don't like this. I am happy to trash this PR.

mtojek: Ok, thank you for the explanation, Akshay (although I still consider the statement more valuable than just "beta" label :) ).

mtojek (approved):

bmorelli25 (approved):

PR #1085, opened 2021-09-21
[Uptime] Clarify docs around mapping issues / standalone agent

andrewvc:
Prior to this change the docs didn't factor in standalone agent. They now do, and correctly describe what to do in this case. In this
situation they must be vague because the standalone agent use case gives users a lot of power and discretion over how they set things up. We can only give a rough guide.

### DIFF
--- a/docs/en/observability/troubleshoot-uptime-mapping-issues.asciidoc
+++ b/docs/en/observability/troubleshoot-uptime-mapping-issues.asciidoc
@@ -7,7 +7,7 @@
 == Mapping issues
 
 There are situations in which {heartbeat} data can be indexed without the correct mappings applied.
-These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat} and {elastic-agent} running in standalone mode.
+These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat} or {elastic-agent} running in standalone mode.
 This can occur when the underlying `heartbeat-VERSION` ILM alias is deleted manually or when {heartbeat} writes data
 through an intermediary such as {ls} without the `setup` command being run. 
 When running {elastic-agent} in standalone mode this can happen if manually setup datastreams have incorrect mappings.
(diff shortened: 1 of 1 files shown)


docs/en/observability/troubleshoot-uptime-mapping-issues.asciidoc
```diff
@@ -7,30 +7,31 @@
 == Mapping issues
 
 There are situations in which {heartbeat} data can be indexed without the correct mappings applied.
-These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat}.
+These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat} and {elastic-agent} running in standalone mode.
```
justinkambic, line 10:
```suggestion
These situations cannot occur with the {elastic-agent} configured via {fleet}, only with standalone {heartbeat} or {elastic-agent} running in standalone mode.
```

docs/en/observability/troubleshoot-uptime-mapping-issues.asciidoc
```diff
@@ -7,30 +7,31 @@
...
 This can occur when the underlying `heartbeat-VERSION` ILM alias is deleted manually or when {heartbeat} writes data
-through an intermediary such as {ls} without the `setup` command being run.
+through an intermediary such as {ls} without the `setup` command being run. 
+When running {elastic-agent} in standalone mode this can happen if manually setup datastreams have incorrect mappings.
```
justinkambic, line 13:
```suggestion
When running {elastic-agent} in standalone mode this can happen if manually-setup datastreams have incorrect mappings.
```

justinkambic (approved):
I am glad we are removing the numbered steps, not needed and too error-prone!

Had a few minor nits, sounds good to me.

docs/en/observability/troubleshoot-uptime-mapping-issues.asciidoc
andrewvc: I don't think this is grammatically more correct, do you have a strong feeling here? Maybe @bmorelli25 or @dedemorton have a thought

PR #1091, opened 2021-09-23
 Update Fleet Server docs to include production mode option

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/1008 

Preview link: <to be added>

Updates the steps about adding a Fleet Server to include deployment mode options. Also clarifies questions around deploying a self-managed Fleet Server when using Elastic Cloud

@mostlyjason @EricDavisX I need some help here because it's still not clear to me how cloud users deal with certs when they want to run self-managed Fleet Servers. Can you clarify?  I have not tested all these steps because the setup with certs is very time consuming.

Eventually we might want to rethink the structure of the docs for cloud vs self-managed because some of this content is kind of meaty to include in our tabbed widgets. This is not a problem I can solve easily today, though, so let's get the details right and we can improve the format over time.


EricDavisX:
Hi - I want to confirm the minimal text changes here so far are ok by me, but indeed they don't really close out the ticket.  We are treading on a topic that isn't 100% agreed as to if we can or do support it (using Self-Managed Fleet Servers against a ESS Elastic cloud-based stack.  I'd like @ruflin and @mostlyjason to chime in.

It may be something we solve not just in the Docs, too, maybe the UI option referenced needs a spec enhancement.

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -21,15 +21,20 @@ APM & {fleet} node. This node is required to use {fleet-server}.
...
 // end::ess[]
 
 // tag::self-managed[]
 
-To deploy {fleet-server} on a self-managed cluster, install an {agent} and
-enroll it in the default {fleet-server} policy. The policy is created
-automatically when you run {fleet} for the first time.
+//QUESTION: Is it still recommended that users configure the Fleet Settings up
+//front, or does the Fleet Server host setting in the in-product steps make
+//that unnecessary?
```
mostlyjason, line 34: When the user adds a new fleet server they need to add the host to the list of hosts in fleet settings

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -21,15 +21,20 @@ APM & {fleet} node. This node is required to use {fleet-server}.
...
 
-To add {fleet-server}:
+To deploy a self-managed {fleet-server}, install an {agent} and enroll it in a
+{fleet-server} policy.
```
mostlyjason, line 37: enroll it in an agent policy containing the {fleet-server} integration

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -52,20 +57,65 @@ For more information about these settings, see <<fleet-settings>>.
 [role="screenshot"]
 image::images/add-fleet-server.png[In-product instructions for adding a {fleet-server}]
 
+//QUESTION: Why does the Fleet Server install command include the --url flag
+//when Production mode is selected)? It's not there in Quick Start mode.
+//I thought --url was only required when you are enrolling subsequent agents in
+// Fleet Server.
```
mostlyjason, line 63: I'm not sure the url is needed here. Can you confirm @nimarezainia @blakerouse?

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -52,20 +57,65 @@ For more information about these settings, see <<fleet-settings>>.
...
+
+//QUESTION: How do cloud users handle the problem with certs if they select
+//Production mode?
```
mostlyjason, line 66: This is a good question. I think Elastic Cloud uses public CAs so the user does not need to supply certs when enrolling agents. I think the same would be true if the self-managed fleet server is using either public certs, like we offer on Elastic cloud, or private certs if the host has the CA installed at the OS level. I'm less sure if we can handle the case where they are using a cloud-based fleet server and a self-managed fleet server with a private CA. Should we say this use case is not supported at the moment @nimarezainia @blakerouse?

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -52,20 +57,65 @@ For more information about these settings, see <<fleet-settings>>.
...
+// Fleet Server.
+
+//QUESTION: How do cloud users handle the problem with certs if they select
+//Production mode? 
+
 *Notes:*
 
 * Make sure you download an x64 architecture installation package.
+
+* For the agent policy, choose a {fleet-server} policy. If one doesn't exist,
```
mostlyjason, line 72: choose one with a {fleet-server} integration

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -52,20 +57,65 @@ For more information about these settings, see <<fleet-settings>>.
...
+create it.
++
+.What is a {fleet-server} policy?
+****
+It is an agent policy that configures {agent} to run as a {fleet-server}. In
```
mostlyjason, line 77: It is an agent policy that contains a {fleet-server} integration

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -52,20 +57,65 @@ For more information about these settings, see <<fleet-settings>>.
...
+----
++
+The following command installs a {fleet-server} and uses certificates you
+provide. Make sure you replace the values in angle brackets.
 +
 [source,yaml]
 ----
-sudo ./elastic-agent install -f --fleet-server-es=http://localhost:9200 \
---fleet-server-service-token=AAEAAWVsYXN0aWMvZmxlZXQtc2VydmVyL3Rva2VuLTE2MTkxMzg3MzIzMTg6dzEta0JDTmZUZGlDTjlwRmNVTjNVQQ
+sudo ./elastic-agent install --url=https://192.0.2.1:8220 \
```
mostlyjason, line 110: double check on the url

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
dedemorton: oh good point! I was trying to be descriptive, but it's better to be precise here.

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
dedemorton: @mostlyjason Ah so it looks like Pius opened an issue in the beats repo saying that --url is required when certs are specified: https://github.com/elastic/beats/issues/27245

dedemorton: ...but I do remember testing those commands fairly rigorously, so I'm pretty sure the --url setting was not always required when passing certs. Would be good to get clarification from the dev team about this.

docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
dedemorton: @mostlyjason  This is potentially confusing to users on multiple levels. Since users need to select either Quick Start or Production, Cloud users will see the command that shows all the certs flags, even if they aren't required to pass them. 

It feels like these options shouldn't be available when running Kibana on cloud (unless I'm missing something here).

mostlyjason: That's a good point. I'd like to confirm with engineering if the above is correct. If so, we should probably remove those options on cloud.

dedemorton:
This PR will address https://github.com/elastic/beats/issues/27245 in versions 7.15 and later, but I will need to create another PR to backport the change to other versions.

docs/en/ingest-management/commands.asciidoc
```diff
@@ -87,13 +90,18 @@ elastic-agent enroll --fleet-server-es <string>
...
                      [global-flags]
 ----
 <1> If no `fleet-server-cert*` flags are specified, {agent} auto-generates a
 self-signed certificate with the hostname of the machine. Remote {agent}s
 enrolling into a {fleet-server} with self-signed certificates must specify
-the `insecure` flag.
+the `--insecure` flag.
+<2> Required when enrolling in a {fleet-server} with custom certificates. The
+URL must match the DNS name used to generate the certificate specified by
+`--fleet-server-cert`.
```
blakerouse, line 102: This is a nice add!

blakerouse (approved):
Looks good.

PR #1093, opened 2021-09-23
Add osquery link

EamonnTP:
### Summary

This PR adds a link to the osquery topic in the Kibana docs.

### Docs preview

https://observability-docs_1093.docs-preview.app.elstc.co/guide/en/observability/master/host-metrics.html


ollyhowell (approved):
Looks good to me.
We should resolve the osquery capitalization and be consistent across the docs - their own docs do use "osquery" but makes sense to keep this page internally consistent and we can bring it up with the group and do a wholesale 'find and replace' perhaps.

Had a quick read of whole page and there is a minor "mintues" typo in the screenshot: 

image

PR #1110, opened 2021-09-30
Add 'Choosing between Agent and Beats' section

ollyhowell:
Potential update to https://www.elastic.co/guide/en/fleet/7.15/beats-agent-comparison.html based on current discussions.
Open to feedback - depending on outstanding threads this could change quite a bit.

### DIFF
--- a/docs/en/ingest-management/beats-agent-comparison.asciidoc
+++ b/docs/en/ingest-management/beats-agent-comparison.asciidoc
@@ -26,6 +26,35 @@ This article summarizes the features and functionality you need to be aware of
 before adding new {agent}s or replacing your current {beats} with {agent}s.
 Starting in version 7.14.0, {agent} is generally available (GA).
 
+[[choosing-between-agent-and-beats]]
+== Choosing between Elastic Agent and Beats
+
+Elastic Agent is a single binary designed to provide the same functionality that the various Beats provide today. However, there are some current functionality gaps that are continuously being addressed as we strive to reach feature parity.
+
+The following steps will help you determine if Elastic Agent can support your use case:
+
+**Step 1:** Determine if the Integrations you need are supported and Generally Available (GA) on Elastic Agent. Integrations that are currently GA are:
+
+* `Auditd`
+* `Elastic Endpoint Security`
+* `Fortinet`
+* `Netflow`
+* `Office365`
+* `Okta`
+* `Palo Alto Networks`
+* `Suricata`
+* `System`
+* `Windows`
+* `Zeek`
+
+NOTE: See [Integrations documentation](https://docs.elastic.co/integrations) for further information and to view the status of all Integrations.
+
+**Step 2:** If the Integration is available, check to see whether the required output is also supported. The supported outputs are outlined in this document ([Supported Outputs](https://docs.google.com/document/d/16qAC3iuNudiiMS0vL5PNHGvD6NQOZ0FVhaUZsKSHavw/edit#heading=h.ntxz032s5u9r))
+
+**Step 3:** If Elastic Agent supports the integration and output you require, you should then review [Supported Capabilities](https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#additional-capabilties-beats-and-agent) to determine if any specific features required by your deployment are supported. Elastic Agent should support the majority of features available on Beats and will be updated for each release.
+
+If you are satisfied with all three steps, then Elastic Agent is suitable for deployment. If any steps fail your assessment, you should continue to use Beats and review future updates or contact us in the [discuss forum](https://discuss.elastic.co/).
+
 [discrete]
 [[supported-inputs-beats-and-agent]]
 == Supported inputs
@@ -117,7 +146,7 @@ The following table shows the inputs supported by the {agent} in {version}:
 |{y} (beta)
 |{y} (beta)
 
-|Osquery 
+|Osquery
 |{n}
 |{y} (beta)
 |{n}
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -7,7 +7,7 @@ want to monitor. These steps assume that you're running a fresh installation. If
 {agent} is already running on your system and you want to upgrade to a new
 version, see <<upgrade-elastic-agent>>.
 
-You have a couple options for installing and managing an {agent}:
+You have a couple of options for installing and managing an {agent}:
 
 * **Install a {fleet}-managed {agent} (recommended)**
 +


EamonnTP (commented):
Looks great! Just some minor suggestions. I would always check a local build of the updates to ensure that links are displayed correctly. We also use [shared attributes](https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc) for doc links and product names.

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -26,6 +26,35 @@ This article summarizes the features and functionality you need to be aware of
 before adding new {agent}s or replacing your current {beats} with {agent}s.
 Starting in version 7.14.0, {agent} is generally available (GA).
 
+[[choosing-between-agent-and-beats]]
+== Choosing between Elastic Agent and Beats
```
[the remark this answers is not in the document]
ollyhowell: Thanks Eamonn! I should have checked the shared attributes

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -26,6 +26,35 @@ This article summarizes the features and functionality you need to be aware of
...
+
+Elastic Agent is a single binary designed to provide the same functionality that the various Beats provide today. However, there are some current functionality gaps that are continuously being addressed as we strive to reach feature parity.
+
+The following steps will help you determine if Elastic Agent can support your use case:
+
+**Step 1:** Determine if the Integrations you need are supported and Generally Available (GA) on Elastic Agent. Integrations that are currently GA are:
```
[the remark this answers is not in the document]
ollyhowell: I did think about that but never switched them...yeah, you're right - will make the change.

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -26,6 +26,35 @@ This article summarizes the features and functionality you need to be aware of
...
+* `Okta`
+* `Palo Alto Networks`
+* `Suricata`
+* `System`
+* `Windows`
+* `Zeek`
+
+NOTE: See [Integrations documentation](https://docs.elastic.co/integrations) for further information and to view the status of all Integrations.
+
+**Step 2:** If the Integration is available, check to see whether the required output is also supported. The supported outputs are outlined in this document ([Supported Outputs](https://docs.google.com/document/d/16qAC3iuNudiiMS0vL5PNHGvD6NQOZ0FVhaUZsKSHavw/edit#heading=h.ntxz032s5u9r))
```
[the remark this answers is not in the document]
ollyhowell: No, this should be the anchor link in page - good spot.

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -26,6 +26,38 @@ This article summarizes the features and functionality you need to be aware of
...
+* `Windows`
+* `Zeek`
++
+NOTE: For more details, refer to {integrations-docs}[{integrations}].
++
+. If the integration is available, check <<supported-outputs-beats-and-agent>> to see whether the required output is also supported.
+
+. Review <<additional-capabilties-beats-and-agent>> to determine if any features required by your deployment are supported. {agent} should support most of the features available on {beats} and is updated for each release.
+
+// QUESTION: would we ever correct a typo in anchor? i.e. https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#additional-capabilties-beats-and-agent
```
dedemorton, line 57: Yes, especially for anchors that aren't used to derive the html filename. Links to the topic will still work; they just won't send users to the anchor which IMO is OK.

I also think we should fix typos in html filenames, but we'll also need to set up redirects if we do.

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -26,15 +26,16 @@ This article summarizes the features and functionality you need to be aware of
...
 [[choosing-between-agent-and-beats]]
 == Choosing between {agent} and {beats}
 
 {agent} is a single binary designed to provide the same functionality that the various {beats} provide today. However, some functionality gaps are being addressed as we strive to reach feature parity.
 
 The following steps will help you determine if {agent} can support your use case:
 
-**Step 1:** Determine if the Integrations you need are supported and Generally Available (GA) on Elastic Agent. Integrations that are currently GA are:
-
+. Determine if the integrations you need are supported and Generally Available (GA) on Elastic Agent. Integrations that are currently GA are:
```
dedemorton, line 37:
```suggestion
. Determine if the integrations you need are supported and Generally Available (GA) on {agent}. Integrations that are currently GA are:
```

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -46,14 +47,16 @@ The following steps will help you determine if {agent} can support your use case
...
 
-**Step 3:** Review https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#additional-capabilties-beats-and-agent[Supported Capabilities] to determine if any features required by your deployment are supported. {agent} should support most of the features available on {beats} and is updated for each release.
+// QUESTION: would we ever correct a typo in anchor? i.e. https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#additional-capabilties-beats-and-agent
 
-If you are satisfied with all three steps, then Elastic Agent is suitable for deployment. If any steps fail your assessment, you should continue to use Beats and review future updates or contact us in the [discuss forum](https://discuss.elastic.co/).
+If you are satisfied with all three steps, then {agent} is suitable for your deployment. However, if any steps fail your assessment, you should continue using {beats}, and review future updates or contact us in the https://discuss.elastic.co/[discuss forum].
```
dedemorton, line 59:
```suggestion
If you are satisfied with all three steps, then {agent} is suitable for your deployment. However, if any steps fail your assessment, you should continue using {beats}, and review future updates or contact us in the {forum}[discuss forum].
```

dedemorton (commented):
Just a few minor places where we can use attributes instead of hard coding values.

Thanks for doing this!

We should have a general discussion about line wraps. In the past, GitHub didn't handle long (unwrapped lines) very well, but it does a better job now.

Historically I've wrapped at 80 chars because Atom has a command for that; other teams like APM wrap at logical phrases, but I'm not a fan. Let's leave your long lines as-is and decide how to proceed as a team. I'm not sure how long lines affect editing in the GitHub UI, but we'll need to look at that for external contributors.

docs/en/ingest-management/beats-agent-comparison.asciidoc
dedemorton: The `forum` attribute is defined in https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc#L109.

docs/en/ingest-management/beats-agent-comparison.asciidoc
ollyhowell: Good point re: anchors vs. page links.
I'll fix this one then unless any objections.

docs/en/ingest-management/beats-agent-comparison.asciidoc
ollyhowell: Thanks for both (think I forgot to fold in one of Eamonn's {agent} corrections in too).

I've been using Atom with Default: 80 too but might need a steer to what to fix or we can discuss/document as a group later.

nimarezainia (approved):

nimarezainia:
thanks @ollyhowell

dedemorton:
> Not critical but some of the table widths are a little sub-optimal - will make a note to have a look into options here (e.g. one table we could make the headings a little more concise and the spacing would improve).

Feel free to have a go if you can come up with better col headings. It's super frustrating that the doc build doesn't respect col attributes like width because we should be able to control at least relative col sizes. I've given up on trying to make our tables look good. :-P

PR #1114, opened 2021-10-04
Fix Fleet setting key for registry proxy URL

jen-huang:
While working on https://github.com/elastic/kibana/pull/113858, I was looking for if we have any documentation around licensing for custom registry URL that needs to be corrected. I did not find any, but came across this which needs to be updated to reflect the `ingestManager` -> `fleet` setting renaming that happened a long time ago.

Could whoever reviews this help me with merging and backporting this to all versions of the docs that contain this? (Looks like >= 7.13)

### DIFF
--- a/docs/en/ingest-management/overview.asciidoc
+++ b/docs/en/ingest-management/overview.asciidoc
@@ -56,7 +56,7 @@ To do so, add the following setting to your `kibana.yml` file:
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


dedemorton (approved):
LGTM.

TBH, I think this level of detail doesn't really belong in an overview, but let's leave it there for now until we have it covered elsewhere. 

I can get this backported. Typically we only backport to the latest in-service branches, but I'll try to get this to go back as far as I can.

PR #1139, opened 2021-10-07
Show username setting in Fleet quick start prereqs

### LINKED ISSUE #665: Show username setting in Fleet quick start example
In the [quick start](https://www.elastic.co/guide/en/fleet/current/fleet-quick-start.html) prereqs section, under self-managed, we show a security config example to help users get started quickly.

At some point, we removed the username/password settings (probably because we don't want users hard coding their password in config files). But this makes the example less useful. At the very least, we need to show the username setting:

```
elasticsearch.username: "kibana_system"
```

Then maybe add a note that points to the [security docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.13/security-minimal-setup.html#add-built-in-users) for info about storing the ES password in the keystore.

dedemorton:
Fixes #665


EamonnTP (approved):
LGTM!

PR #1146, opened 2021-10-07
Update air-gapped.asciidoc

mostlyjason:
Changing to fleet since ingestManager is the old name.

CC @akshay-saraswat


dedemorton (approved):
LGTM

PR #1172, opened 2021-10-15
[monitor-gcp] add Google Dataflow template

andremm:
As requested by @ravikesarwani and @melvynator, this PR includes the Google Dataflow template to the Monitor GCP tutorial. The instructions are based on this [blog post](https://www.elastic.co/blog/ingest-data-directly-from-google-pub-sub-into-elastic-using-google-dataflow).


ravikesarwani:
I feel that mixing both the Beats/Agent (we will have to add Elastic Agent here soon) and Agentless methods into a single tutorial is making the tutorial very long and maybe confusing for some users.

I would prefer that we break the native integration into a separate page linked from the main one.
I think this is a scalable pattern that we probably can follow in Azure and later in AWS as well.

Before “What you’ll learn” we can add something like “Skip installing Agent and use Google Dataflow templates to ship logs directly to Elastic Stack, with just a few clicks in the Google Cloud Console.” and provide a link for user to jump to that page.

I know some content will have to be repeated like “creating Pub/Sub topic and subscription, before you begin etc.” but I think it’s worth it compared to the potential user confusion mixing the 2 cases in one tutorial.

Thoughts?

andremm:
> I would prefer that we break the native integration into a separate page linked from the main one.

I agree. I will make the changes and ping you when they are ready for review.

melvynator:
So the goal of the tutorial section is to provide tutorials to specific challenges that the customer may face, like monitor aws, gcp, aws, etc... 

We can split this page in two but we need to think, where, in the doc, should this new page about Google Dataflow resides. It should probably reside in a different section of the doc than the tutorial one as it's not really "a challenge" on its own.

melvynator:
Ok so if I understand it correctly the tutorial section will look like this:

Tutorial:
 - Monitor AWS
 - Monitor GCP
 - Monitor Azure
 - Monitor Kubernetes
 - Monitor a java application
 - New Page about Datablow
 
Can we call this new page "Serverless Monitoring"? Or shall we call it something less generic like Monitoring GCP using Dataflow?

ravikesarwani:
I was thinking more like this for the time being =>

Tutorial:

- Monitor AWS
- Monitor GCP
      - GCP Dataflow Templates
- Monitor Azure
- Monitor Kubernetes
- Monitor a java application

melvynator:
Make sense. Thanks a lot for your input.

ravikesarwani:
Hi @andremm I reviewed the new page and look really great. Thanks for all the hard work on putting this together.

Few minor comments I had, more for discussion purposes and hear your thoughts to see if it makes sense:
- I was wondering if it makes more sense to make Step 2 (Install the GCP Integration in Kibana) as Step 1.
Here we can also ask the users to collect the Cloud ID and API key to be used in the tutorial later.
This will ensure that all GCP console steps can be executed in one go and we have less transitions.
- Around the API key: I was wondering if we need to provide a little bit more details OR link to some help tutorial for users to set the privileges correctly. At a minimum point to something like https://www.elastic.co/guide/en/kibana/current/api-keys.html. Not sure if we can copy the pattern from some other tutorial.

Overall this is really great and easy to follow.

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+And you can use {kibana-ref}/api-keys.html#create-api-key[{kib}] to create a
+Base64-encoded API key to authenticate on your deployment.
+
+[IMPORTANT]
+====
+You can optionally restrict the privileges of your API Key, otherwise they'll
+be a point in time snapshot of permissions of the authenticated user.
+For this tutorial the data is written to the `logs-gcp.audit-default` data streams.
+====
+
```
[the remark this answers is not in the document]
dedemorton: Here I'm suggesting some minor edits plus a formatting change that puts a blue box around your description so that all the related (but somewhat ancillary) info is grouped together. I didn't bother copying over the image, but you might need to crop it further to get it to work in the blue box. It'll look something like this:

image

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+[discrete]
+==== Step 2: Create a Pub/Sub topic and subscription
+
+Before configuring the Dataflow template, you will have to create a Pub/Sub
```
dedemorton, line 68:
```suggestion
Before configuring the Dataflow template, create a Pub/Sub
```

docs/en/observability/gcp-topic.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+Set *sink name* as `monitor-gcp-audit-sink`. Select the *Cloud Pub/Sub topic* as the
+*sink service* and *Create new Cloud Pub/Sub topic* named `monitor-gcp-audit`:
++
+image::monitor-gcp-create-pubsub-topic.png[Create Pub/Sub topic]
++
+Finally, under *Choose logs to include in sink*, add
+`logName:"cloudaudit.googleapis.com"` (it includes all audit logs).
+Click *create sink*.  It will look something like the following:
++
+image::monitor-gcp-create-sink.png[Create logs routing sink]
```
dedemorton, line 19: I wonder if this document needs two screen captures that are fairly similar? Screen captures of third-party tools can be hard to keep up-to-date over time, so it's best to limit them, if possible.

I'd show the finished screen rather than in-progress.

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
++
+Set `monitor-gcp-audit-sub` as the *Subscription ID* and leave the
+*Delivery type* as pull:
++
+image::monitor-gcp-pub-sub-subscription-id.png[GCP Pub/Sub Subscription ID]
++
+Finally, scroll down and click *Create*.
+
+[discrete]
+==== Step 3: Configure the Google Dataflow Template
```
dedemorton, line 91:
```suggestion
==== Step 3: Configure the Google Dataflow template
```

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+
+image::monitor-gcp-dataflow-jobs.png[GCP Dataflow Jobs]
+
+To create a job, click *Create Job From Template*.
+Set *Job name* as `auditlogs-stream` and select `Pub/Sub to Elasticsearch` from
+the *Dataflow template* dropdown menu:
+
+image::monitor-gcp-dataflow-pub-sub-elasticsearch.png[GCP Dataflow Pub/Sub to Elasticsearch]
+
+Before running the job, you will need to fill in required parameters:
```
dedemorton, line 104:
```suggestion
Before running the job, fill in required parameters:
```

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+
+image::monitor-gcp-dataflow-required-parameters.png[GCP Dataflow Required Parameters]
+
+[NOTE]
+====
+The *Cloud Pub/Sub subscription* is the same you created in the previous step.
```
dedemorton, line 110:
```suggestion
For *Cloud Pub/Sub subscription*, use the subscription you created in the previous step.
```

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+The *Cloud ID* and *Base64-encoded API Key* are the same from first step.
```
dedemorton, line 111:
```suggestion
For *Cloud ID* and *Base64-encoded API Key*, use the values you got earlier.
```

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+If you don't have an *Error output topic*, you can created one like in the
+previous step.
```
dedemorton, line 113:
```suggestion
If you don't have an *Error output topic*, create one like you did
in the previous step.
```

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+====
+
+After filling the required parameters, click *Show Optinal Parameters* and add
+`audit` as the log type parameter.
+
+image::monitor-gcp-dataflow-optional-parameters.png[GCP Dataflow Optional Parameters]
+
+When you are all set, click *Run Job* and wait Dataflow to execute the
```
dedemorton, line 121:
```suggestion
When you are all set, click *Run Job* and wait for Dataflow to execute the
```

docs/en/observability/gcp-dataflow.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+template, which takes about few minutes.
```
dedemorton, line 122:
```suggestion
template, which takes a few minutes.
```

dedemorton (commented):
Thanks for your contribution. I've suggested some consistency edits. You can add them all to a batch and commit them at the same time to avoid adding a bunch of commits.

PR #1176, opened 2021-10-18
Add docs for elastic-agent diagnostics collect

michel-laterman:
Add documentation for the elastic-agent diagnostics collect subvommand.
This command produces an archive that can be used to help debug issued
with the elastic-agent.

### DIFF
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -31,11 +31,11 @@ run these commands.
 
 Gather diagnostics information from the {agent} and applications it's running.
 
-If no options are specified, this command will display version numbers and application meta-data.
+If no options are specified, this command displays version numbers and application metadata.
 
-If the `collect` option is specified, it will produce an archive with this meta-data, configuration information, the policy, as well as any local logs.
+If `collect` is specified, it produces an archive containing application metadata, configuration information, the policy, and any local logs.
 
-Note that the archive *will not redact credentials*, they may appear in plain text in the configuration or policy files inside the archive.
+Note that *credentials are not redacted* in the archive; they may appear in plain text in the configuration or policy files inside the archive.
 
 [discrete]
 === Synopsis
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/commands.asciidoc
```diff
@@ -31,21 +31,34 @@ run these commands.
 
 Gather diagnostics information from the {agent} and applications it's running.
 
-This command displays version numbers and application metadata.
+If no options are specified, this command will display version numbers and application meta-data.
```
dedemorton, line 34:
```suggestion
If no options are specified, this command displays version numbers and application metadata.
```

docs/en/ingest-management/commands.asciidoc
```diff
@@ -31,21 +31,34 @@ run these commands.
...
+
+If the `collect` option is specified, it will produce an archive with this meta-data, configuration information, the policy, as well as any local logs.
```
dedemorton, line 36:
```suggestion
If `collect` is specified, it produces an archive containing application metadata, configuration information, the policy, and any local logs.
```

docs/en/ingest-management/commands.asciidoc
```diff
@@ -31,21 +31,34 @@ run these commands.
...
+
+Note that the archive *will not redact credentials*, they may appear in plain text in the configuration or policy files inside the archive.
```
dedemorton, line 38:
```suggestion
Note that *credentials are not redacted* in the archive; they may appear in plain text in the configuration or policy files inside the archive.
```

docs/en/ingest-management/commands.asciidoc
```diff
@@ -31,21 +31,34 @@ run these commands.
...
+                                  [global-flags]
 ----
 
 [discrete]
 === Options
 
 `--output <string>`::
 Output format. Specify `human`, `json`, or `yaml`. The default is `human`.
+If using the `collect` option, specify `yaml`, or `json`. The default in this case is `yaml`.
+
```
dedemorton, line 58:
```suggestion
Output format. If using `collect`, specify `json` or `yaml` (the default).
Otherwise specify `json`, `yaml`, or `human` (the default).

```

dedemorton: Swapped things around because I think it's better to put something that has a condition first in the list (for folks who skim read).

For the other bits, I think it makes more sense to put the default setting at the end because users are less likely to explicitly set the value if it's a default. I don't love the parens, but maybe worth using here to avoid repeating "the default is" twice in the description. WDTY?

docs/en/ingest-management/commands.asciidoc
```diff
@@ -31,21 +31,34 @@ run these commands.
...
+`--file`::
+Output archive name for the `collect` option.
+Defaults to "elastic-agent-diagnostics-<timestamp>.zip" where the timestamp is the current time in UTC.
```
dedemorton, line 61:
```suggestion
Defaults to `elastic-agent-diagnostics-<timestamp>.zip` where the timestamp is the current time in UTC.
```

dedemorton (commented):
Some edits/comments. Thanks for adding this!

PR #1196, opened 2021-10-29
Update docs for metric threshold alert filter and group

Zacqary:
Adds updated documentation about the filtering and grouping fields on Metric Threshold alerts. Necessary to be able to link to the passage about redundant group-bys.


docs/en/observability/metrics-threshold-alert.asciidoc
```diff
@@ -34,14 +34,27 @@ To help you determine which thresholds are meaningful to you, the preview charts
...
-can also add multiple fields.
+[discrete]
+[[filtering-and-grouping]]
+== Filtering and grouping
+[role="screenshot"]
+image::images/metrics-alert-filters-and-group.png[Metric threshold filter and group fields]
+
+The *Filters* control the scope of the rule. If used, the rule will only evaluate metric data that matches the query in this field. In this example, the rule will only alert on metrics reported from a Cloud region called `us-east`.
+
+The *Groups alerts by* creates an instance of the alert for every unique value of the `field` added. For example, create a rule per host, or per every mount point of each host. You can also add multiple fields. In this example, the rule will individually track the status of each `host.name` in your infrastructure; if `host.name: host-1` crosses the threshold, but `host-2` and `host-3` do not, you will only receive an alert about `host-1`.
```
EamonnTP, line 50:
```suggestion
The *Groups alerts by* creates an instance of the alert for every unique value of the `field` added. For example, you can create a rule per host or every mount point of each host. You can also add multiple fields. In this example, the rule will individually track the status of each `host.name` in your infrastructure. You will only receive an alert about `host-1, if `host.name: host-1` passes the threshold, but `host-2` and `host-3` do not.
```

docs/en/observability/metrics-threshold-alert.asciidoc
```diff
@@ -34,14 +34,27 @@ To help you determine which thresholds are meaningful to you, the preview charts
...
+
+When you select *Alert me if a group stops reporting data*, the rule is triggered if a group that previously reported metrics does not report them again over the expected time period.
+
+[IMPORTANT]
+==============================================
+If you include the same field in both your **Filter** and your **Group by**, you may receive less results than you're expecting. For example, if you have a Filter of `cloud.region: us-east`, then grouping by `cloud.region` will have no effect, because the filter query is only able to match one region.
```
EamonnTP, line 56:
```suggestion
If you include the same field in both your **Filter** and your **Group by**, you may receive fewer results than you're expecting. For example, if you filter by `cloud.region: us-east`, then grouping by `cloud.region` will have no effect because the filter query can only match one region.
```

EamonnTP (approved):
Thanks @Zacqary looks good. Just a couple of minor suggestions.

EamonnTP:
Hi @Zacqary It looks like this PR needs back ported to the 8.0 and 7.16 branches. Let me know if you would like me to take care of that for you. Thanks.

Zacqary:
@EamonnTP Sorry forgot we didn't have auto-backport in this repo. Just pushed.

PR #1201, opened 2021-11-01
Enhance k8s docs regarding provider configuration and metadata addition

ChrsMark:
This PR will cover all the features that k8s provider currently has:
1. complete configuration option list
2. list of metadata that are available per resource

Closes https://github.com/elastic/integrations/issues/1743.

@dedemorton @bmorelli25 @mukeshelastic this is more or less the content that I plan to add. However with this addition the k8s provider part will look really big and maybe not so handy. Could we consider a re structure here? Maybe have one page per provider so as k8s provider to stand on its own and be easily discoverable. Let me know what you think.


dedemorton:
I think that's a good idea (having a topic for each provider). There is a limit to the number of nesting levels allowed in the TOC, so I would pop the content up one level in the hierarchy. Let me know if you need help with the asciidoc.

I would also highly recommend that you make each topic a separate source file (easier to update and move stuff around...plus it will be required in the next docs system).

akshay-saraswat:
@ChrsMark I discussed the structure with @dedemorton and we are good to go with the structure that is proposed in the attached screenshot.

Screen Shot 2021-11-02 at 12 34 35 PM

ChrsMark:
@dedemorton I tried to move the provider's part to a separate level but I was wondering how I could see the actual preview of it? Any way to build the docs locally like what we were doing with beat's docs? 

Also How I could add an extra level so as to have the various providers a list under the providers' level?

dedemorton:
@ChrsMark The instructions for building the docs locally are here: https://github.com/elastic/docs/blob/master/README.asciidoc#building-documentation

The steps describe the doc build alias file. I would use that. Make sure you pull down the latest version of the `docs` repo. You'll also need to pull down the `apm-server` repo because the build has a dependency (notice the dependency is shown in the resource setting): `alias docbldim='$GIT_HOME/docs/build_docs --doc $GIT_HOME/observability-docs/docs/en/ingest-management/index.asciidoc --resource=$GIT_HOME/apm-server/docs --chunk 2'`

I always need to reload my bash profile after updating the alias file:

`source ~/.bash_profile`

Then you can run the build from the command line:

`docbldim --open`

`--open` is optional and opens the book for a preview

dedemorton:
@ChrsMark Now I'll respond to your question about how to add another level to the navigation. It's very difficult with our current build system because the setting that controls nesting levels (chunk) is applied to ALL versions of the book. You also need to understand the interplay between section levels and `[discrete]` tags to figure out how the navigation will look in the end.

The navigation will support 4 levels (I think), so we could try setting the chunk level to 3 (instead of 2), but we will definitely need to add a bunch of `[discrete]` tags (where they're missing) to prevent extra files, and then we'll need to backport those changes all the way to 7.8. The book structure has changed a lot, so it won't be a clean backport. We'll probably need to do manual cleanup a few times.

It's not a huge amount of work (maybe a day?). I wonder if it's worth the effort though. WDYT?

ChrsMark:
Thank you for the detailed information here @dedemorton ! 
I'm wondering why it is so complicated to handle the docs and why it would require so much effort for something that initially seems simple. Just to be clear my goal here is:
1. Add the extra content about the Provider so as to have a complete documentation about its features
2. Re-structure how providers are documented cause now we have a huge file with very big content inside which is not so handy.

Item 1 could happen easily since I already have the content but I also feel that without the restructure it won't be so useful. 

In this regard, based on your feedback I would suggest taking a step back again and  answering the question answered at https://github.com/elastic/observability-docs/pull/1201#issuecomment-958130689 again. We need a handy way to represent it without investing a lot of time to ship it. So maybe providers shouldn't go under inputs but under configuration of standalone agent directly (same level with inputs) since actually setting up a provider does not fall under input's section in config.

@akshay-saraswat could you revisit this approach with @dedemorton please?

dedemorton:
@ChrsMark Our legacy (monolithic) doc build makes minor changes complicated sometimes. Writers experience this pain practically every day, but the good news is that there is a new system in progress that will make all of this easier (someday).

I think it's fine to move the providers up a level. I mentioned that earlier, but I think I wasn't clear enough about the nesting limitation. If you want to provide a topic for each provider, your options are to move the content up a level or change the chunk level (which is a big pain).

We'll need to revisit the structure when we move the docs to the new system anyhow because the navigation will work quite differently.

ChrsMark:
Thanks @dedemorton! Yeap, I think we can move on with the "one level up" approach.

ChrsMark:
@elasticmachine, run elasticsearch-ci/docs rebuild

ChrsMark:
@dedemorton I applied what we discussed, it would be really helpful if you could have a look :). Also how I could trigger the preview build like the one happened at https://github.com/elastic/observability-docs/pull/1201#issuecomment-956297038?

dedemorton:

dedemorton:
@ChrsMark I've kicked off another PR build, but the build should run automatically when you push new commits. If your structural change doesn't appear in the new build, I'll take a closer look.

ChrsMark:
Is the link https://observability-docs_1201.docs-preview.app.elstc.co/guide/en/fleet/master/index.html that I would expect to see the changes? I checked this but the changes do not appear. Sorry if I'm missing something here.

docs/en/ingest-management/elastic-agent/configuration/providers/elastic-agent-providers.asciidoc
```diff
@@ -0,0 +1,103 @@
+[discrete]
```
dedemorton, line 1:
```suggestion
```

dedemorton: Ah, I see the problem. The discrete tag basically forces the build to ignore chunking for that section. When you commit my suggestion, the docs build will run again, and the preview will be updated.

The preview link doesn't change. In fact, if you're ever in a situation where you don't see a link for the docs preview, just look at the console output and search for `preview`. You'll find the link in the log.

ChrsMark: Thank you!

ChrsMark:
@dedemorton thank you for your guidance here. It looks like I have brought to the from that we discussed. Could you please review this since I think we are close to merging?


@MichaelKatsoulis could you also have a look in the k8s internals please?

docs/en/ingest-management/elastic-agent/configuration/providers/kubernetes-provider.asciidoc
```diff
@@ -0,0 +1,429 @@
...
+  #sync_period: 600s
+  #cleanup_timeout: 60s
+  resources:
+    pod:
+      enabled: true
+----
+
+`node`:: (Optional) Specify the node to scope {agent} to in case it
+cannot be accurately detected when running {agent} in host network
+mode.
```
MichaelKatsoulis, line 25: We could state the default value here and to the rest of the settings we haven't already done so

docs/en/ingest-management/elastic-agent/configuration/providers/kubernetes-provider.asciidoc
```diff
@@ -0,0 +1,429 @@
...
+`cleanup_timeout`:: (Optional) Specify the time of inactivity before stopping the
+running configuration for a container. This is `60s` by default.
+`sync_period`:: (Optional) Specify the timeout for listing historical resources.
+`kube_config`:: (Optional) Use the given config file as configuration for Kubernetes
+client. If kube_config is not set, the KUBECONFIG environment variable will be
+checked and will fall back to InCluster if not present.
```
MichaelKatsoulis, line 31: I am not sure how helpful the `will fall back to InCluster if not present` is to a user. Maybe we could state exactly the steps that are followed

docs/en/ingest-management/elastic-agent/configuration/providers/kubernetes-provider.asciidoc
```diff
@@ -0,0 +1,429 @@
...
+]
+----
+
+{agent} automatically prefixes the result with `kubernetes`:
+
+
+[source,json]
+----
+[
+    {"kubernetes": {"id": "1", "namespace": "kube-system", "pod": {"name": "kube-controllermanger"}},
```
MichaelKatsoulis, line 209:
```suggestion
    {"kubernetes": {"id": "1", "namespace": {"name": "kube-system"}, "pod": {"name": "kube-controllermanger"}},
```

docs/en/ingest-management/elastic-agent/configuration/providers/kubernetes-provider.asciidoc
```diff
@@ -0,0 +1,429 @@
...
+    {"kubernetes": {"id": "2", "namespace": "kube-system", "pod": {"name": "kube-scheduler"}},
```
MichaelKatsoulis, line 210:
```suggestion
    {"kubernetes": {"id": "2", "namespace": {"name": "kube-system"}, "pod": {"name": "kube-scheduler"}},
```

docs/en/ingest-management/elastic-agent/configuration/providers/kubernetes-provider.asciidoc
```diff
@@ -0,0 +1,429 @@
...
+|`kubernetes.namespace.name`
+|`string`
+|Namespace of the Pod
+
+|`kubernetes.namespace.uuid`
+|`string`
+|UUID of the Namespace of the Pod
+
+|`kubernetes.namespace.labels.*`
+|`string`
```
MichaelKatsoulis, line 97: labels and annotations are objects

MichaelKatsoulis (commented):
Great job @ChrsMark updating the structure. I left some small comments regarding the internals of k8s provider

MichaelKatsoulis (approved):

dedemorton:
@ChrsMark What level of review are you looking for? Since some of the content is just copied over, I think you probably don't want a line-by-line review. Let me know which sections require closer scrutiny, and I'll take a look. Thanks!

ChrsMark:
@dedemorton no need for thorough review, just have a look since you are already involved in this PR :).

dedemorton (approved):
LGTM.

This structure is about the best you can get in our current docs without a lot of extra churn.

ChrsMark:
Thank you @dedemorton, this one will target v8.0 of the product. Not sure if any backport is required.

PR #1206, opened 2021-11-01
mark logging.ecs and logging.json as deprecated

michel-laterman:
Mark `logging.ecs` and `logging.json` as deprecated in 7.16 as both settings will be disables (defaulted to true) in the 8.x releases as part of the Elastic Stack shipping ECS logs.


dedemorton (approved):
LGTM (just let me know if you need another review after you mark this as ready for review)

PR #1239, opened 2021-11-06
Add proxy setup docs for Elastic Agent and Fleet

### LINKED ISSUE #593: Proxy setup for agent
Let's document how users can set up and use proxies with Elastic Agent.

Related:
- https://github.com/elastic/beats/issues/25319
- https://github.com/elastic/beats/pull/25219#issuecomment-824490680
- https://github.com/elastic/obs-dc-team/issues/520

dedemorton:
Describes how to set up Elastic Agent and Fleet to communicate through a proxy server.

I have quite a few questions about these steps because I don't have the environment needed to test them. I also need help with examples. But hopefully the basic structure and info is correct.

Preview link: https://observability-docs_1239.docs-preview.app.elstc.co/guide/en/fleet/master/fleet-agent-proxy-support.html

Closes #593


docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,259 @@
...
+
+* <<host-proxy-env-vars>>
+* <<es-output-proxy-settings>>
+* <<cli-proxy-settings>>
+
+[discrete]
+[[host-proxy-env-vars]]
+=== Host environment variables
+
+//QUESTION: Do these only work on Linux and Unix? What about Windows?
```
michel-laterman, line 58: These should work with Windows as well

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,259 @@
+[[fleet-agent-proxy-support]]
+= Use a proxy server with {agent} and {fleet}
+
+Your organization’s security strategy and other considerations may require you
+to use a proxy server between some components in your deployment. Support is
+available in {agent} and {fleet} for connections through HTTP Connect (HTTP 1
+only) and Socks5 proxy servers.
+
+//QUESTION: Do we want to add more context around when users might want to
```
michel-laterman, line 9: I don't think we need to list why orgs might want to use a proxy

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,259 @@
...
+
+//QUESTION: What should this config actually look like ^^.
+
+For more information, refer to <<elastic-agent-configuration>>.
+
+[discrete]
+[[cli-proxy-settings]]
+=== Command-line flags when enrolling in {fleet}
+
+//QUESTION: It sounds like the user need to visit Fleet to load policies before
```
michel-laterman, line 158: If they are deploying everything on their own yes, they currently need to visit the fleet app on Kibana in order to generate the default policies

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,259 @@
...
+//running the command? Or can we just assume users will do that first before
+//getting to this step?
+
+Specify proxy settings on the command line when you are enrolling in {fleet} and
+the proxy configuration is not yet available. During enrollment, the proxy
+settings specified at the command line are copied to internal configuration
+files.
```
michel-laterman, line 165: Are the proxy settings copied to internal files? I think in my testing I still had to specify a proxy if one was between the agent and ES

dedemorton: I saw that in one of the GitHub issues, but it's possible that the info was old. It's hard to tell sometimes what is the actual implementation vs just discussion. :-/

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+
+[discrete]
+[[host-proxy-env-vars]]
+=== Host environment variables
+
+Set environment variables on the host to configure default proxy settings.
+The {agent} uses host environment settings by default if no proxy settings are
+specified elsewhere. You can override host proxy settings later when you
+configure the {agent} and {fleet} settings. The following environment variables
+are available on the host:
```
gbanasiak, line 50: It's not clear where to put these environment variables. For instance, if Elastic Agent runs as a service in Linux they should go into `/etc/sysconfig/elastic-agent` instead of system-wide shell settings such as `/etc/bashrc`.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
+[[fleet-agent-proxy-support]]
+= Use a proxy server with {agent} and {fleet}
+
+Your organization’s security strategy and other considerations may require you
+to use a proxy server between some components in your deployment. Support is
+available in {agent} and {fleet} for connections through HTTP Connect (HTTP 1
+only) and Socks5 proxy servers.
+
+//QUESTION: Do we want to provide a list of popular proxy servers that will work?
```
mostlyjason, line 9: Not without testing them, not sure we should couple it to basic docs.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
+[[fleet-agent-proxy-support]]
+= Use a proxy server with {agent} and {fleet}
+
+Your organization’s security strategy and other considerations may require you
+to use a proxy server between some components in your deployment. Support is
```
mostlyjason, line 5: For example, you may have a firewall rule that prevents endpoints from connecting directly to Elasticsearch. In this scenario, you'd have the Elastic Agent connect to a proxy, and the proxy connects to Elasticsearch through the firewall.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
mostlyjason: @ruflin @michel-laterman Where should users put these environment variables? A user on our community slack said "This is not very generic for a standard linux package given environment variables differ between redhat and debian-based OSs. we are testing with Ubuntu so it would be VERY unusual to even have a /etc/sysconfig directory." 

It looks like these docs are stalled on input from our engineering team. CC @nimarezainia in case we need to prioritize time to get this across the line. I've seen multiple people asking for these in the past few days.

mostlyjason: The user also wants to know where to set these for Windows as well. They tried setting env vars in system properties but they are unsure if that is the right place.

gbanasiak: This Powershell sequence worked for me in Windows:

```
$environment = [string[]]@(
  "HTTPS_PROXY=https://proxy-hostname:proxy-port",
  "HTTP_PROXY=http://proxy-hostname:proxy-port"
  )

Set-ItemProperty "HKLM:SYSTEM\CurrentControlSet\Services\Elastic Agent" -Name Environment -Value $environment

Restart-Service "Elastic Agent"
```

This sets `HKLM\SYSTEM\CurrentControlSet\Services\Elastic Agent\Environment` registry key of `REG_MULTI_SZ` type.

That's similar to https://www.elastic.co/guide/en/apm/agent/dotnet/current/configuration-for-windows-services.html

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+
+|===
+|Variable |Description
+
+|`HTTP_PROXY`
+|URL of the proxy server for HTTP traffic. 
+
+|`HTTPS_PROXY`
+|URL of the proxy server for HTTPS traffic. `HTTPS_PROXY` takes precedence over
+`HTTP_PROXY` for https requests.
```
gbanasiak, line 60: > `HTTPS_PROXY` takes precedence over `HTTP_PROXY` for https requests.

I find this wording misleading as it suggests `HTTPS_PROXY` is used for HTTPS, but if it's not present then `HTTP_PROXY` is used. That's not the case. HTTPS traffic will not use `HTTP_PROXY` variable based on my tests.

michel-laterman: +1 for this, `HTTP_PROXY` is not used for HTTPS requests

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: @mostlyjason, you're correct about the way rhel based and debian based packages differ. We should not be recommending a specific file for all distros. rhel based has `/etc/sysconfig`, debian i believe uses `/etc/default` (though i may be incorrect there).
Though I'm not sure if we should mention these here as they are OS specific, and "environment variable" is a more general concept.
Also there may be more then one way to set it for an OS (for instance, a linux instance running systemd can provide a dropin file as well)

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+
+[discrete]
+[[es-output-proxy-settings]]
+=== Proxy settings in the {agent} policy
+
+Configure proxy settings in the {agent} policy to override proxy settings
+specified by environment variables. For example, you can configure proxy
+settings in the {es} output to override host environment settings.
+
+//QUESTION: Is there a way to set the proxy URL for an input?
```
michel-laterman, line 80: I have not tested this, but I would expect it to be done as part of the input configuration on the policy?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+
+The following proxy settings are valid in the {agent} policy:
+
+//QUESTION: Where are these proxy settings valid for Elastic Agent? Just under
+//output? What about input (for services behind a proxy)? I'm not sure how you
+//would configure that if you're not using standalone.
```
michel-laterman, line 86: They can be added to `fleet.yml` (if enrolled in fleet) or `elastic-agent.yml` if in stand-alone.
For example, currently there is no way to define the artifact download proxy through the UI for agents registered through fleet, so the `agent.download.proxy_url` needs to be set in `fleet.yml`

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+----
+outputs:
+  default:
+    type: elasticsearch
+    hosts:
+      - 'http://localhost:9200'
+    proxy_url: 'https://192.0.2.0:8080'
+----
+
+//QUESTION: What should this config actually look like ^^.
```
michel-laterman, line 136: looks correct

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: I agree with this

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: The `--proxy-url` value used when installing/enrolling in fleet is added as `fleet.proxy_url`, but it's not automatically  added to the `outputs` section (or `agent.download`).
Outputs settings can be added through the UI (in the settings panel).
However there's no easy way to set the `agent.download.proxy_url` so must be added manually to `fleet.yml`.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,250 @@
...
+//running the command? Or can we just assume users will do that first before
+//getting to this step?
+
+Specify proxy settings on the command line when you are enrolling in {fleet} and
+the proxy configuration is not yet available. During enrollment, the proxy
+settings specified at the command line are copied to internal configuration
+files.
+
+//QUESTION: What does it mean for these settings to be copied to the internal
+//config? Will they be persisted and used after enrollment?
```
michel-laterman, line 154: yes, they are persisted as part of the `fleet.yml` file.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: I'll just delete that sentence. It was copied from the issue in GitHub, but I agree that it's misleading/confusing.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: @michel-laterman I couldn't find out where to do this. Who can help me with this question? Should I just not mention setting this in the input (if it's not tested, IMO we shouldn't document it).

michel-laterman: I think that someone who worked on an integration may know more?
I haven't done testing of this feature.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
gbanasiak: > Though I'm not sure if we should mention these here as they are OS specific, and "environment variable" is a more general concept.

We did provide an example for APM in case of Windows: https://www.elastic.co/guide/en/apm/agent/dotnet/current/configuration-for-windows-services.html

Also RPM and DEB packages ship with legacy `init.d` files which reference `/etc/sysconfig/elastic-agent` (RPM) and `/etc/default/elastic-agent` (DEB) files explicitly. If it's not the right place for systemd then perhaps we should mention `Environment` option in the service file itself? Even better, why don't we ship with examples commented out?

```
[Service]
# Environment="HTTPS_PROXY=..."
```

In case of Windows we should also explain where to put environment variables for the Endpoint.

dedemorton: It doesn't feel like we've reached a consensus on where to set these environment vars (and it sounds like there's a lot of "it depends"  based on the OS, whether using systemd, etc). Can someone please distill this discussion into something I can use in the docs?  Preferably something that's been tested.

@EricDavisX My apologies if I've asked you this before, but has someone on the QA team tested the proxy setup? Maybe they can provide some insight here?

I'm pretty much blocked on moving forward until I get more info.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: @ruflin Do you know how to configure the proxy URL for an input? Of if that's possible? I'm not sure who else to ask.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: @gbanasiak, I like that idea.

@dedemorton, I did the initial proxy testing for the agent, however my efforts were centred around using fleet-managed agents.

I added environment variables in `/etc/environment` - however these would not be used by any service running under systemd. Those services would load environment variables from unit files like in the example provided by @gbanasiak above.
Currently on linux the service files will also refer to the legacy environment files located in places like `/etc/sysconfig` and `/etc/default`

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
ruflin: This depends on each input. Each input has its own configuration and likely in most cases a proxy config does not make sense. This is the same config that is used for modules, for example like here for the http module https://www.elastic.co/guide/en/beats/metricbeat/current/metricbeat-module-http.html If this module / input does not support a proxy, it can't be used.

I don't think we should document anything around this here as it is very input specific and I don't see how it fits into the overall proxy topic.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
craiglawson: Examples would be perfect in my opinion, the difficulty I had was finding any reference points, also thanks to @mostlyjason and @gbanasiak for your assistance so far :)

michel-laterman:
@dedemorton, we need to add a section to state that `agent.download.proxy_url` must be manually set in stand alone mode (in `elastic-agent.yml`), or **injected into** `fleet.yml` as there is no way to set this through the UI

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: Thanks, Nicolas. I won't mention inputs in this topic.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,252 @@
...
+
+* `outputs` - set the proxy server for communicating with {es}.
+
+* `fleet` - set the proxy server for retrieving agent policies from
+{fleet}.
+
+* `agent.download` - set the proxy server for retrieving artifacts or binary
+upgrades from `artifacts.elastic.co`.
+
+//TODO: Clarify where to add these settings ^^. For Fleet-managed agents, it
```
dedemorton, line 116: @michel-laterman See my comment here. Can you help me with some examples that show these settings for Fleet-managed and self-managed agents (including YAML configs, which file they're in, screen captures, if relevant). I feel I am not connecting the dots completely here. I suspect the flow of info on this page isn't quite right either but it's hard to know without having all the technical details.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,252 @@
...
+`--proxy-header` flag multiple times to add additional headers. You can use
+this setting to pass keys/tokens required for authenticating with the proxy.
+
+|===
+
+For example:
+
+[source,sh]
+----
+//TODO: Add an example here
```
dedemorton, line 184: @michel-laterman Can you also provide a realistic command-line example?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,252 @@
...
+----
+
+[discrete]
+[[es-proxy-setting]]
+=== Access {es} through a proxy
+
+When {fleet} is behind a proxy server, configure {fleet} settings to
+specify the URL for the proxy server:
+
+//QUESTION: Is correct? There does not appear to be a Proxy URL setting under
```
dedemorton, line 233: @michel-laterman This is another section where I have questions. Are the steps here correct? Can you provide a screen capture?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,252 @@
...
+--
+* *{es} hosts*
+
+* *{fleet-server} hosts*
+--
++
+//TODO: Add screen capture that shows valid settings in this dialog.
+
+
+//QUESTION: Unrelated to this section, but do we need to say anything about
```
dedemorton, line 251: @michel-laterman I don't want to complicate this content, but I think certs are going to make this configuration more difficult for users. Do you think we should say anything in this topic about certs?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
jlind23: @dedemorton do you have enough content with @michel-laterman comment?
https://github.com/elastic/observability-docs/pull/1239#discussion_r759515626

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: OK, I think I've leave certs out for now. If we get a lot of questions, we can revisit this decision.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,252 @@
...
+|===
+
+Proxy settings are valid in the {agent} policy under the following sections:
+
+* `outputs` - set the proxy server for communicating with {es}.
+
+* `fleet` - set the proxy server for retrieving agent policies from
+{fleet}.
+
+* `agent.download` - set the proxy server for retrieving artifacts or binary
```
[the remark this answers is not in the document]
dedemorton: @michel-laterman One follow-up question about your comment: "...(agent.download.proxy_url) needs to be injected into fleet.yml when in fleet mode." When you say "needs to be injected into fleet.yml," do you mean that users need to modify that file manually? What happens if they update the policy if Fleet after adding this setting? Is the setting preserved, or do they need to go back and set it again?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: @jlind23 Yes, this is great. I have one follow-up question for Michel, but I can start making updates with the info I have now. Thanks everyone for following up on this.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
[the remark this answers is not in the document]
michel-laterman: Yes, the user needs to manually add it to the file.
It does not get removed when policies update/the agent restarts
I've tested restarting the elastic-agent service (`systemctl restart elastic-agent.service`), calling `elastic-agent restart`, and updating the policy;

michel-laterman (commented):
confirmed that it works as I've detailed in 7.15.2

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,310 @@
...
+|`NO_PROXY`
+|IP addresses or domain names that should not use the proxy. Supports patterns.
+|===
+
+The proxy URL can be a complete URL or `host[:port]`, in which case the `http`
+scheme is assumed. An error is returned if the value is a different form.
+
+[discrete]
+[[where-to-set-proxy-env-vars]]
+==== Where to set proxy environment variables
```
dedemorton, line 72: @michel-laterman Can you take a look at this section about where to set the environment variables? I had to do a bit of digging and want to make sure I got the details right. I didn't include macOS since I don't have the details about that. Maybe we can add that later. I'll make the other changes on Monday because it's late and I'm out of steam.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,310 @@
...
++
+Then add the environment variables under `[Service]`
++
+[source,shell]
+----
+[Service]
+
+Environment="HTTPS_PROXY=https://my.proxy:8443"
+Environment="HTTP_PROXY=http://my.proxy:8080"
+----
```
michel-laterman, line 129: (nitpick) I'de move this chunk before the explanation of `/etc/default` and `/etc/sysconfig` as I think more systems use systemd

michel-laterman (approved):
environment section looks fine to me. I think we're just missing a section about adding `agent.download.proxy_url` but we're fine otherwise

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
mostlyjason: @michel-laterman Recent versions of ubuntu also use systemd, so should an ubuntu user put them in `[Service]` or in `/etc/default/elastic-agent`?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,252 @@
...
+| (boolean) If set to `true`, all proxy settings, including the `HTTP_PROXY` and
+`HTTPS_PROXY` environment variables, are ignored.
+
+|===
+
+Proxy settings are valid in the {agent} policy under the following sections:
+
+* `outputs` - set the proxy server for communicating with {es}.
+
+* `fleet` - set the proxy server for retrieving agent policies from
```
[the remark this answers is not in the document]
mostlyjason: @michel-laterman should users modify the fleet.yml file directly? If this file is autogenerated is there a chance it could get overwritten? If users should not modify it, should we remove this line and tell users that `--proxy-url` flag at enrollment is the desired interface?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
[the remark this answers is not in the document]
mostlyjason: Fleet is supposed to provide a way to centrally manage agents, which means users should not have to modify files manually on each host. I see this file more like an internal file than something users edit directly. It seems like the Kibana UI would provide a better solution when its available https://github.com/elastic/kibana/issues/100413. However, that means the future UI may override these user settings. I'm a little uncomfortable documenting this if it could lead to breaking changes later. If that's the case, we could defer documenting this until we have a more fully designed solution. What do you think @joshdover?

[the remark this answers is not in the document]
mostlyjason: I believe a workaround is that users could set a proxy server via environment variable. The question is here is whether we want to allow users to set a different proxy server for the downloads from other connections, which I think is a less common use case.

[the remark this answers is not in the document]
bataya0: > 

It would be beneficial to my situation  if I could set a different proxy value  for this,  different to some of the other proxy settings I use/set in other areas of  the yml.   

As long as the documentation is clear on its intention people will be happy.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,310 @@
...
+
+[discrete]
+[[fleet-proxy-configuration]]
+== Configure {fleet} proxy settings
+
+Configure proxy settings for {fleet} when it must connect through a proxy server
+to:
+
+* Access the Elastic Package Registry
+* Communicate with {es}
```
mostlyjason, line 258: This point is about how Kibana can use a proxy server to access ES, so it doesn't have anything to do with Fleet specifically. Should we remove this line? Actually, we could remove the whole section since the only other bullet is EPR and that has its own section below.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,310 @@
...
+----
+
+NOTE: These commands require default policies to be loaded in {fleet}. Default
+policies are loaded automatically when you visit {fleet} for the first time. If
+you're not sure whether default policies are loaded, log in to {kib} and go to
+*Management > {fleet}*.
+
+[discrete]
+[[fleet-proxy-configuration]]
+== Configure {fleet} proxy settings
```
mostlyjason, line 252: This is about the Fleet app in Kibana? Might help to be more explicit in the title

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
mostlyjason: @dedemorton Are you talking about Fleet Server here? It might be helpful to specify the components being connected to be more clear.

mostlyjason: Also @michel-laterman what type of proxies is this section relevant for? Would users generally configure a reverse proxy for ES and FS? If so, it might help to be more explicit. If its a forward proxy, then users would need to set `--proxy-url`? I am not an expert at this, so please correct me if I'm wrong.

mostlyjason:
@dedemorton as a general comment, I feel like we should test the steps in this doc to make sure all the functionality works. Are you doing that or should we ask the QA team to try it? One advantage of asking QA to do it, is that we might discover steps that need more clarification.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,310 @@
...
+* Send data to {es}
+* Retrieve agent policies from {fleet-server}
+* Retrieve agent policies from {es} (only needed for agents running {fleet-server})
+* Monitor services (not shown)
+
+image::images/agent-proxy-server.png[Image showing connections between {agent}, {fleet-server}, and {es}]
+
+You can configure {agent} proxy settings by specifying:
+
+* <<host-proxy-env-vars>>
```
mostlyjason, line 40: Can we say why a user would choose one over the other? Eg. Env vars are good when you want to proxy everything on the agent or even across the whole host. The agent policy really only lets you control the proxy to the ES output (I'm proposing we remove the other cases below). The CLI commands let you override environment variables set at the system level and specify a different proxy for agent. @michel-laterman does that sound correct?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,310 @@
...
+----
+
+After adding environment variables, restart the service.
+
+[discrete]
+[[proxy-settings-in-agent-policy]]
+=== Proxy settings in the {agent} policy
+
+Proxy settings in the {agent} policy override proxy settings specified by
+environment variables. The following proxy settings are valid in the agent
```
mostlyjason, line 138: The environment variables may be set at the system or host level, and this gives you the option to specify a different proxy for the Elastic Agent.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
mostlyjason: Nevermind, it looks like these scenarios are described below.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: @mostlyjason To avoid passive voice, is this OK?

This means you can  specify proxy settings for {agent} that are different from host or system-level environment settings.

(I'm afraid this statement is a little redundant, but I'm trying to rephrase your suggestion to avoid passive voice and make the statement clearer.)

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: I think so, unless users need to specify the proxy url under Fleet hosts. @michel-laterman can you clarify here?

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: I think using systemd syntax as our first documented option is better as it's applicable to rhel/debian based instances.
I believe Our service files for .deb and .rpm installers should have (optional) references to the distro specific files (like `/etc/default/elastic-agent`)

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: @mostlyjason, the agent may need a proxy to communicate with fleet and another proxy to communicate with ES

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
michel-laterman: the fleet-app in kibana can be used to set the output proxy (i.e., `agent->proxy->es`)

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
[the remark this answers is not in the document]
dedemorton: I prefer to document this, along with the caveat that the setting might get overridden in the future. IMO it will save users and the support team some effort in figuring out how to do this until we get to the point where it's properly supported in the UI.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: Marking this as "resolved" because my next commit will include a lot of changes to this section in response to your comments.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: I'll mark this as resolved for now and you can add corrections when I push the commit.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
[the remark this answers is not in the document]
dedemorton: I've removed this bullet point in favor of showing how to set this from the command line.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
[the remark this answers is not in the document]
dedemorton: I'm going to mark this as closed, and you can comment on the commit after I push it.

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,314 @@
...
+    hosts:
+    - https://10.0.1.2:9200
+    proxy_url: http://10.0.1.7:3128
+    type: elasticsearch
+----
+
+For more information, refer to <<elastic-agent-configuration>>.
+
+[discrete]
+=== Set the proxy for downloading artifacts
```
mostlyjason, line 205: (Experimental)

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -0,0 +1,314 @@
...
+== Set the proxy URL of the Elastic Package Registry
+
+{fleet} might be unable to access the Elastic Package Registry because {kib} is
+behind a proxy server.
+
+Also your organization might have network traffic restrictions that prevent {kib}
+from reaching the public Elastic Package Registry endpoints, like
+https://epr.elastic.co/[epr.elastic.co], to download package metadata and
+content. You can route traffic to the public endpoint of EPR through a network
+gateway, then configure proxy settings in the {kib} configuration file,
```
mostlyjason, line 307: Link to https://www.elastic.co/guide/en/kibana/current/fleet-settings-kb.html

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
mostlyjason: Looks like Dede's refactor addresses this point, which is mainly to reorganize the content

mostlyjason (approved):

mostlyjason:
@EricDavisX Adding you here since it'd be great for the QA to test it once its done. It sounds like Dede didn't have the proper test environment, and the original engineer who worked on it is no longer with the company.

CC @nimarezainia since this is a control plane feature

docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
dedemorton: good idea

michel-laterman (approved):
lgtm

dedemorton:
@elasticmachine run elasticsearch-ci/docs skiplinkcheck

PR #1247, opened 2021-11-08
[monitor-gcp] create Pub/Sub subscription in both GCP tutorials

andremm:
I just realized that the steps for creating the GCP Pub/Sub subscription should be used in both **Monitor Google Cloud Platform** and **GCP Dataflow templates** tutorials. I'm sending this PR to amend this.


dedemorton (approved):
Makes sense! LGTM (with the caveat that I haven't tested the steps; I've just confirmed that the text looks right).

andremm:
thanks @dedemorton ! I just confirmed the steps.

Issue #996, opened 2021-08-23
Automatically upgrade integration policies


mostlyjason:
## Description

We are planning to add the ability for users to automatically upgrade integration policies when they upgrade an integration. We'd like to document this new capability, and what users should do if the upgrade fails. 

Implementation issue: https://github.com/elastic/kibana/issues/106048

## Collaboration

* [ ] The docs team will lead producing the content
* [x] The product team will provide the initial content and the docs team will edit / review
* [ ] The docs team will define with product team the structure and location, and the product team will provide the initial content
* [ ] Other (please describe)

**Contact Person:**

*(We need to have a contact person in the product/development team to provide information about how the item to be documented works.)*
@kpollich 

## Suggested Target Release

7.16

## Stakeholders

@jen-huang @mostlyjason

andresrc:
@mostlyjason can you describe the required collaboration. Please use the issue template.

mostlyjason:
@andresrc updated

dedemorton:
Adding my name to this issue because it's next up in my work queue.

amolnater-qasource:
Hi @kpollich
We have tested automatic integration upgrade on 7.16.0 BC-5 Kibana cloud environment
- We disabled the automatic package upgrade for System integration.
- Degraded the System integration version to `v0.12.7`.
- Added the integration to Policy.
- Still System integration upgraded to latest version however policies didn't upgrade.

**Question:**
Is this feature only restricted for integration in policies whether to upgrade automatically or not?
Is this expected that on disabling automatic upgrade, actual system integration would get back to latest version automatically?

cc: @mostlyjason and @EricDavisX
Thanks

kpollich:
> Hi @kpollich We have tested automatic integration upgrade on 7.16.0 BC-5 Kibana cloud environment
> 
> * We disabled the automatic package upgrade for System integration.
> * Degraded the System integration version to `v0.12.7`.
> * Added the integration to Policy.
> * Still System integration upgraded to latest version however policies didn't upgrade.
> 
> **Question:** Is this feature only restricted for integration in policies whether to upgrade automatically or not? Is this expected that on disabling automatic upgrade, actual system integration would get back to latest version automatically?
> 
> cc: @mostlyjason and @EricDavisX Thanks

[...]

PR #1254, opened 2021-11-11
Break topic about managing integrations into multiple topics

dedemorton:
Originally we put all the content about managing integrations into a single topic based on how we thought the next docs system would work.

The content I'm adding for #996, however, will make the page even longer. To make the content usable in the existing system where the TOC is not sticky, I'm breaking this page up into multiple pages.

NOTE: I've changed a few IDs for consistency, but there is no new content here. Please do not review the content, just the structure.

Preview link: https://observability-docs_1254.docs-preview.app.elstc.co/guide/en/fleet/master/integrations.html


ollyhowell (approved):
LGTM!
Much better as a section rather than a single page.

bmorelli25 (approved):
LGTM2

PR #1159, opened 2021-10-13
Docs for --fleet-server-es-insecure

michalpristas:
Flag introduced here https://github.com/elastic/beats/pull/28007

PR #1263, opened 2021-11-15
Indicate --url is required when enrolling a fleet server with a cert

dedemorton:
Closes https://github.com/elastic/beats/issues/27245

This is a loose backport of https://github.com/elastic/observability-docs/pull/1091 without most of the changes/improvements from the later PR. My main goal is to get the syntax right for 7.14 with minimal changes since we don't currently release doc bug fixes for 7.14 and earlier.

### DIFF
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -88,13 +88,18 @@ elastic-agent enroll --fleet-server-es <string>
                      [--fleet-server-port <uint16>]
                      [--force]
                      [--help]
-                     [--insecure ]
+                     [--url <string>] <2>
                      [global-flags]
 ----
 <1> If no `fleet-server-cert*` flags are specified, {agent} auto-generates a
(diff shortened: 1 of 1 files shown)


ppf2:
@dedemorton I see that the changed file is replacing `--insecure` option with `--url`.  Did we mean to add the `--url` option, not to replace `--insecure`?  I think it is best for a Fleet Server developer to be the technical reviewer to confirm the current/correct settings.  Thx!

dedemorton:
@ppf2 Thanks for looking! For 7.14, the `--insecure` option wasn't used when enrolling a fleet server and shouldn't have been included in the syntax or example. It was causing confusion about where the flag was being applied.

Recently we did introduce a `--fleet-server-es-insecure` flag (https://github.com/elastic/beats/pull/28007 and https://github.com/elastic/observability-docs/pull/1159), but that's not available in 7.14.

@michalpristas -- Can you review this change for 7.14?

PR #1265, opened 2021-11-16
Add agent.monitoring.pprof doc

michel-laterman:
Add some explanation for the agent.monitoring.pprof option, and when to
disable it, in the stand-alone documentation.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
@@ -22,7 +22,7 @@ agent.monitoring:
   # enables metrics monitoring
   metrics: true
   # exposes /debug/pprof/ endpoints
-  # recommended that these endpoints are only enabled if the monitoring endpoint is set to localhost
+  # enable these endpoints if the monitoring endpoint is set to localhost
   pprof: true
   # specifies output to be used
   use_output: monitoring
@@ -39,5 +39,5 @@ collected. If neither setting is specified, monitoring is turned off. Set
 
 The `agent.monitoring.pprof` option controls whether the {agent} exposes the
 `/debug/pprof/` endpoints with the monitoring endpoints. It is set to `true`
-by default. It is recommended that these endpoints are disabled on the agent
-if the monitoring endpoint is accessible over a network (not recommended).
+by default. If the monitoring endpoint is accessible over a network (not recommended),
+set this option to `false` to disable the `/debug/pprof/` endpoints.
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
```diff
@@ -21,6 +21,9 @@ agent.monitoring:
   logs: true
   # enables metrics monitoring
   metrics: true
+  # exposes /debug/pprof/ endpoints
+  # recommended that these endpoints are only enabled if the monitoring endpoint is set to localhost
```
dedemorton, line 25:
```suggestion
  # enable these endpoints if the monitoring endpoint is set to localhost
```

dedemorton: suggest shortening this a bit for skim readers.

docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
```diff
@@ -32,4 +35,9 @@ are ignored.
...
 `logs` and `metrics` settings to control whether logs, metrics, or both are
 collected. If neither setting is specified, monitoring is turned off. Set
-`use_output` to specify the output to which monitoring events are sent.
\ No newline at end of file
+`use_output` to specify the output to which monitoring events are sent.
+
+The `agent.monitoring.pprof` option controls whether the {agent} exposes the
+`/debug/pprof/` endpoints with the monitoring endpoints. It is set to `true`
+by default. It is recommended that these endpoints are disabled on the agent
+if the monitoring endpoint is accessible over a network (not recommended).
```
dedemorton, line 43:
```suggestion
by default. If the monitoring endpoint is accessible over a network (not recommended),
set this option to `false` to disable the `/debug/pprof/` endpoints.
```

dedemorton: Suggest changing sentence order a bit so that it's clear that "not recommended" applies to the first part of the sentence only.

Also...unless there are legit reasons why users would want to leave the endpoints enabled, I would just tell them to set the option to false. 

If there are legit reasons, then change my edit to:

If the monitoring endpoint is accessible over a network (not recommended),
we recommend setting this option to `false` to disable the `/debug/pprof/` endpoints.

One other point: by "is accessible" do you mean "is accessed". It's a subtle difference, but worth clarifying in the sentence if you really mean "is accessed"

dedemorton (commented):
thanks for the contribution. A few edits/suggests. Nice work!

dedemorton (approved):
LGTM. Thanks for contributing.

PR #1362, opened 2021-12-14
Fix link to air gapped registry docs

dedemorton:
Points to the new location for the docs about using integrations in an air-gapped environ.


dedemorton:
Had to remove the link entirely because the integrations guide is not versioned. There is no longer any content about the experimental standalone server in 7.15.

bmorelli25 (approved):

PR #1374, opened 2021-12-14
Add docs for --fleet-server-es-ca-trusted-fingerprint

### LINKED ISSUE #1308: Update Elastic Agent and Fleet docs to include ES CA fingerprints
For Standalone agent and managed agent.

**Owner**
@michel-laterman

michel-laterman:
Add documentation for the `-fleet-server-es-ca-trusted-fingerprint` flag in `elastic-agent`.
This flag will ensure that processes started by the agent trust the CA with the specified fingerprint.
The CA does not have to be an existing CA on the host (as it would for the existing sha256 option in libbeat).
This allows the agent to send data into an ES cluster with a self signed CA for 8.0

- Closes #1308


docs/en/ingest-management/commands.asciidoc
```diff
@@ -176,24 +176,26 @@ elastic-agent enroll --fleet-server-es <string>
...
+                     [--fleet-server-es-insecure ]
                      [global-flags]
 ----
 <1> If no `fleet-server-cert*` flags are specified, {agent} auto-generates a
 self-signed certificate with the hostname of the machine. Remote {agent}s
 enrolling into a {fleet-server} with self-signed certificates must specify
 the `--insecure` flag.
-<2> Required when enrolling in a {fleet-server} with custom certificates. The
+<2> SHA256 fingerprint of the CA that {es} generates for self-signed certificates.
+Required when using self-signed certificate on {es} side.
```
dedemorton, line 195:
```suggestion
<2> Required when using self-signed certificates with {es}.
```

dedemorton: I'd avoid providing a description in the callout. Otherwise, we'll end up with a bunch of callouts.

docs/en/ingest-management/commands.asciidoc
```diff
@@ -227,14 +229,18 @@ specified {es} URL.
 `--fleet-server-es-ca <string>`::
 Path to certificate authority to use to communicate with {es}.
 
+`--fleet-server-es-ca-trusted-fingerprint <string>`::
+The hash of {es}'s certificate authority that will used by
```
dedemorton, line 233: Sentence structure here makes the details a bit harder to follow. Also it's not clear to me where the user gets the fingerprint. Does elasticsearch-certutil spit it out when you generate self-signed certs? Will users know how to get this hash? Are there related ES docs we should be pointing to?

I'll suggest an edit here with the caveat that I'm not sure I have all the technical details quite right. 

```
The SHA256 fingerprint (hash) of the certificate authority used to self-sign {es} certificates.
This fingerprint will be used to verify self-signed certificates presented by {fleet-server} and
any inputs started by {agent} for communication. This flag is required when using self-signed
certificates with {es}.
```

Another questions: is this related to  `--ca-sha256`?

docs/en/ingest-management/security/certificates.asciidoc
```diff
@@ -111,6 +111,8 @@ for {fleet-server}.
 |The CA certificate to use to connect to {es}. This is the CA used to generate
 certs for {es} (see <<prereqs>>).
 
+// NOTE Should we mention that we can use the --fleet-server-es-ca-trusted-fingerprint flag here instead of the elasticsearch-ca.crt file?
```
dedemorton, line 114: Yes, we should.

dedemorton (commented):
A couple comments. I don't deal with certs very often, so I need to re-learn concepts whenever I do. I wonder if @lockewritesdocs has a couple minutes to check my edits to make sure they are consistent with the security docs.

docs/en/ingest-management/commands.asciidoc
```diff
@@ -176,24 +176,26 @@ elastic-agent enroll --fleet-server-es <string>
                      [--fleet-server-cert <string>] <1>
                      [--fleet-server-cert-key <string>]
                      [--fleet-server-es-ca <string>]
+                     [--fleet-server-es-ca-trusted-fingerprint <string>] <2>
```
dedemorton, line 179: I think this should be shown in the install command, too, right? (I plan to put all the options into one reference topic at some point but haven't had time. For now, we need to duplicate.

docs/en/ingest-management/commands.asciidoc
lockewritesdocs: @dedemorton, your description looks correct. The only change I'd made is to add a hyphen for `SHA-256`. 

I have a related PR open for [connecting clients to Elasticsearch](https://github.com/elastic/elasticsearch/pull/81279) in 8.0+. When you start Elasticsearch for the first time, a CA certificate is generated and stored on disk, and the hex-coded SHA-256 fingerprint of this certificate is also output to the terminal. In the terminal will be printed something like this:

```
The hex-encoded SHA-256 fingerprint of the generated HTTPS CA DER-encoded certificate:
b9a10bbe64ee9826abeda6546fc988c8bf798b41957c33d05db736716513dc9c
```

This is the fingerprint of the CA certificate stored on disk, which is named `http_ca.crt`. Here's a list of the [security certificates](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/configuring-stack-security.html#stack-security-certificates) that are created and stored on disk.

dedemorton: @michel-laterman Maybe we should all more detail about where users find this file? Or point to the security docs.

docs/en/ingest-management/commands.asciidoc
michel-laterman: done

docs/en/ingest-management/security/certificates.asciidoc
```diff
@@ -111,6 +111,10 @@ for {fleet-server}.
 |The CA certificate to use to connect to {es}. This is the CA used to generate
 certs for {es} (see <<prereqs>>).
 
+Note that the CA certificate's SHA-256 fingerprint (hash) may be used instead of the `elasticsearch-ca.crt` file for securing connections to {es}.
+
+// NOTE Should we mention that we can use the --fleet-server-es-ca-trusted-fingerprint flag here instead of the elasticsearch-ca.crt file?
```
dedemorton, line 116:
```suggestion
```

dedemorton (approved):
LGTM. Made one minor suggestion (deleted the note).

PR #1388, opened 2021-12-16
[8.0] Delete CODEOWNERS

bmorelli25:
Since we branched 8.0 from main, the CODEOWNERS file that requires obs-docs team reviews in main now also exists in the 8.0 branch. Here’s a look at my email inbox since this started happening: :incoming_envelope: :chart_with_upwards_trend:.

I don't think we need to require code owner reviews in 8.0. What do you think?


dedemorton (approved):
Agreed!

bmorelli25:
Merged to stop the emails, but happy to reconsider if anyone disagrees.

PR #1409, opened 2021-12-20
Add deprecation notices for username/passwords for env vars

michel-laterman:
Relates to:
- https://github.com/elastic/fleet-server/pull/995
- https://github.com/elastic/beats/pull/29434


dedemorton:
@michel-laterman Sorry I didn't get to review this before I went on vacation. Typically we use the `deprecated` tag to mark deprecations so that the formatting is consistent and we can potentially use automation later on to remove deprecations. (We aren't using automation right now, so it's not a big deal, but we should use the tag for consistent formatting.)

See https://github.com/elastic/docs/blob/master/README.asciidoc#additions-and-deprecations

PR #1422, opened 2021-12-22
Add workaround for Fleet Server integration policy upgrade problem

### LINKED ISSUE #1294: [Request] Add Fleet Server Integration Policy Upgrades Known Issue
## Description

Add "known issue" under https://www.elastic.co/guide/en/fleet/current/fleet-troubleshooting.html for errors after upgrading Fleet Server integration policies in Cloud in 7.16.

Root cause issue: https://github.com/elastic/kibana/issues/119523

## Collaboration

- The product team (@kpollich) will provide the initial content and the docs team will edit / review

**Contact Person:** @kpollich 

## Suggested Target Release

7.16.x

## Stakeholders

@elastic/fleet

dedemorton:
Closes #1294 

Really sorry about the delay. This got lost in my inbox.

### DIFF
--- a/docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
@@ -24,7 +24,7 @@ Have a question? Read our <<fleet-faq,FAQ>>, or contact us in the
 
 [discrete]
 [[agents-in-cloud-stuck-at-updating]]
-== {agent}s hosted on {ecloud} are stuck in `Updating` or go `Offline`
+== {agent}s hosted on {ecloud} are stuck in `Updating` or `Offline`
 
 In {ecloud}, after <<upgrade-integration,upgrading>> {fleet-server} and its
 integration policies, agents enrolled in the Elastic Cloud agent policy
@@ -49,13 +49,6 @@ Copy the policy ID at the end of the URL.
 DELETE .kibana/_doc/ingest-package-policies:<integration_policy_id> <1>
 ----
 <1> `<integration_policy_id>` is the ID you copied in the previous step.
-
-. Clean the existing `.fleet-policies` doc. In the Dev Tools Console, run:
-+
-[source,console]
-----
-POST .fleet-policies/_delete_by_query?q=policy-elastic-agent-on-cloud
-----
 
 . Remove existing policy references to the {agent} cloud policy. In the Dev
 Tools Console, run:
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
```diff
@@ -22,6 +22,87 @@ Have a question? Read our <<fleet-faq,FAQ>>, or contact us in the
...
+. Force unenroll the agent stuck in `Updating`:
+
+.. To find agent's ID, go to *Fleet > Agents* and click the agent to see its
+details. Copy the Agent ID.
+
+.. In a terminal window, run:
++
+[source,shell]
+----
+curl '<kibana_url>/api/fleet/agents/<agentID>/unenroll' \ <1>
```
dedemorton, line 92: I took this command from another place, but the style is different from the previous command. I think we should use the same style, but I'm not sure which is preferred.

docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
```diff
@@ -22,6 +22,87 @@ Have a question? Read our <<fleet-faq,FAQ>>, or contact us in the
...
+curl -u <username>:<password> --request POST \
+  --url <kibana_url>/api/fleet/package_policies \
+  --header 'content-type: application/json' \
+  --header 'kbn-xsrf: xyz' \
+  --data '{"name": "Fleet Server","description": "","namespace": "default","policy_id": "policy-elastic-agent-on-cloud","enabled": true,"output_id": "","inputs": [{"type": "fleet-server","policy_template": "fleet_server","enabled": true,"streams": [],"vars": {"host": {"value": ["0.0.0.0"],"type": "text"},"port": {"value": [8220],"type": "integer"},"max_connections": {"type": "integer"},"custom": {"value": "server.runtime:\n  gc_percent: 20","type": "yaml"}}}],"package": {"name": "fleet_server","title": "Fleet Server","version": "1.0.1"},"force": true}'
+----
+
+. Force unenroll the agent stuck in `Updating`:
+
+.. To find agent's ID, go to *Fleet > Agents* and click the agent to see its
```
dedemorton, line 85: This did not actually work for me. When I tried this, I saw the following message, which didn't disappear until after I managed to unenroll the agent that was hanging. I was able able to get the agent ID by copying the URL link to the details page, but maybe there is a better way.

Is this error expected when you view details for a "stuck" agent?

image

kpollich: Hmm no, this isn't expected. I'm not sure where this would be coming from. Maybe @nchaulet has an idea? I will try to reproduce this as well.

docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
kpollich: I think the style above in the fleet server policy recreation step is very slightly preferential to this one, as it explicitly provides the `POST` HTTP verb. We can blame this on `curl`'s implementation choice of having a thousand ways to do the same thing 🙃

docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
dedemorton: @kpollich Were you able to reproduce this problem?

dedemorton: (I was not able to reproduce this problem on 7.16.2, but didn't try very hard.)

kpollich: No I wasn't able to reproduce in 7.16.2 either after a few tries. Might have been something intermittent with the agent status. Probably something to keep in mind but not a blocker for this PR.

kpollich (approved):
Looks great to me! Thanks for getting this together - this will be super helpful to add to the SKI here: https://github.com/elastic/support-known-issues/issues/993

Thanks so much again 🚀

PR #1442, opened 2022-01-10
[Fleet] Fix troubleshouting agent stuck in update to avoid ghost package policy

nchaulet:
## Description 

Update the instructions to fix Fleet Server agent stuck in updating to use Fleet API instead of devtools command to avoid ghost package policy.

Using the devtool command result in ghost package policy in the agent policy UI.


kpollich (approved):
Good catch, thank you

dedemorton (approved):
Assuming the build passes, LGTM!

Issue #1191, opened 2021-10-26
[Request][Fleet] Clarify licensing requirements for bulk management of agents


jen-huang:
## Description

On the page describing how to upgrade Elastic Agents via Fleet (https://www.elastic.co/guide/en/fleet/current/upgrade-elastic-agent.html#upgrade-agent) it says:
> To upgrade multiple agents, bulk select the agents and click Upgrade agents.

We should add licensing requirements for this as bulk management of agent in Fleet is a Gold+ feature.

## Suggested Target Release

As far back as possible - the licensing requirement has been in place since the feature was introduced.

## Stakeholders

@mostlyjason

PR #1443, opened 2022-01-11
Clarify licensing requirements for bulk agent upgrades

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/1191.

@jen-huang Can you please confirm that "selective agent binary updates" is the feature that covers bulk selection?  I'm not very keen on the phrasing used there, but I want to make sure we mention it in the docs.

Note that we don't document specific subscription levels in the docs because we point to the subscriptions page as a single source of truth about licensing requirements. (We've been told over and over again not to mention specific license levels in the docs.)


jen-huang:
> @jen-huang Can you please confirm that "selective agent binary updates" is the feature that covers bulk selection? I'm not very keen on the phrasing used there, but I want to make sure we mention it in the docs.

Yes it is (though I have likewise not seen this phrasing before, but it tracks with the other "selective" aka bulk features listed on the subscriptions page). Thanks DeDe!

jen-huang (approved):

PR #1454, opened 2022-01-12
Remove stale list of integrations from the Fleet/Agent docs

### LINKED ISSUE #1453: [Integrations] Update or remove the list of GA-level integrations in the Fleet/Agent guide
## Description

This topic has not been updated since 7.14: https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html

We really need to remove the list of GA-level integrations from this topic because integrations are released and promoted to GA on their own schedule. The docs will get stale quickly. We should point to https://docs.elastic.co/en/integrations as the source of truth. However the topic that lists GA integrations was removed because the script for creating it was not working.

For now, I think we should:

1. Remove the list of integrations from [Choosing between Elastic Agent and Beats](https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#choosing-between-agent-and-beats).
2. Tell users to go to https://docs.elastic.co/en/integrations to verify that the integrations they want to use are GA.

When the list of GA-level integrations is available in the docs, we can point to that instead of telling users to check individual integrations.

## Collaboration

**Contact Person:**

@bmorelli25 

## Suggested Target Release

7.15+

## Stakeholders

n/a

dedemorton:
Closes #1453

### DIFF
--- a/docs/en/ingest-management/beats-agent-comparison.asciidoc
+++ b/docs/en/ingest-management/beats-agent-comparison.asciidoc
@@ -37,10 +37,8 @@ Starting in version 7.14.0, {agent} is generally available (GA).
 The following steps will help you determine if {agent} can support your use case:
 
 . Determine if the integrations you need are supported and Generally Available
-(GA) on {agent}. To find out if an integration is GA, go to
-{integrations-docs}[{integrations}] and navigate to the docs for the
-integration. If there is no label, such as Beta or Experimental, the integration
-is GA.
+(GA) on {agent}. To find out if an integration is GA, see the
+{integrations-docs}/all_integrations[{integrations} quick reference table].
 
 . If the integration is available, check <<supported-outputs-beats-and-agent>> to see whether the required output is also supported.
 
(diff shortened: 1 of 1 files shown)


EamonnTP (approved):

bmorelli25 (approved):

docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -36,22 +36,12 @@ Starting in version 7.14.0, {agent} is generally available (GA).
...
-* `Windows`
-* `Zeek`
-+
-NOTE: For more details, refer to {integrations-docs}[{integrations}].
-+
+. Determine if the integrations you need are supported and Generally Available
+(GA) on {agent}. To find out if an integration is GA, go to
+{integrations-docs}[{integrations}] and navigate to the docs for the
+integration. If there is no label, such as Beta or Experimental, the integration
+is GA.
```
bmorelli25, line 43: > If there is no label, such as Beta or Experimental, the integration
is GA.

This isn't the greatest user experience. If you give me a few days, I can have the overview table working again. That would be better to link to, I think.

dedemorton: Agreed! That's awesome. I will wait.

bmorelli25, line 43:
```suggestion
(GA) on {agent}. To find out if an integration is GA, see the
{integrations-docs}/all_integrations[{integrations} quick reference table].
```

PR #1473, opened 2022-01-20
Add docs about debugging standalone Elastic Agents

dedemorton:
Closes #1473 

TODO:

- [ ] This PR also needs to be backported to 8.0 and 8.1, but I'll do that manually because the config setting for exposing /debug/pprof/ endpoints is different in 8.0.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
@@ -84,10 +84,6 @@ agent.logging.files:
 
 For other log settings, refer to <<elastic-agent-standalone-logging-config>>.
 
-//REVIEWERS: Should we also talk about exposing the agent monitoring endpoint
-//here? Sounds like that could be risky. We could point to the topic and
-//let users decides.
-
 [discrete]
 [[inspect-configuration]]
 == Inspect the {agent} configuration
--- a/docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
@@ -1,0 +1,96 @@
+[[debug-standalone-agents]]
+= Debug standalone {agent}s
+
+When you run standalone {agent}s, you are responsible for monitoring the status
+of your deployed {agent}s. You cannot view the status or logs in {fleet}.
+
+Use the following tips to help identify potential issues.
+
+Also refer to <<fleet-troubleshooting>> for guidance on specific problems.
+
+NOTE: You might need to log in as a root user (or Administrator on Windows) to
+run these commands.
+
+[discrete]
+== Check the status of the running {agent}
+
+To check the status of the running {agent} daemon and other processes managed by
+{agent}, run the `status` command. For example:
+
+[source,shell]
+----
+elastic-agent status
+----
+
+Returns something like: 
+
+[source,yaml]
+----
+Status: HEALTHY
+Message: (no message)
+Applications:
+  * metricbeat  (HEALTHY)
+                Running
+  * filebeat    (STOPPING)
+                (no message)
+----
+
+By default, this command returns the status in human-readable format. Use the
+`--output` flag to change it to `json` or `yaml`.
+
+For more information about this command, refer to
+<<elastic-agent-status-command>>.
+
+//REVIEWERS: Looks like JSON and YAML return status codes instead of healthy,
+//stopping, etc. Where can I find the mapping of status code to status?
+
+[discrete]
+[[inspect-standalone-agent-logs]]
+== Inspect {agent} and related logs
+
+If the {agent} status is unhealthy, or behaving unexpectedly, inspect the logs
+of the running {agent}. The log location varies by platform. Refer to
+<<installation-layout>> for the correct location.
+
+//REVIEWERS: Can we add more guidance here about how to explore logs, what to
+//look for, etc.
+
+[discrete]
+[[increase-log-level]]
+== Increase the log level of the running {agent}
+
+The log level of the running agent is set to `info` by default. At this level,
+{agent} will log informational messages, including the number of events that are
+published. It also logs any warnings, errors, or critical errors.
+
+To increase the log level, set it to `debug` in the `elastic-agent.yml` file.
+
+The `debug` setting configures {agent} to log debug messages, including a
+detailed printout of all flushed events, plus all the information collected at
+other log levels.
+
+Set other options if you want write logs to a file. For example:
+
+[source,yaml]
+----
+agent.logging.level: debug
+agent.logging.to_files: true
+agent.logging.files:
+  path: /var/log/elastic-agent
+  name: elastic-agent
+  keepfiles: 7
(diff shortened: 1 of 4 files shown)


docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+----
+
+For other log settings, refer to <<elastic-agent-standalone-logging-config>>.
+
+//REVIEWERS: Should we also talk about exposing the agent monitoring endpoint
+//here? Sounds like that could be risky. We chould point to the topic and
+//let users decides.
+
+//REVIEWERS: Do we want to cover the diagnostics command here? Not sure there
+//is much info here that's helpful to end users.
```
michel-laterman, line 92: I don't think it's too useful, it can show more process meta-data, or be used to access pprof endpoints (if endabled) but does not have a lot of additional functionality

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+
+//What about the inspect command? I think it could be useful to inspect the
+//config, but only if we are mentioning it during a longer procedure that
+//involves troubleshooting config problems.
```
michel-laterman, line 96: I think `inspect` and `inspect output` are more useful especially as `inspect output` shows what's passed to processes

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
```diff
@@ -0,0 +1,96 @@
...
+  name: elastic-agent
+  keepfiles: 7
+  permissions: 0600
+----
+
+For other log settings, refer to <<elastic-agent-standalone-logging-config>>.
+
+//REVIEWERS: Should we also talk about exposing the agent monitoring endpoint
+//here? Sounds like that could be risky. We chould point to the topic and
+//let users decides.
```
michel-laterman, line 89: Agreed, currently I think our best-practices is to bind it to localhost (or a npipe/unix socket)

dedemorton: @michel-laterman So are you saying that we should not talk about exposing the agent monitoring endpoint in this topic? I need more info and maybe a pointer to additional docs (or a meeting with you) because I still don't understand how monitoring works for standalone agents.

dedemorton:
@michel-laterman OK, so I've added a section about the inspect command, but I think this topic in general is missing the meaty guidance info that users need to know (like how to interpret what they're seeing in the logs). It's also missing info about how to analyze monitoring data because I'm not quite sure how users do that if they aren't using Fleet. Maybe if you can walk me through the steps, I can fill in this topic with more detail. Or perhaps you might want to push some updates  to my fork. WDYT?

Also note there are still some questions marked as comments `//` that need to be answered.

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
```diff
@@ -0,0 +1,203 @@
...
+[discrete]
+[[inspect-standalone-agent-logs]]
+== Inspect {agent} and related logs
+
+If the {agent} status is unhealthy, or behaving unexpectedly, inspect the logs
+of the running {agent}. The log location varies by platform. Refer to
+<<installation-layout>> for the correct location.
+
+//REVIEWERS: Can we add more guidance here about how to explore logs, what to
+//look for, etc.
```
michel-laterman, line 56: The `elastic-agent` logs would be in the folder specified by installation-layout, and beats/fleet-server would be in folders named by the output, i.e., `default`.
Endpoint is a special case as its logs are written in its install dir.

For debugging we generally start with looking for `error` level logs and investigate from there.
Repeated lines can show whatever strange behaviour is occurring (such as connection issues).
Sometimes there are no issues that can be found through the logs, and it may be that the agent or one of it's processes had an oom incident and the OS should be checked.

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
```diff
@@ -0,0 +1,203 @@
...
+----
+
+By default, this command returns the status in human-readable format. Use the
+`--output` flag to change it to `json` or `yaml`.
+
+For more information about this command, refer to
+<<elastic-agent-status-command>>.
+
+//REVIEWERS: Looks like JSON and YAML return status codes instead of healthy,
+//stopping, etc. Where can I find the mapping of status code to status?
```
michel-laterman, line 45: I think it's defined in the protob file: https://github.com/elastic/beats/blob/522c8624fe83d4a810af2d02bf3ecf6e10d0fa80/x-pack/elastic-agent/pkg/agent/control/proto/control.pb.go#L35-L44

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
```diff
@@ -0,0 +1,203 @@
...
+  name: elastic-agent
+  keepfiles: 7
+  permissions: 0600
+----
+
+For other log settings, refer to <<elastic-agent-standalone-logging-config>>.
+
+//REVIEWERS: Should we also talk about exposing the agent monitoring endpoint
+//here? Sounds like that could be risky. We could point to the topic and
+//let users decides.
```
michel-laterman, line 89: We should talk about the monitoring settings but repeat the recommendation not to expose the handler over a network (by default it binds to a unix socket/npipe)

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
dedemorton: @michel-laterman Thanks. BTW, it seems weird that when I run the `status` command, the response _sometimes_ includes the status in the message field, but sometimes it does not. That's my rationale for documenting the status codes. Let me know if this behavior is expected or a bug.

Here's an example of a response:

```
{
    "Status": 2,
    "Message": "",
    "Applications": [
        {
            "ID": "filebeat--7.17.0-a1bd5481",
            "Name": "filebeat",
            "Status": 0,
            "Message": "Starting",
            "Payload": null
        },
        {
            "ID": "metricbeat--7.17.0-732bc2d2",
            "Name": "metricbeat",
            "Status": 5,
            "Message": "",
            "Payload": null
        }
    ]
}
```

ph (approved):

docs/en/ingest-management/elastic-agent/debug-standalone-elastic-agent.asciidoc
ph: @michel-laterman is `the response sometimes includes the status in the message field, but sometimes it does not.` a known bug? Maybe the process aren't started yet?

dedemorton: @michel-laterman Can you open an issue if this is a bug? I'm going to go ahead and merge/backport this PR.

PR #1476, opened 2022-01-20
agent kibana env vars update deprecation notices

michel-laterman:
Update deprecation messages to indicate that the behaviour will change.


dedemorton (approved):
LGTM

PR #1487, opened 2022-01-21
Describe filter for showing Elastic Agent integrations

### LINKED ISSUE #1185: [meta] Update documentation to show unified integrations view
## Description

Starting in 7.16, the integrations view is changing and will include cards for Beats and other technologies (not just Elastic Agent integrations). We need to update the docs to reflect this change.

Related issue: https://github.com/elastic/docs.elastic.co/issues/1

## Collaboration

* [x] The docs team will lead producing the content
* [ ] The product team will provide the initial content and the docs team will edit / review
* [ ] The docs team will define with product team the structure and location, and the product team will provide the initial content
* [ ] Other (please describe)

**Contact Person:**

@goodroot 

## Suggested Target Release

7.16

dedemorton:
Closes #1185 

Dunno if this will really help much. I think we need to make the selector more prominent and maybe add some more visual distinction to the page to make it clear which integrations work with Elastic Agent. I think it's possible for users to end up in a weird situation where they work through Beats steps that aren't relevant.

Also when you're editing a policy and click *Add integration* the Integrations view should be constrained to integrations that work with Elastic Agent. @mostlyjason Do you know if there's any plan to do that? If not, I will open an issue to track these usability problems.


mostlyjason:
What is the user persona or task you are trying to address? There shouldn't be a need to filter on "Elastic Agent only" because we already recommend those when there is a GA integration available. I'm not sure we should recommend beta or experimental integrations unless the user really knows what they are doing.

> when you're editing a policy and click Add integration

When would a user do this? If they want to edit an existing integration policy, they should not click the Add button.

It would be useful to describe this filter, but can we describe the logic? We recommend Elastic Agent integration when there is a GA integration available. This is the best option for a user who does not have a preference for shipper. Existing Beats users who are not interested in Elastic Agent should click "Beats only". Likewise, Elastic Agent users who are not interested in Beats should click "Elastic Agent only".

dedemorton:
@mostlyjason In this task, the user is adding an integration to an existing agent policy. The only integrations that are valid in this particular context are Elastic Agent integrations, so it's useful to tell users to select Elastic Agent. It doesn't make sense here to describe the other options  because they aren't valid in this context. I'm planning to work through the quick start again and shift the focus towards starting with integrations, but haven't had a chance to do that yet. The quick start is probably a better place to describe this selector.

BTW, we document this procedure here for completeness because it's a valid task that you can perform from the Agent policies tab. If we don't want users to drill down into agent policies and add integrations, then we should remove the Add integration button: 

image

Personally, I think this button is useful for users who need to add an integration to an existing agent policy.

mostlyjason:
I see thanks for that context! This is making more sense to me now. This documentation page is in the Fleet and Elastic Agent guide, so only Elastic Agent integrations are relevant. In that case, it makes sense to filter on Elastic Agent only. It also means we are missing documentation on the Integrations page in Kibana. This is owned by the Journey team so I'll pass a note to them.

The button you pointed out does not have the ideal behavior. To add an integration to a policy, you'd expect to see only Elastic Agent integrations, but we don't provide a way to filter on those. You'd also expect that it would add it to the policy we are viewing. I don't think that context carries through automatically. The user needs to do some work to bridge these gaps. It'd be good to enhance this behavior in the future.

dedemorton:
@mostlyjason Is this OK to merge? If so, can you please approve. thanks!

mostlyjason (approved):

docs/en/ingest-management/integrations/add-integration-to-policy.asciidoc
```diff
@@ -4,8 +4,18 @@
...
 policy:
 
-. In {kib}, go to *Management > Integrations*. Search for and select an
-integration.
+. In {kib}, go to *Management > Integrations*.
++
+Notice that the Integrations page shows {agent} integrations along with other
+types, such as {beats}.
+
+. Scroll down and select *Elastic Agent only* so the view only shows
```
mostlyjason, line 12: so the view ~only~ shows

other non-agent and non-beats integrations are still shown

dedemorton, line 12:
```suggestion
. Scroll down and select *Elastic Agent only* so the view shows
```

dedemorton:

PR #1531, opened 2022-02-04
Use main for integrations/ecs-logging/elastic-package

v1v:
### What

Some repositories are now main based.

I don't know where these changes should be backported, so please feel free to add the right label :)


bmorelli25 (approved):
Thank you! Not backporting this is correct. I'll merge.

PR #1557, opened 2022-02-11
Add docs on upgrading Agent in air-gapped environment

joshdover:
We've gotten several questions about how to upgrade Agents via the UI when in an air-gapped environment. This is not yet supported until https://github.com/elastic/kibana/issues/100413 is implemented.

This adds a note to our air-gapped documentation about this limitation and how to workaround it.

Open questions:
- [ ] These instructions work for users who installed via the tarball distribution (not DEB or RPM). Do we need to include documentation for users who are using an internal DEB or RPM repository as well or should they already know what they're doing in that case?


joshdover:
Oh my hold on

joshdover:
Created a new PR on the correct base branch https://github.com/elastic/observability-docs/pull/1558

PR #1558, opened 2022-02-11
Add docs on upgrading Agent in air-gapped environment

joshdover:
We've gotten several questions about how to upgrade Agents via the UI when in an air-gapped environment. This is not yet supported until https://github.com/elastic/kibana/issues/100413 is implemented.

This adds a note to our air-gapped documentation about this limitation and how to workaround it.

Open questions:
- [ ] These instructions work for users who installed via the tarball distribution (not DEB or RPM). Do we need to include documentation for users who are using an internal DEB or RPM repository as well or should they already know what they're doing in that case?

### DIFF
--- a/docs/en/ingest-management/fleet/air-gapped.asciidoc
+++ b/docs/en/ingest-management/fleet/air-gapped.asciidoc
@@ -127,13 +127,13 @@ docker run -it -p 443:443 \
 
 [discrete]
 [[air-gapped-limitations]]
-== Limitations for Agent Upgrades
+== Limitations for {agent} upgrades
 
-The [[upgrade-elastic-agent]] feature in technical preview does not currently support network restricted environments without access to artifacts.elastic.co.
+The <<upgrade-elastic-agent>> feature in technical preview does not currently support network restricted environments without access to artifacts.elastic.co.
 To upgrade {agent}s:
 
 1. Download the new version from the https://www.elastic.co/downloads/elastic-agent[download page]
-2. Place the tarball on a secure internal network location
+2. Place the tarball in a secure internal network location
 3. Use the <<elastic-agent-upgrade-command,`upgrade`>> command with a custom `--source-uri` option
 
 [discrete]
--- a/docs/en/ingest-management/fleet/air-gapped.asciidoc
+++ b/docs/en/ingest-management/fleet/air-gapped.asciidoc
@@ -124,3 +124,20 @@ docker run -it -p 443:443 \
   -e EPR_TLS_CERT=/etc/ssl/package-registry.crt \
   docker.elastic.co/package-registry/distribution:{version}
 ----
+
+[discrete]
+[[air-gapped-limitations]]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -124,3 +124,22 @@ docker run -it -p 443:443 \
   -e EPR_TLS_CERT=/etc/ssl/package-registry.crt \
   docker.elastic.co/package-registry/distribution:{version}
 ----
+
+[discrete]
+[[air-gapped-limitations]]
+== Limitations for Agent Upgrades
```
dedemorton, line 130:
```suggestion
== Limitations for {agent} upgrades
```

docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -124,3 +124,22 @@ docker run -it -p 443:443 \
...
+
+The [[upgrade-elastic-agent]] feature in technical preview does not currently support network restricted environments without access to artifacts.elastic.co.
```
dedemorton, line 132:
```suggestion
The <<upgrade-elastic-agent>> feature in technical preview does not currently support network restricted environments without access to artifacts.elastic.co.
```

docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -124,3 +124,22 @@ docker run -it -p 443:443 \
...
+To upgrade {agent}s:
+
+1. Download the new version from the https://www.elastic.co/downloads/elastic-agent[download page]
+2. Place the tarball on a secure internal network location
```
dedemorton, line 136:
```suggestion
2. Place the tarball in a secure internal network location
```

docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -124,3 +124,22 @@ docker run -it -p 443:443 \
...
+3. Use the <<elastic-agent-upgrade-command,`upgrade`>> command with a custom `--source-uri` option
+
+[discrete]
+=== Example
+
+["source", "sh"]
```
dedemorton, line 142: You need `subs="attributes"` here, too, if you want `{version}` to resolve to a value. It will resolve to the current version, though, which might be a little weird. So maybe you want to leave this as-is.

dedemorton (approved):
Thanks for updating the docs. Just a few minor changes. I will approve pending acceptance of my suggestions. Thanks!

docs/en/ingest-management/fleet/air-gapped.asciidoc
dedemorton: yah actually that makes sense. Then you should change line 142 to say `["source", "sh", subs="attributes"]` (my thinking was a bit backwards yesterday for some reason).

PR #1559, opened 2022-02-11
Add experimental banner to Fleet API docs page

joshdover:
Until https://github.com/elastic/kibana/issues/123150 is complete we need to make this more clear that this is experimental. The page it links to indicates this too, but this needs to be more clear in the main docs.


dedemorton (approved):
LGTM

PR #1563, opened 2022-02-11
Delete .github directory

bmorelli25:
Removes code owners and issue template from `8.1`. This happened when branching `8.1`, but they should only exist in `main`.


dedemorton (approved):
LGTM

PR #1600, opened 2022-02-17
Correct improper capitalization

cachedout:
This is not the correct capitalization for the acronym which stands for Distributed Denial of Service.


cachedout:
> Do you want me to backport/forward port this for you?

I am happy to do it but would need a bit of guidance on which branches should be targeted. If it's easier for you to do it, that's fine too. :)

dedemorton:
I'll take it from here.

I'm gonna try the backport labels, but I don't think the PR needs to be based on main. I'll do a manual backport if necessary. Thanks for making the update!

PR #1616, opened 2022-02-24
Update install command for rpm/deb detection

michel-laterman:
Update the install command docs to correspond with rpm/deb detection changes in https://github.com/elastic/beats/pull/30289


dedemorton (approved):
Yay!! Do we want to change the installation instructions to show `install` instead of enroll, or is this change meant for folks who wrong `install` by mistake?

michel-laterman:
It's meant for people who use install by mistake or blindly copy the commands that are provided by Kibana

PR #1623, opened 2022-03-01
Update steps for APM Server Quick Start

kilfoyle:
While working through the steps in the [APM Quick Start](https://www.elastic.co/guide/en/apm/guide/master/apm-quick-start.html), I noticed that in the Cloud version of "Step 3: Add the APM integration", the steps don't match what I see in the Kibana UI, on a version 8.1 Cloud deployment. Specifically, Step 3.3 (Switch to Elastic Agent) and Step 3.4 (Confirm your choice) are no longer present, and the UI for Step 2.2 (Click APM integration) has changed.

Note: I was not able to build this PR locally to test, but I think that's just something with my local stack setup and not related to the contents of this PR.


bmorelli25 (approved):
Thanks, David! I confirmed that this is what it looks like in 8.0 as well — which is weird because I ran thorough this process for the 8.0 upgrade guide and it didn't look like this then.

PR #1640, opened 2022-03-03
Update the add agent graphic

kilfoyle:
I noticed that in this super fantastic [Ingest logs, metrics, and uptime data with Elastic Agent](https://www.elastic.co/guide/en/observability/8.0/ingest-logs-metrics-uptime.html) tutorial, the graphic for the Add Agent UI doesn't match what I see in Kibana on a version 8.1.0 deployment. I'm not sure when the UI changed but anyhow here's an updated version.


dedemorton (approved):
LGTM. Thanks!

I must have had my viewport on the wrong setting 'cause the original image is biG.

Issue #1489, opened 2022-01-24
[Request][Fleet] Account for removal of default agent policy 


jen-huang:
## Description

Account for removal of default agent policy, as implemented in 8.1 via https://github.com/elastic/kibana/issues/108456.

At minimum this page:
https://www.elastic.co/guide/en/fleet/current/install-fleet-managed-elastic-agent.html

## Collaboration

- The docs team will lead producing the content

**Contact Person:** @juliaElastic  

## Suggested Target Release

8.1

## Stakeholders

Fleet UI team

### Docs to update:

1. https://www.elastic.co/guide/en/fleet/current/add-a-fleet-server.html
Update Self managed tab: 
- add step to create Fleet Server policy (create button available on the form)
- update screenshot
- update elastic-agent install commands to remove -f flag (removed a while back)

2. https://www.elastic.co/guide/en/fleet/current/install-fleet-managed-elastic-agent.html
Installation steps:
- add step to create Agent policy (create button available on the form)
- update screenshot
- remove references of default agent policy from text (there will be no default created from 8.1)
- update elastic-agent install commands to remove -f flag (removed a while back)

3. https://www.elastic.co/guide/en/fleet/current/elastic-agent-container.html
Self-managed example, Docker compose example:
- remove references of default policy from text
- add `--env FLEET_SERVER_POLICY_ID=fleet-server-policy` parameter
- link to the new page below on how to create agent policies with API or preconfiguration

4. https://www.elastic.co/guide/en/fleet/current/running-on-kubernetes-managed-by-fleet.html
- Add parameter in the manifest file:
```
- name: FLEET_SERVER_POLICY_ID
- value: fleet-server-policy
```
- add a link to the new page below on how to create agent policies
- update text about getting default policy, FLEET_ENROLLMENT_TOKEN should not be empty

5. https://www.elastic.co/guide/en/fleet/current/fleet-overview.html
- update screenshot at Elastic Agent policies, to show the new UI of Add integration
- Add a link about the info on how to create agent policies

6. https://www.elastic.co/guide/en/fleet/8.0/fleet-agent-proxy-support.html
- update Note about default policies, change it to saying that policies have to be created first.

7. https://www.elastic.co/guide/en/fleet/8.0/agent-policy.html
- remove Note about Default policy

8. Create a new page to summarize the steps to create agent policies without going to the UI:

Option 1.   Add agent policy with the API

  ```
[...]

PR #1642, opened 2022-03-05
Add 8.1.0 Fleet and Elastic Agent release notes

dedemorton:
Please make sure I haven't missed anything for Elastic Agent. We should sort out a better way of handling the changelog now that Elastic Agent will be in a new repo.

### DIFF
--- a/docs/en/ingest-management/release-notes/release-notes-8.1.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-8.1.asciidoc
@@ -1,0 +1,174 @@
+// Use these for links to issue and pulls. 
+:kib-issue: https://github.com/elastic/kibana/issues/
+:kib-pull: https://github.com/elastic/kibana/pull/
+:agent-issue: https://github.com/elastic/beats/issues/
+:agent-pull: https://github.com/elastic/beats/pull/
+:fleet-server-issue: https://github.com/elastic/beats/issues/fleet-server/
+:fleet-server-pull: https://github.com/elastic/beats/pull/fleet-server/
+
+
+[[release-notes]]
+= Release notes
+
+This section summarizes the changes in each release.
+
+* <<release-notes-8.1.0>>
+
+Also see:
+
+* {kibana-ref}/release-notes.html[{kib} release notes]
+* {beats-ref}/release-notes.html[{beats} release notes]
+
+// begin 8.1.0 relnotes
+
+[[release-notes-8.1.0]]
+== {fleet} and {agent} 8.1.0
+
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/release-notes/release-notes-8.1.asciidoc
```diff
@@ -0,0 +1,174 @@
...
+folder `{path.config}/inputs.d`. {agent-pull}30087[#30087]
+
+[discrete]
+[[enhancements-8.1.0]]
+=== Enhancements
+
+{fleet}::
+* Add shipper label. {kib-pull}122491[#122491]
+* Add support for non-superuser access to *Fleet* and *Integrations*. {kib-pull}122347[#122347]
+* Add support for bundling packages as zip archives with {kib} source. {kib-pull}122297[#122297]
```
jen-huang, line 47: I think we ought to remove this one due to adding this patch: https://github.com/elastic/kibana/pull/126699. This means that we don't actually invoke the bundled packages functionality yet, so this release note could be misleading for users. WDYT @kpollich @joshdover?

nimarezainia (approved):

dedemorton:
Also something I missed earlier from https://github.com/elastic/observability-docs/issues/1489 that we need to highlight here. I might do this after merging this PR because I'm working on the related doc changes:

Highlight the change that Default policy will not be available by default, and has to be created on UI/API or preconfiguration.
Highlight the change that for Self-managed, Default Fleet Server policy will not be available by default, and has to be created on UI/API or preconfiguration.
Elastic Cloud agent policy is not changed, will continue to be provided by default.
Add a link to the new doc page created above for creating agent policies.

dedemorton:
I've created another issue to track the release notes work since that has a dependency on getting this PR merged: https://github.com/elastic/observability-docs/issues/1649

docs/en/ingest-management/release-notes/release-notes-8.1.asciidoc
dedemorton: This info is in the Kibana docs, too, which are already published. I think I will go ahead and merge this PR, and I can remove this item in a follow-up PR when I hear back. That way, I can remove it from Kibana, too. Just ping me if you want it removed. Thanks!

kpollich: @jen-huang I think this note can remain because all the code that supports installation from bundled packages still shipped here, even if we had to patch the actual bundled package assets out due to bugs. In 8.2 we can add a release note that we're bundling APM, Elastic Agent, Endpoint, Fleet Server, and Synthetics integration assets with the Kibana distributable.

PR #1644, opened 2022-03-07
Update whats-new.asciidoc

bmorelli25:
Update `main` what's new to only include branches from 7.17 on.


dedemorton:
Did you mean to leave these commented out?

bmorelli25:
Yes. They need to be updated when the What's New content is added.

Issue #1096, opened 2021-09-23
Renaming of index patterns to data views


EamonnTP:
## Description

Related issue: https://github.com/elastic/kibana/issues/100844

Looks like we now have shared attributes for **Data view** in the docs repo.

cc @elastic/obs-docs 

## Collaboration

* [x] The docs team will lead producing the content
* [ ] The product team will provide the initial content and the docs team will edit / review
* [ ] The docs team will define with product team the structure and location, and the product team will provide the initial content
* [ ] Other (please describe)

## Suggested Target Release

7.16

bmorelli25:
The Kibana index pattern documentation URL is also changing as a part of this renaming. All documentation links will need to be updated: `index-patterns.html` (old) → `data-views.html` (new). See https://github.com/elastic/kibana/pull/110421.

colleenmcginnis:
#1671 updates docs related to parts of the UI that have already been updated to use "data views". 

Not all parts of the Observability UI have switched from "index patterns" to "data views" yet. We should continue to track:

- [ ] https://github.com/elastic/kibana/issues/125498 for updates to Observability's **Metrics** > **Settings** > **Indices** > **Metric indices**
- [ ] https://github.com/elastic/kibana/issues/128192 for updates to Observability's **Uptime** > **Settings** > **Indices** > **Uptime indices**

bmorelli25:
The linked development issue was moved to the bottom of the backlog https://github.com/elastic/kibana/issues/125498#issuecomment-1609284915.

I'm keeping this in blocked and moving to the bottom of our backlog as well.

dedemorton:
https://github.com/elastic/kibana/issues/125498 was closed as "not planned." Should we just close this issue, too?

PR #1671, opened 2022-03-14
Replace "index patterns" with `data-source` shared attribute

colleenmcginnis:
Related https://github.com/elastic/observability-docs/issues/1096
Related https://github.com/elastic/kibana/issues/100844

The term "index patterns" (or `index-pattern` in URLs) appears on 10 pages. It looks like it's primarily in various **Settings** pages and after looking at the Kibana UI, it looks like some of these references have been changed to "data views" but others still use index patterns.

I'm trying to figure out if all references to "index patterns" should be changed or only where the UI text has changed. Here's what I found: 

### UI now using "data view":

**Logs** > **Settings** > **Data view** (new option)

Pages that mention this part of the UI: 
- [x] [Log monitoring » Configure data sources](https://www.elastic.co/guide/en/observability/current/configure-data-sources.html)
- [x] [Send data to Elasticsearch » Deploy Beats to send data » Ingest logs with Filebeat](https://www.elastic.co/guide/en/observability/current/ingest-logs.html)
- [x] [Send data to Elasticsearch » Deploy Beats to send data » Ingest metrics with Metricbeat](https://www.elastic.co/guide/en/observability/current/ingest-metrics.html)
- [x] [Centrally manage Elastic Agents in Fleet » Data streams](https://www.elastic.co/guide/en/fleet/current/data-streams.html)
    - Also line 42?

### UI using "index pattern":

**Metrics** > **Settings** > **Indices** > **Metric indices**

Pages that mention this part of the UI: 
- [ ] [Metrics monitoring » Configure settings](https://www.elastic.co/guide/en/observability/current/configure-settings.html)
- [ ] [Alerting » Create a metrics threshold rule](https://www.elastic.co/guide/en/observability/current/metrics-threshold-alert.html)
- [ ] [Alerting » Create an infrastructure threshold rule](https://www.elastic.co/guide/en/observability/current/infrastructure-threshold-alert.html)

**Uptime** > **Settings** > **Indices** > **Uptime indices**

Pages that mention this part of the UI: 
- [ ] [Synthetic monitoring via the Uptime app » Lightweight synthetic monitoring » Configure settings](https://www.elastic.co/guide/en/observability/current/configure-uptime-settings.html)

### Not sure: 

Pages that mention "index pattern" that I'm not sure if I should replace: 
- [x] [Alerting » Create a logs threshold rule](https://www.elastic.co/guide/en/observability/current/logs-threshold-alert.html)
- [x] [Package specification](https://www.elastic.co/guide/en/integrations-developer/current/package-spec.html)

### Question for @dedemorton:

Should I update references in the pages listed under "UI using index pattern" and "Not sure"?

### DIFF
--- a/docs/en/integrations/package-spec.asciidoc
+++ b/docs/en/integrations/package-spec.asciidoc
@@ -45,7 +45,7 @@ The following assets are typically found in an Elastic package:
 * {kib}
 ** Dashboards
 ** Visualization
-** {Data-sources}
+** {data-sources-cap}
 ** ML Modules
 ** Map
 ** Search
--- a/docs/en/integrations/package-spec.asciidoc
+++ b/docs/en/integrations/package-spec.asciidoc
@@ -45,7 +45,7 @@ The following assets are typically found in an Elastic package:
 * {kib}
 ** Dashboards
 ** Visualization
-** Index patterns
+** {Data-sources}
 ** ML Modules
 ** Map
 ** Search
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -39,7 +39,8 @@ and another dataset for disk I/O metrics with a field describing the number of b
 A user-configurable arbitrary grouping, such as an environment (`dev`, `prod`, or `qa`),
 a team, or a strategic business unit.
 A `namespace` can be up to 100 bytes in length (multibyte characters will count toward this limit faster).
-Using a namespace makes it easier to search the data from a given source by using index patterns, or to give users permissions to data by assigning an index pattern to user roles.
+Using a namespace makes it easier to search data from a given source by using a matching pattern.
+You can also use matching patterns to give users access to data when creating user roles.
 // Corresponds to the `data_stream.dataset` field.
 
 The naming scheme separates each components with a `-` character:
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -76,10 +76,10 @@ Under the hood though, things are a bit more complex.
 All of the juicy details are available in {ref}/data-streams.html[{es} Data streams].
 
 [discrete]
-[[data-streams-index-pattern]]
-== Index patterns
+[[data-streams-data-view]]
+== {data-sources-cap}
 
-When searching your data in {kib}, you can use an {kibana-ref}/index-patterns.html[index pattern]
+When searching your data in {kib}, you can use a {kibana-ref}/data-views.html[{data-source}]
 to search across all or some of your data streams.
 
 [discrete]
--- a/docs/en/observability/configure-logs-sources.asciidoc
+++ b/docs/en/observability/configure-logs-sources.asciidoc
@@ -28,13 +28,17 @@ default configuration settings.
 to read log data from.
 
 Each log source now integrates with {kib} index patterns which support creating and
-querying {kibana-ref}/managing-index-patterns.html[runtime fields]. You can continue
+querying {kibana-ref}/managing-data-views.html[runtime fields]. You can continue
 to use log sources configured to use an index name pattern, such as `filebeat-*`,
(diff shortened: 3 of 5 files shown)


bmorelli25:
> I'm trying to figure out if all references to "index patterns" should be changed or only where the UI text has changed.

All instances should be updated. It is curious that both the logs and metrics settings pages in Kibana still use "index pattern". AFAIK, that term shouldn't exist anywhere after 8.0.

dedemorton:
I think the term "index pattern" is still valid if you are talking about using a simple matching pattern on an index name (the old way). The UI still supports that:

image

I need to do a little testing today then I'll add some commments.

dedemorton:
Sorry it took me a little while to get back to you. So I agree with Brandon...in most cases, you do want to use data view instead of index pattern. Here are my suggestions:

**https://www.elastic.co/guide/en/fleet/current/data-streams.html**

I would change the following text to avoid saying index pattern.

OLD: "Using a namespace makes it easier to search the data from a given source by using index patterns..."

NEW: "Using a namespace makes it easier to search data from a given source by using a matching pattern..."

The second part of that sentence also needs work (it's confusing even if you fix "index pattern"):

"...or to give users permissions to data by assigning an index pattern to user roles." I'd suggest text to replace this, but it's late and I'm running on one brain cell.

Here's the UI that I think this topic is talking about. Here you are specifying a matching pattern (not a data view). Full disclosure, tho. I have not tested this to confirm that it works! Might be worth confirming with the dev team, then you can rewrite it to make more sense:
image

Also, in the following text, you DO want to use data views:

OLD:
Index patterns
When searching your data in Kibana, you can use an index pattern to search across all or some of your data streams.

NEW:
Data views
When searching your data in Kibana, you can use a data view to search across all or some of your data streams.

----

**https://www.elastic.co/guide/en/integrations-developer/current/package-spec.html**

Yes, change the bullet under Kibana to say Data views instead of Index patterns.

----

For all the other links, which I think say "index patterns" because it's still used in the UI, I would check with the dev team. I think it's better to match what the UI says for now and then update the docs when the UI gets updated. Might be that there's a mistake in the UI, and they've implemented changes inconsistently.

colleenmcginnis:
>For all the other links, which I think say "index patterns" because it's still used in the UI, I would check with the dev team.

Thanks @dedemorton! By "the dev team" do you mean the Kibana team (like the author of https://github.com/elastic/kibana/issues/100844)?

dedemorton:
>By "the dev team" do you mean the Kibana team (like the author of https://github.com/elastic/kibana/issues/100844)?

Yes exactly. I would start there and ask why the UI still uses index patterns.

colleenmcginnis:
From https://github.com/elastic/kibana/issues/109284#issuecomment-1071926485: 

>Should those references be updated? I'm not sure if I should update the docs referencing these pages or keep them as they are to match the UI.

>As best I can tell they're accurate as is - they're referring to index pattern strings rather than a kibana saved object.

So it sounds like the Metrics and Uptime settings pages should not be updated, right?

dedemorton:
> So it sounds like the Metrics and Uptime settings pages should not be updated, right?

That's right. But it does seem a little odd to me that the Logs UI gives users the option of using data views, but Metrics and Uptime do not. Maybe that's coming? Could be really confusing to users though. I'll add a response to Matt in that other PR.

dedemorton (approved):
LGTM!

docs/en/integrations/package-spec.asciidoc
```diff
@@ -45,7 +45,7 @@ The following assets are typically found in an Elastic package:
 * {kib}
 ** Dashboards
 ** Visualization
-** Index patterns
+** {Data-sources}
```
colleenmcginnis, line 48: This wasn't showing up as expected:

```suggestion
** {data-sources-cap}
```

image

colleenmcginnis:
Merging here, but confirmed that not all parts of the Observability UI have switched from "index patterns" to "data views" yet and documented issues to track going forward in https://github.com/elastic/observability-docs/issues/1096#issuecomment-1074243336.

PR #1680, opened 2022-03-16
Clarify that extra privileges are required to delete system indices

dedemorton:
Also removes a troubleshooting topic that is old and should have been removed in 7.15.


nchaulet (approved):
🚀 We know have an API to reset preconfigured API that is probably better to use than deleting from system indices, I can push a draft PR later with the right API call

dedemorton:
>We know have an API to reset preconfigured API that is probably better to use than deleting from system indices, I can push a draft PR later with the right API call

@nchaulet That would be great. Thanks!

PR #1687, opened 2022-03-17
[Fleet] Use reset API instead of system indices for reseting cloud policy

nchaulet:
## Description

document how to use the reset API available since 8.0 to reset the cloud policy instead of having the user manipulating system indices


dedemorton (commented):
Looks good! I love it when we make things easier.

I missed something that I should have updated earlier. Where we tell users to restart the APM& Fleet container, it should say:

```
. Restart the {integrations-server}:
+
In the {ecloud} console under {integrations-server}, click *Force Restart*.
```

Do you mind changing this since you have the PR open?

dedemorton (approved):
LGTM. Thank you!

PR #1720, opened 2022-03-24
Update Synthetics Integration with new UI options

### LINKED ISSUE #1411: Update Synthetics Integration with new UI options
## Description

The Synthetics Integration UI has two additional options added that provide extra functionality.  This will require updated screenshots in the existing documentation, as well as a little description about each option:
* Throttling options (from [kibana#114155](https://github.com/elastic/kibana/issues/114155))
  * We already document throttling in the [Heartbeat docs](https://www.elastic.co/guide/en/beats/heartbeat/master/monitor-browser-options.html#monitor-browser-throttling) but this has now also been exposed as a configuration in the Integration UI
* Add enabled option to Synthetics Agent Integration https://github.com/elastic/kibana/issues/119519
  * There is a slider to enable/disable the integration (so users don’t need to delete it if they just want to stop testing for a period of time)

## Collaboration

**Contact Person:**
* Throttling: @lucasfcosta 
* Enabled toggle: @dominiqueclarke 

## Suggested Target Release

8.0.0

## Stakeholders

* @paulb-elastic 
* @drewpost 
* @liciavale

colleenmcginnis:
Fixes #1411 

Updates screenshots and text to reflect changes to the Elastic Synthetics integration UI. Some notes and questions for @paulb-elastic @elastic/obs-docs:

* I split the screenshot into two images because it was getting very long with the addition of new fields, which made it more difficult to reference the image in the text (especially the new _Enabled_ toggle mentioned in #1411). 
* I added information on _Throttling options_ as requested in #1411, but there's no mention of other _Advanced Browser options_. Should those also be added?
* This doc outlines the fields for the Zip URL _Source Type_, but doesn't touch on the _Inline Script_ option. Should it?


paulb-elastic:
> I added information on Throttling options as requested in #1411, but there's no mention of other Advanced Browser options. Should those also be added?

Good point, I hadn’t actually realised they weren’t documented (and so through the throttling addition was going into existing docs).  It would be useful to include these - could you do this in a follow up issue/PR?

> This doc outlines the fields for the Zip URL Source Type, but doesn't touch on the Inline Script option. Should it?

Similarly, it probably is worth adding something for `Inline script` too, especially as this is actually the easiest way to get something up and running.  Could this also be done as part of a follow up issue/PR?

It’s worth mentioning that the integration will be deprecated in the next few releases (8.4 or 8.5), as we’re getting closer to Synthetics being GA, it’s all managed a different way.  However, these fields / capabilities will all still be in there, so we’ll be able to re-organise the layout, but the info will all still be relevant.

PR #1736, opened 2022-03-30
Add screenshots and text for inline script source type

### LINKED ISSUE #1727: [Request] Add info on Elastic Synthetics "Inline Script" option to quickstart
Follow up from #1720
May be impacted by #1725 

## Description

>>This doc outlines the fields for the Zip URL Source Type, but doesn't touch on the Inline Script option. Should it?
>
>Similarly, it probably is worth adding something for Inline script too, especially as this is actually the easiest way to get something up and running. Could this also be done as part of a follow up issue/PR?

## Collaboration

The docs team will lead producing the content

**Contact Person:** @paulb-elastic 

## Suggested Target Release

@paulb-elastic in which release was this introduced?

colleenmcginnis:
Fixes #1727 

Adds information on the "Inline script" source type to [Synthetic monitoring using Elastic Agent and Fleet](https://www.elastic.co/guide/en/observability/master/synthetics-quickstart-fleet.html).


docs/en/observability/synthetics-quickstart-fleet.asciidoc
```diff
@@ -50,16 +50,14 @@ You can use the _Enabled_ toggle to turn the monitor on or off in {fleet} withou
 [role="screenshot"]
 image::images/synthetics-integration.png[Synthetics integration]
 
-Set the _Source Type_ for the browser monitor.
-If you're using a <<synthetics-test-suite, synthetic test suite>>, use the _Zip URL_ type.
+Set the *Source Type* for the browser monitor.
```
paulb-elastic, line 53: I wonder if it would be worth making it clearer somehow, that the user is selecting **either** a `Zip URL` **or** `Inline`?

Maybe something like
> Set the **Source Type** for the browser monitor to either a **ZipURL** or an **Inline** script.

And then somehow nest the next two sections that document each?  At the moment, I am not sure it's clear as the Zip URL text then leads straight into the Inline text.

paulb-elastic (changes requested):
Left a comment about adding something to make it clearer that the user should select either a ZipURL **or** an Inline script

docs/en/observability/synthetics-quickstart-fleet.asciidoc
dedemorton, line 53: There are a couple of other places where italics are used instead of bold (can't mark them because they aren't in the diff). Might be a good idea to clean them up while you have the file open. Otherwise, looks great!

colleenmcginnis: >And then somehow nest the next two sections that document each? At the moment, I am not sure it's clear as the Zip URL text then leads straight into the Inline text.

Breaking this into sections would require going down to the `h7` level, which is _a lot_ of nesting (although I think it should be the fourth level for this page, but I might be understanding how books work). The fact that the CSS for a heading level this deep is essentially undistinguishable from the body text makes me thing we shouldn't create an `h7`. @dedemorton do we have guidelines on how many heading levels deep we go in docs? 

@paulb-elastic what do you think about the solution I pushed in https://github.com/elastic/observability-docs/pull/1736/commits/856fb5c5466e10258a355ff32d99171027ac2ac6? ([See preview here](https://observability-docs_1736.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-quickstart-fleet.html#synthetics-quickstart-fleet-configure-policy-source-type))

<details>
<summary>Click to see a screenshot that uses heading nesting</summary>

Screen Shot 2022-04-04 at 1 07 56 PM
Screen Shot 2022-04-04 at 1 08 16 PM

</details>

dedemorton: nevermind...I think you've covered those in another PR. :-)

dedemorton: h7 is too deep. You will end up with no formatting. The [docs readme](https://github.com/elastic/docs/blob/master/README.asciidoc#basic-book-structure) shows that we support up to level 4:  `===== Section title         // level 4`

But I'm pretty sure that the CSS now supports up to level 5, though IMO the formatting isn't distinct enough even at the level.

paulb-elastic: Thanks @colleenmcginnis, I understand your point about how many levels deep the nesting would go.  I think the updated copy makes it clearer (without the nesting) that the user is selecting one of the two options.

Thanks also for including the preview link directly, this makes it so much easier to get straight to change to see what it looks like when rendered.

LGTM

paulb-elastic (approved):
LGTM

PR #1741, opened 2022-03-31
add minimal agent install docs

DanRoscigno:
This is to support the workflow from Kibana > Add Integrations > Agent > Fleet Managed.  In the Kibana flyout there is a step that tells the reader to download the agent, and that goes to a download page that does not have enough information, or a download page that includes too much information (info about installing Fleet, etc.).

### DIFF
--- a/docs/en/observability/index.asciidoc
+++ b/docs/en/observability/index.asciidoc
@@ -46,6 +46,8 @@ include::ingest-traces.asciidoc[leveloffset=+3]
 
 include::ingest-splunk.asciidoc[leveloffset=+3]
 
+include::install-fleet-managed-agent.asciidoc[leveloffset=+3]
+
 include::deploy-beats-to-send-data.asciidoc[leveloffset=+2]
 
 include::ingest-logs.asciidoc[leveloffset=+3]
--- a/docs/en/observability/install-fleet-managed-agent.asciidoc
+++ b/docs/en/observability/install-fleet-managed-agent.asciidoc
@@ -1,0 +1,25 @@
+[[install-fleet-managed-agent]]
+= Install {fleet} managed {agent}
+
+This guide describes how to:
+
+* Install an {agent} that will be managed with {fleet}
+* Enroll the {agent} with {fleet}
+
+If you have come here from the {agent} download page, then you are in the right place.
+
+[discrete]
+[[minimal-add-agent-to-fleet]]
+== Step 1: Add an {agent} to {fleet}
+
+{agent} is a single, unified agent that you can deploy to hosts or containers to
+collect data and send it to the {stack}. Behind the scenes, {agent} runs the
+{beats} shippers or {elastic-endpoint} required for your configuration.
+
+include::{fleet-repo-dir}/elastic-agent/install-fleet-managed-elastic-agent.asciidoc[tag=install-elastic-agent]
+
+[[enroll-agent]]
+== Step 2: Enroll and start the {agent}
+
+Return to {kib} and follow the **Enroll and start the Elastic Agent** instructions.  The command provided in {kib} includes the information that {agent} needs to connect and authenticate with the {stack}.  When the {agent} enrolls it will be configured to send the data associated with the policy that you created.
+


dedemorton:
Let's talk more about this. I'm not sure the observability guide is the right place to send users. We have this topic that we could streamline: 

https://www.elastic.co/guide/en/fleet/current/install-fleet-managed-elastic-agent.html

But we need to decide what is essential info. When we gloss over prereqs, users tend to hit snags.

docs/en/ingest-management/elastic-agent/install-fleet-managed-agent.asciidoc
```diff
@@ -0,0 +1,29 @@
+["appendix",role="exclude",id="install-fleet-managed-agent"]
+= Install a {fleet} managed {agent}
```
dedemorton, line 2:
```suggestion
= Install a {fleet}-managed {agent}
```

docs/en/ingest-management/elastic-agent/install-fleet-managed-agent.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+This guide describes how to:
+
+* Install an {agent} that will be managed with {fleet}
+* Enroll the {agent} with {fleet}
```
dedemorton, line 7:
```suggestion
* Enroll the {agent} in {fleet}
```

dedemorton: Aligns with the wording in the UI

docs/en/ingest-management/elastic-agent/install-fleet-managed-agent.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+
+If you have come here from the {agent} download page, then you are in the right place.
+
+[discrete]
+[[minimal-add-agent-to-fleet]]
+== Step 1: Add an {agent} to {fleet}
+
+{agent} is a single, unified agent that you can deploy to hosts or containers to
```
dedemorton, line 15: This is OK for now, but wanted to let you know that we are going to update this description to encompass data protection.

The description in the overview is a starting point, but I think it needs to be more concise: https://www.elastic.co/guide/en/fleet/current/fleet-overview.html#elastic-agent

Just bringing this up in case you want to tackle it.... otherwise I can fix this when I make the change in other places.

docs/en/ingest-management/elastic-agent/install-fleet-managed-agent.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+Choose the operating system and run the provided commands to download and extract
+{agent} on your system.
+
+include::{fleet-repo-dir}/elastic-agent/install-fleet-managed-elastic-agent.asciidoc[tag=install-elastic-agent]
+
+[discrete]
+[[enroll-agent]]
+== Step 2: Enroll and start the {agent}
+
+Return to {kib} and follow the **Enroll and start the Elastic Agent** instructions.  The command provided in {kib} includes the information that {agent} needs to connect and authenticate with the {stack}.  When the {agent} enrolls it will be configured to send the data associated with the policy that you created.
```
dedemorton, line 28: What about users who go to the download page before they look at Kibana? We should provide something that makes sense for them. Maybe it's enough to mention the path to the Add Agent flyout?

dedemorton (commented):
A few minor comments.

docs/en/ingest-management/elastic-agent/install-fleet-managed-agent.asciidoc
DanRoscigno: Agreed, I looked at the product page for Agent to see what to say about protecting endpoints and did not find it.  As I build the endpoint security getting started guide I may work out wording.  I will also check the video from training.

DanRoscigno:
@dedemorton , this is updated, I think this addresses the problem of people coming to the page before adding an integration and creating a policy.

DanRoscigno:
@kilfoyle can you take a look please?  I think this will fit in with the workflow in your observability guide.  It is meant to be a hidden doc that would be linked to from the Download pages for Agent until the workflow moves completely into Kibana.

kilfoyle (approved):
LGTM! I'm not all that familiar with the Observability docs, but the text looks good to me and this would be nice to include in the Getting Started Observability guide.

dedemorton (approved):
LGTM! Thanks for adding this.

DanRoscigno:
@elasticmachine test this please

PR #1758, opened 2022-04-05
[Fleet] Do not recommand user to run Fleet server in http

nchaulet:
## Description 

Related to https://github.com/elastic/kibana/pull/129371

We should not recommend user to run Fleet server in http and should use the default self generated certificate instead.


dedemorton:
I think you want to remove the whole example plus the lead in sentence, "For example, the following command installs a Fleet Server and uses self-signed certs:

image

nchaulet:
@dedemorton I think we want to keep that example the text is correct as it's generating self signed certificate, the option `--fleet-server-insecure-http` was running fleet server in http mode without generating self signed certificate this was incorrect

dedemorton:
Ohhhhh. TBH, I have to re-learn this stuff every time I look at it. :-)

dedemorton (approved):
LGTM

dedemorton:
I didn't remember having this flag in the command originally, so I looked at the history, and it looks like it was added back in January. Should we try to figure out why, or just say your revision is the correct one? https://github.com/elastic/observability-docs/commit/b85317d9c615125b6a5fd4312f8619d07ba98c87

PR #1725, opened 2022-03-29
Update the structure of Synthetic monitoring section

colleenmcginnis:
⚠️ **This is a work in progress.** Feedback on the approach is welcome any time! Close review of copy/details is not needed _yet_.

Restructures the current [Synthetic monitoring section]() of the Observability guide to better connect the concepts of "lightweight" and "browser" checks and align with the language in the UI.

## Approach

| Before | After |
|---|---|
[...]

PR #1783, opened 2022-04-12
Remove locally defined `release-state` variable

colleenmcginnis:
Ref https://github.com/elastic/observability-docs/pull/1725#discussion_r848863081

From @bmorelli25 on the inclusion of `:release-state: released` and `:!release-state:` in [`/docs/en/observability/ingest-logs.asciidoc`](https://raw.githubusercontent.com/elastic/observability-docs/main/docs/en/observability/ingest-logs.asciidoc):

>That's a mistake and should be removed. That was added to overwrite the `:release-state:` attribute in the docs repo (https://github.com/elastic/docs/blob/master/shared/versions/stack/master.asciidoc#L21) which is always set to `unreleased` in `main` and `master`. Adding the overwrite at the beginning of the file (and then unsetting it at the end) allows the page contents to render for review purposes.

Not sure about backporting on a PR like this one!

### DIFF
--- a/docs/en/observability/ingest-logs.asciidoc
+++ b/docs/en/observability/ingest-logs.asciidoc
@@ -7,7 +7,6 @@
 :beatname_url: {beats-ref-root}/{beatname_lc}/{branch}
 :modulename: nginx
 :has_modules_command:
-:release-state: released
 
 [[ingest-logs]]
 = Ingest logs with {filebeat}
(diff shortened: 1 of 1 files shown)


dedemorton (approved):
LGTM. I would not backport unless you see this in a branch that has not been released yet (like 8.2).

Issue #1310, opened 2021-12-03
[Observability RAC] document xpack.ruleRegistry.write.disabledRegistrationContexts flag 


mgiota:
## Description

In this [issue](https://github.com/elastic/kibana/issues/119217) we introduced a new `xpack.ruleRegistry.write.disabledRegistrationContexts` flag, which user can use to disable writing to alerts-as-data indices in a more granular way. 

## Collaboration

- The docs team will define with product team the structure and location, and the product team will provide the initial content

`xpack.ruleRegistry.write.disabledRegistrationContexts`::
Specifies the observability alert indices that are disabled to be written. Data type is array. Allowed values are: [ 'observability.logs','observability.metrics','observability.apm','observability.uptime' ]

**Contact Person:**
@mgiota 

## Suggested Target Release

8.0.0

colleenmcginnis:
@mgiota should information on this new flag replace the content at the beginning of [View Alerts](https://www.elastic.co/guide/en/observability/current/view-observability-alerts.html)? 

<img width="787" alt="image" src="https://user-images.githubusercontent.com/10479155/161339677-04a781ee-b888-401c-b5fe-f38b1cf8a470.png">

colleenmcginnis:
@vinaychandrasekhar can you help with the question above?

vinaychandrasekhar:
@mgiota would be the right person to answer the query. @mgiota thoughts please when you get a chance?

PR #1789, opened 2022-04-13
Add content on disabling writing to Observability alert indices

colleenmcginnis:
Fixes https://github.com/elastic/observability-docs/issues/1310

Adds information on how to disable writing to specific Observability alert indices using Kibana settings. 

@elastic/obs-docs @mgiota what do you think about this approach? ([Preview the page](https://observability-docs_1789.docs-preview.app.elstc.co/guide/en/observability/master/create-alerts.html#create-alerts-configure.)) Alternatively, I could move this content to its own page or integrate elsewhere (let me know if you have any suggestions)!

cc @vinaychandrasekhar


bmorelli25 (commented):
This approach looks good to me.

docs/en/observability/create-alerts.asciidoc
```diff
@@ -69,6 +69,66 @@ list an alert on the {observability} Alerts page.
...
+[discrete]
+[[create-alerts-remove-page]]
+=== Remove the Alerts page
+
+To disable writing to all alert indices and remove the Alerts page from {kib} altogether, use the following settings:
+
+[source,yml]
+----
+xpack.ruleRegistry.write.enabled: 'false'
+xpack.observability.unsafe.cases.enabled: 'false'
```
mgiota, line 127: @colleenmcginnis We can get your current PR merged, but I put a note to investigate further why we have `xpack.observability.unsafe.cases.enabled: false` as a requirement to disable Alerts page. I don't think it is needed. `xpack.observability.unsafe.cases.enabled` is used to hide or not the `Cases` page. @vinaychandrasekhar I would like your thoughts on that.

docs/en/observability/create-alerts.asciidoc
```diff
@@ -69,6 +69,66 @@ list an alert on the {observability} Alerts page.
...
+while continuing to write to others, use `xpack.ruleRegistry.write.disabledRegistrationContexts`.
+
+You can disable writing to alert indices for:
+
+* Logs (`observability.logs`)
+* Metrics (`observability.metrics`)
+* APM (`observability.apm`)
+* Uptime (`observability.uptime`)
+
+NOTE: Disabling writing to uptime alert indices will affect _all_ uptime rule types including monitor status and TLS rule types.
```
mgiota, line 102: @colleenmcginnis This is definitely correct and you can keep it as it is! I am just wondering if it is clear to the user that same applies to rest observability rule types as well. Maybe we could have a more generic sentence that follows with a more specific example as you have here. For example `Disabling writing to the indices of an observability solution, will affect all rule types of the corresponding solution. For example disabling writing to uptime alert indices will affect _all_ uptime rule types including monitor status and TLS rule types.`

This is just a suggestion. Feel free to follow it or not.

mgiota (approved):
@colleenmcginnis The approach you used LGTM! I left a few comments, that shouldn't be blocking you from merging this PR.

docs/en/observability/create-alerts.asciidoc
vinaychandrasekhar: I agree that alerts can be disabled without disabling cases (although doing that would be a corner case). cc @cnasikas as FYI and for any further feedback.

cnasikas: Hi @mgiota and @vinaychandrasekhar! Indeed Cases are GA and not connected directly to the alerts. However, you can attach an alert to a case and view the alert's details flyout from within the Case. So if Cases are enabled and the alerts are disabled you cannot attach an alert to a case. If Cases are disabled and the alerts are enabled then you have to hide the "attach to a case" functionality from the alerts table. Another scenario that could occur (very unlikely tbh) is to enable both, attach an alert to a case, and then disable the alerts. The case will have an alert attached to a case but I am not sure if it can view the alert's details flyout because it is provided by the o11y.

vinaychandrasekhar: Thanks @cnasikas for the description. Based on that, I'd suggest still treating the two flags separately like @mgiota said at the start of this comment thread. Users can disable just alerts.

PR #1824, opened 2022-04-28
Updates docs about adding agents

dedemorton:
Closes #1792 

I've also updated some old content.

Preview links:

- [Ingest logs, metrics, and uptime data with Elastic Agent](https://observability-docs_1824.docs-preview.app.elstc.co/guide/en/observability/master/ingest-logs-metrics-uptime.html)
- [Install Elastic Agents](https://observability-docs_1824.docs-preview.app.elstc.co/guide/en/fleet/master/elastic-agent-installation.html)
- [Install Fleet-managed Elastic Agents](https://observability-docs_1824.docs-preview.app.elstc.co/guide/en/fleet/master/install-fleet-managed-elastic-agent.html)

Remaining TODOs:

- [ ] Make similar changes to the topic about [ingesting traces](https://www.elastic.co/guide/en/observability/current/ingest-traces.html) -- Open a new PR.
- [x] Delete screen shots that are no longer used.

PR #1833, opened 2022-04-29
Remove old images

dedemorton:
Removes old files that were renamed and replaced in https://github.com/elastic/observability-docs/pull/1824


bmorelli25 (approved):
👍 files match up with previous PR

PR #1854, opened 2022-05-05
Rename remaining APM & Fleet Server references

dedemorton:
Fixes some instances of APM  & Fleet Server that we missed previously.

Also removes an old section for upgrading from 7.12 to 7.13 that's no longer needed in 8.0+ (and is actually confusing because some of the info was inadvertently updated when it shouldn't have been).


dedemorton:
hmm...I'll need to figure out why Kibana still has a doc link to the content I removed.

jen-huang (approved):

dedemorton:

dedemorton:
Note that I'm not backporting this change to 8.0 or 8.1 because the doc link service changes won't be backported to those versions.

PR #1858, opened 2022-05-06
Move note about self-hosted fleet server 1/2

TheRiffRafi:
I am proposing to move this note to the section before the tab-widget. The reason is that it is buried in the widget and one can miss that this is an option when trying to install extra fleet servers with an ESS deployment (Happened to me and to a customer when trying to figure out how to do this). 
See the move in the my next PR.


dedemorton:
Good point, but it looks like this PR deletes the note without adding it to a new location.

Also it's better if you  make changes against the main branch.

dedemorton:
Just realized that you probably clicked the Edit button to create this. Rather than creating two separate PRs, can you put all the changes in one, or let me know if you want me to make the changes. thanks!

TheRiffRafi:
@dedemorton you're fast!! Got to this before I noted the other one with the move! Ah, let me create a single PR then.

PR #1859, opened 2022-05-06
Move note about self-hosted fleet server 2/2

TheRiffRafi:
I am proposing to move this note to the section before the tab-widget. The reason is that it is buried in the widget and one can miss that this is an option when trying to install extra fleet servers with an ESS deployment (Happened to me and to a customer when trying to figure out how to do this).
See the deletion in PR https://github.com/elastic/observability-docs/pull/1858


dedemorton:
As I mention in the related PR, it's better to have one PR with these changes so the commit history is clear on what's happening here.

PR #1860, opened 2022-05-06
Move note about self-hosted fleet server

TheRiffRafi:
I am proposing to move this note, which is listed in [this tab-widget](https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc), to the section before the tab-widget. The reason is that it is buried in the widget and one can miss that this is an option when trying to install extra fleet servers with an ESS deployment (Happened to me and to a customer when trying to figure out how to do this).
I added an extra text to the note to make it clear that we can have extra self-managed fleet server in addition to our hosted one.


TheRiffRafi:
Yeah, we should remove it, I wasn't sure on how to do that on this pull requests as well.

dedemorton (approved):
LGTM!

PR #1944, opened 2022-06-23
[Synthetics] Document changing retention on a per-data-stream basis

andrewvc:
Fixes https://github.com/elastic/synthetics/issues/286

This patch adds a new docs page for the commonly requested task of altering retention periods for the space-hungry
`synthetics-browser.network` and `synthetics-browser.screenshot` indices.

While we will have better defaults in https://github.com/elastic/uptime/issues/462 we need this doc before that change makes it into 8.4. Additionally, we will always need some level of documentation here, though we can simplify it a little after that.

I've tried to keep the docs screenshot free for kibana index management, since I don't think we'll be able to track visual changes there well.

It _is_ a complex process, but I've tried my best to describe it.

Once the better defaults are merged it should be as simple as clicking on the default lifecycle policy and editing it, a lot of the complexity now is due to the fact that we share one policy across all dataset types.

### DIFF
--- a/docs/en/observability/manage-synthetics-retention.asciidoc
+++ b/docs/en/observability/manage-synthetics-retention.asciidoc
@@ -11,7 +11,7 @@ to control storage utilization. This usually means altering the retention of bro
 All types of checks record 'core' metadata, such as which URL was checked, what the status of the check was, and errors that occurred. This document 
 focuses on browser checks, which tend to use much more storage than lightweight checks. 
 Browser based checks store two additional types of data beyond the core metadata; `network` documents and `screenshot` documents.
-These indices are usually many times larger than the core metadata. Nhe relative sizes of each vary depending on the sites being 
+These indices are usually many times larger than the core metadata. The relative sizes of each vary depending on the sites being 
 checked with network data usually being the larger of the two by a significant factor.
 
 `network` documents data consists of detailed metadata around requests for resources required by the pages being checked. An example would be an image, 
@@ -38,4 +38,4 @@ those containing the term 'synthetics'. In the UI there will be three sorts of d
 To change the retention period of a given data stream simply edit the life cycle policy associated with it after ensuring that policy does not apply to addition data streams whose
 retention you do not want to change. If the data stream you wish to change shares a life cycle policy with another, [create a new ILM policy](https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-lifecycle-policy.html), then edit the relevant component template's
 index settings. When editing the component template change the `index.lifecycle.name` value to point toward your new ILM policy.  For `screenshot` documents edit the `synthetics-browser.screenshot@package` component template, for `network_info` documents edit the `synthetics-browser.network@package` template. Check that the settings have taken effect by visiting the data stream management
-page and confirming the the attached life cycle policy is correct.
+page and confirming the attached life cycle policy is correct.
--- a/docs/en/observability/manage-synthetics-retention.asciidoc
+++ b/docs/en/observability/manage-synthetics-retention.asciidoc
@@ -1,41 +1,62 @@
 [[manage-synthetics-retention]]
 
-= Managing data retention
+= Manage data retention
+
+When you set up a synthetic monitor, data from the monitor is saved in
+https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams.html[{es} data streams],
+an append-only structure in {es}.
+
+There are six data streams recorded by synthetic monitors: `http`, `tcp`, `icmp`, `browser`, `browser.network`, `browser.screenshot`.
+Elastic will retain data from each data stream for some time period,
+and the default time period varies by data stream.
+If you want to reduce the amount of storage required or store data for longer,
+you can customize how long to retain data for each data stream.
 
 [discrete]
-== Overview
+== Synthetics data streams
 
-Synthetics browser monitors can require large amounts of storage, this document provides information on how synthetics stores data and how to optimize retention
-to control storage utilization. This usually means altering the retention of browser `network` and `screenshot` documents.
+There are six data streams recorded by synthetic monitors:
 
-All types of checks record 'core' metadata, such as which URL was checked, what the status of the check was, and errors that occurred. This document 
-focuses on browser checks, which tend to use much more storage than lightweight checks. 
-Browser based checks store two additional types of data beyond the core metadata; `network` documents and `screenshot` documents.
-These indices are usually many times larger than the core metadata. The relative sizes of each vary depending on the sites being 
+[options="header"]
+|===
+| Data stream | Data includes | Default retention period |
+| `http` | The URL that was checked, the status of the check, and any errors that occurred | 1 year |
+| `tcp` | The URL that was checked, the status of the check, and any errors that occurred | 1 year |
+| `icmp` | The URL that was checked, the status of the check, and any errors that occurred | 1 year |
+| `browser` | The URL that was checked, the status of the check, and any errors that occurred | 1 year |
+| `browser.screenshot` | Binary image data used to construct a screenshot and metadata with information related to de-duplicating this data | 14 days |
+| `browser.network` | Detailed metadata around requests for resources required by the pages being checked | 14 days |
+|===
+
+// preserving the text below in case we want to expand on the definitions above
+
+// `network` documents data consists of detailed metadata around requests for resources required by the pages being checked. An example would be an image, 
+// a javascript file, or a font referenced and loaded by the page under test. 
+// For each of these requests the URL, timing, headers, and other metadata are stored. While individually this metadata is small, 
+// modern websites often request hundreds of additional resources per page load, adding up to a significant amount of storage.
+
+// Screenshot data consists of binary image data used to construct a screenshot and metadata with information related to de-duplicating this data. De-duplication 
+// efficiency makes screenshot data less burdensome than network data, especially for sites that are mostly visually unchanged across test runs. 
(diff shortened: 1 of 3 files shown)


docs/en/observability/manage-synthetics-retention.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+[discrete]
+== Overview
+
+Synthetics browser monitors can require large amounts of storage, this document provides information on how synthetics stores data and how to optimize retention
+to control storage utilization. This usually means altering the retention of browser `network` and `screenshot` documents.
+
+All types of checks record 'core' metadata, such as which URL was checked, what the status of the check was, and errors that occurred. This document 
+focuses on browser checks, which tend to use much more storage than lightweight checks. 
+Browser based checks store two additional types of data beyond the core metadata; `network` documents and `screenshot` documents.
+These indices are usually many times larger than the core metadata. Nhe relative sizes of each vary depending on the sites being
```
emilioalvap, line 14: nit: typo at `Nhe`

docs/en/observability/manage-synthetics-retention.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+structure in Elasticsearch. Synthetics data streams can be managed via the Elasticsearch API or via [Kibana index management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-mgmt.html). 
+
+If Synthetics browser data streams are storing data longer than necessary, users can opt to retain `screenshot` and `network_info` datasets for a shorter period than the core metadata.
+To do so, first navigate to [Kibana index management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-mgmt.html), then filter the list of data streams for
+those containing the term 'synthetics'. In the UI there will be three sorts of data stream present `synthetics-browser-*`, `synthetics-browser.network-*`, and `synthetics-browser.screenshot-*`. From this page you can retrieve the size of each data stream on disk, as well as which [index life cycle](https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-lifecycle-policy.html) is associated with it. 
+
+To change the retention period of a given data stream simply edit the life cycle policy associated with it after ensuring that policy does not apply to addition data streams whose
```
emilioalvap, line 38: > apply to addition data streams whose ...

@andrewvc  Should this be `additional` instead?

docs/en/observability/manage-synthetics-retention.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+retention you do not want to change. If the data stream you wish to change shares a life cycle policy with another, [create a new ILM policy](https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-lifecycle-policy.html), then edit the relevant component template's
+index settings. When editing the component template change the `index.lifecycle.name` value to point toward your new ILM policy.  For `screenshot` documents edit the `synthetics-browser.screenshot@package` component template, for `network_info` documents edit the `synthetics-browser.network@package` template. Check that the settings have taken effect by visiting the data stream management
+page and confirming the the attached life cycle policy is correct.
```
emilioalvap, line 41: nit: the the -> the

docs/en/observability/manage-synthetics-retention.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+
+Synthetics data is recorded in Elasticsearch [data streams](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams.html), an append-only
+structure in Elasticsearch. Synthetics data streams can be managed via the Elasticsearch API or via [Kibana index management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-mgmt.html). 
+
+If Synthetics browser data streams are storing data longer than necessary, users can opt to retain `screenshot` and `network_info` datasets for a shorter period than the core metadata.
+To do so, first navigate to [Kibana index management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-mgmt.html), then filter the list of data streams for
+those containing the term 'synthetics'. In the UI there will be three sorts of data stream present `synthetics-browser-*`, `synthetics-browser.network-*`, and `synthetics-browser.screenshot-*`. From this page you can retrieve the size of each data stream on disk, as well as which [index life cycle](https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-lifecycle-policy.html) is associated with it.
```
paulb-elastic, line 36: nit - I know we're trying to avoid UI specific screenshots etc., but might be worth adding that the user needs to click on (or at least, select) the data stream to see the size and Index lifecycle policy, as it's not shown on the main _Data Streams_ table, e.g. adding something like:
> From this page, select a data stream to retrieve the size on disk...

docs/en/observability/manage-synthetics-retention.asciidoc
paulb-elastic: Without screenshots, I think this is quite difficult to follow.  Also, it contains details on how to change the retention, but also nests info about how to create a new one (if it’s shared by something else where you don’t want to change the retention).

It may be easier to follow if these were split out?

andrewvc: @colleenmcginnis @ollyhowell I intentionally omitted screenshots because they're hard to maintain... but I fear you're right. It _is_ very hard to follow without them. Does anyone on the docs team have thoughts there?

I'm inclined to just include them. Perhaps it's better to have them and let them get a little out of date if need be. Thoughts?

docs/en/observability/manage-synthetics-retention.asciidoc
andrewvc: I agree, this is another casualty of me trying to be agnostic to changes in the UI and clarity suffering as a result :( . @colleenmcginnis @ollyhowell thoughts on this? How can we write maintainable docs here.

docs/en/observability/manage-synthetics-retention.asciidoc
colleenmcginnis: It's ok to use screenshots when they are necessary! The fact that you've gone through this thought process before deciding they are necessary shows that you're not at risk of _overusing_ them. I think that's what the docs team is usually looking out for -- _unnecessary_ screenshots. 🙂

docs/en/observability/manage-synthetics-retention.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+
+To change the retention period of a given data stream simply edit the life cycle policy associated with it after ensuring that policy does not apply to addition data streams whose
+retention you do not want to change. If the data stream you wish to change shares a life cycle policy with another, [create a new ILM policy](https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-lifecycle-policy.html), then edit the relevant component template's
+index settings. When editing the component template change the `index.lifecycle.name` value to point toward your new ILM policy.  For `screenshot` documents edit the `synthetics-browser.screenshot@package` component template, for `network_info` documents edit the `synthetics-browser.network@package` template. Check that the settings have taken effect by visiting the data stream management
```
dominiqueclarke, line 40: Some other documentation includes a bit more specifics about how to edit the component template, exactly what to change with example configuration. I think we should include that here.

docs/en/observability/manage-synthetics-retention.asciidoc
```diff
@@ -0,0 +1,41 @@
...
+structure in Elasticsearch. Synthetics data streams can be managed via the Elasticsearch API or via [Kibana index management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-mgmt.html). 
+
+If Synthetics browser data streams are storing data longer than necessary, users can opt to retain `screenshot` and `network_info` datasets for a shorter period than the core metadata.
+To do so, first navigate to [Kibana index management](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-mgmt.html), then filter the list of data streams for
+those containing the term 'synthetics'. In the UI there will be three sorts of data stream present `synthetics-browser-*`, `synthetics-browser.network-*`, and `synthetics-browser.screenshot-*`. From this page you can retrieve the size of each data stream on disk, as well as which [index life cycle](https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-lifecycle-policy.html) is associated with it. 
+
+To change the retention period of a given data stream simply edit the life cycle policy associated with it after ensuring that policy does not apply to addition data streams whose
```
dominiqueclarke, line 38: > simply edit the life cycle policy associated with it
Since our ILM policy defaults will be based off the integration package, we have to be careful with this. If the user edits existing ILM managed by the policy, their changes will be overwritten on the next update to the integration policy.

dominiqueclarke:
> Once the better defaults are merged it should be as simple as clicking on the default lifecycle policy and editing it, a lot of the complexity now is due to the fact that we share one policy across all dataset types.

That's unfortunately not accurate. ILM policies owned by integrations are managed by those integrations, and will update to overwrite any changes to those default settings when the integration package itself is updated. 

The Fleet team documented how to apply a custom ILM policy to your data streams in a way that preserves changes here: https://github.com/elastic/observability-docs/pull/1194/files

colleenmcginnis (commented):
@andrewvc I left some comments and suggestions below on the structure of this content. I'll come back and copy edit after those comments are resolved (to avoid editing things that might change).

This guide does a good job of describing why the problem happens and what to do to fix it. I think it's light on _how a user would know this is the problem_. When I'm writing troubleshooting content I usually try to guess what keywords the user is likely going to type into Google, _especially_ related error messages (since my first troubleshooting step is almost always to copy and paste the error into Google 🙂).

Is there anything else we can add to make sure users using search terms that describe what they're seeing in a moment of frustration end up on this page (to get the answers they need)?

colleenmcginnis:
👋  @andrewvc what can I do to help here?

colleenmcginnis (approved):

PR #1964, opened 2022-06-29
update Beats capability for DStreams

Hythloday-zero:
As of Beats 8.0, data streams are the default output to Elasticsearch.  Line 245 needs change to {y} and the phrasing in Line 248 should reflect that all three options can take advantage of data streams, easier ILM, and the data stream naming scheme.

Reference: https://github.com/elastic/beats/pull/28450 linked in https://www.elastic.co/guide/en/beats/libbeat/8.0/release-notes-8.0.0.html

### DIFF
--- a/docs/en/ingest-management/beats-agent-comparison.asciidoc
+++ b/docs/en/ingest-management/beats-agent-comparison.asciidoc
@@ -242,10 +242,10 @@ The following table shows a comparison of capabilities supported by {beats} and 
 |{fleet} can automatically give {agent}s minimal output permissions based on the inputs running. With standalone {agent} and {beats}, users often give overly broad permissions because it's more convenient.
 
 |Data streams support
-|{n}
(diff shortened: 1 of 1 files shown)


Hythloday-zero:
> Good catch! LGTM. This PR will need to be forward ported to main, and should be backported to 8.0-8.2. Do you want me to take care of that for you?

Yes, thank you, that would be great- I wasn't sure how the process worked; I'm over on the Consulting side in AMER-FED and happened across this doing some research for a client.

dedemorton:
>I wasn't sure how the process worked

This is great, but next time, select master from the version selector before you click Edit. Just makes it a bit easier to backport. No worries, though. It's good to have the content change.

Hythloday-zero:
> but next time, select master from the version selector before you click Edit.

Understood.  ezpz

PR #1971, opened 2022-06-30
Document some Fleet Server timeout configs

joshdover:
These may be useful for our scaling efforts, thought making them public made sense as well. Would like @ph to confirm before we publish these.

### DIFF
--- a/docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
+++ b/docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
@@ -61,14 +61,15 @@ Total size of the cache.
 
 `server.timeouts`::
 `checkin_timestamp`:::
-How often Fleet Server updates the "last activity" field for each agent.
-Defaults to `30s`. When deploying higher scales of agents, increasing this
-setting may improve performance. If set higher than `2m`, most agents will be
-shown as "offline" in the Fleet UI and is not recommended in a typical set up.
+How often {fleet-server} updates the "last activity" field for each agent.
+Defaults to `30s`. In a large-scale deployment, increasing this
+setting may improve performance. If this setting is higher than `2m`,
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
```diff
@@ -59,6 +59,18 @@ connections.
 `max_cost`:::
 Total size of the cache.
 
+`server.timeouts`::
+`checkin_timestamp`:::
+How often Fleet Server updates the "last activity" field for each agent.
+Defaults to `30s`. When deploying higher scales of agents, increasing this
+setting may improve performance. If set higher than `2m`, most agents will be
```
joshdover, line 66: In 8.4.0 we're changing this UI timeout to 10m instead of 2m. We may want to make the UI configurable too though

docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
```diff
@@ -59,6 +59,18 @@ connections.
...
+shown as "offline" in the Fleet UI and is not recommended in a typical set up.
```
dedemorton, line 67:
```suggestion
How often {fleet-server} updates the "last activity" field for each agent.
Defaults to `30s`. In a large-scale deployment, increasing this
setting may improve performance. If this setting is higher than `2m`,
most agents will be shown as "offline" in the Fleet UI. For a typical setup,
it's recommended that you set this value to less than `2m`.
```

dedemorton (commented):
Thanks for your contribution. Just a couple editorial changes.

docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
joshdover: Thank you so much, makes sense to me :)

PR #1976, opened 2022-07-05
Fix wrong config in elastic-agent-debug-input-configs.asciidoc

ChrsMark:
Replace `{{var}}` with `${var}`.


dedemorton (approved):
LGTM

Want to note that I did not test this change but it looks right.

dedemorton:
I've added some backport labels. Feel free to adjust. I think we should backport to 7.17 at least, especially if the auto backports work. :-)

PR #1987, opened 2022-07-08
Indicate when there are no bug fixes for Agent or Fleet

dedemorton:
Let's be explicit in the release notes when there are no bug fixes in patch releases.


bmorelli25 (approved):
Makes sense

PR #2007, opened 2022-07-20
[synthetics] Move Heartbeat tab to first position in all synthetics docs

colleenmcginnis:
As discussed with @andrewvc @lucasfcosta, we should default to Heartbeat as the primary method of creating and managing monitors while other methods are still in beta.

### DIFF
--- a/docs/en/observability/synthetics-run-test.asciidoc
+++ b/docs/en/observability/synthetics-run-test.asciidoc
@@ -10,6 +10,18 @@ Once you've <<synthetics-create-test, written a synthetic test>>, you can create
 There are three ways to create a browser monitor: project monitors, {heartbeat}, and {agent}.
 
 [discrete]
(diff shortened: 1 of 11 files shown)


lucasfcosta (commented):
Thanks for tackling this!

The overall text LGTM but when I was look at the docs preview it seemed like the text within the tabs was incorrect until I selected another tab and moved back (as shown in the video below).

https://user-images.githubusercontent.com/6868147/180165106-b0f42b0e-8549-43f5-8677-1403936376df.mov

colleenmcginnis:
Thanks @lucasfcosta! I missed a step initially, but added it in https://github.com/elastic/observability-docs/pull/2007/commits/a0a25c3f6bb957f8633d7f167bd864ccbf608b5a.

andrewvc (approved):
LGTM!

PR #2042, opened 2022-07-29
misuse of word

cs-emiryildiz:
For logging option it also said deselect.


dedemorton:
Thanks for your contribution! 

If you read the intro paragraph, the topic is telling users how to turn logging off, which I agree is confusing. I can see why you thought it was a mistake.

I can take over the changes so you don't have to sign the CLA or do more work. I've created https://github.com/elastic/observability-docs/issues/2050 to track this work. Thanks again!

PR #2077, opened 2022-08-10
[DOCS] Add notes about security for ML anomaly detection job

lcawl:
Relates to https://github.com/elastic/stack-docs/pull/2204

This PR adds machine learning requirements to [Inspect log anomalies](https://www.elastic.co/guide/en/observability/master/inspect-log-anomalies.html) and [Categorize log entries](https://www.elastic.co/guide/en/observability/master/categorize-logs.html) in the Observability Guide. It also moves a link from the top to the bottom of the latter page, since the target provides more advanced options than are initially required. 

The requirements text is partially borrowed from the UI text here:

image


droberts195 (approved):
LGTM

PR #2078, opened 2022-08-11
Add docs about setting agent binary download location

dedemorton:
Closes #2006 

Main preview links:

https://observability-docs_2078.docs-preview.app.elstc.co/guide/en/fleet/master/air-gapped.html
https://observability-docs_2078.docs-preview.app.elstc.co/guide/en/fleet/master/fleet-settings.html

There are changes to other pages that you can just review in the diff.

Other changes:
- Moves the topic about air gapped environments to the installation section because it now contains info about download artifacts.
- Adds an updated version of the reference.yml (8.4.0 version).

PR #2088, opened 2022-08-12
Add hint about standalone agents for upgrade

ckauf:
Following the[ command reference](https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-upgrade-command
) the upgrade with --source-uri is only supported for standalone agents:
> Upgrade the currently running Elastic Agent to the specified version. This should only be used with agents running in standalone mode. Agents enrolled in Fleet should be upgraded through Fleet.

This information is missing here and will lead to frustration on the user site because the upgrade will fail with an error message when trying to upgrade fleet managed agents. Adding this information here will make clear for what use case the command is intended.

### DIFF
--- a/docs/en/ingest-management/fleet/air-gapped.asciidoc
+++ b/docs/en/ingest-management/fleet/air-gapped.asciidoc
@@ -141,7 +141,7 @@ NODE_EXTRA_CA_CERTS="/etc/kibana/certs/ca-cert.pem"
 [[air-gapped-limitations]]
 == Limitations for {agent} upgrades
 
-The <<upgrade-elastic-agent>> feature in technical preview does not currently support network restricted environments without access to artifacts.elastic.co.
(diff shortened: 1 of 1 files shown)


ckauf:
Maybe the additional hint about fleet managed agents is not 100% accurate. To my knowledge there is no way up until 8.4 to upgrade an agent without access to artifacts.elastic.co. Can we clarify that with the agent team?

dedemorton:
@michalpristas Can you respond to the question that @ckauf asks here:
> To my knowledge there is no way up until 8.4 to upgrade an agent without access to artifacts.elastic.co. Can we clarify that with the agent team?

@ckauf Starting in 8.4, we've added a setting to Fleet that enables users to set the source URI from the UI. This whole section is being removed in https://github.com/elastic/observability-docs/pull/2078 (not reviewed yet, but will be merged for 8.4).

PR #2093, opened 2022-08-17
Update commands.asciidoc

nimarezainia:
the "non-interactive" flag was never documented.

### DIFF
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -288,9 +288,9 @@ Note: if the {agent} is already installed on the host, using --force may result 
 unpredictable behaviour with duplicate {agent}s appearing in {fleet}
 
 `--non-interactive`::
-The {agent} will be installed in a non-interactive mode. This flag is helpful when 
-using automation software or scripted deployments and will check if there is an {agent}
-already installed. If there is an {agent} on the host already, installation will abort.
+Install {agent} in a non-interactive mode. This flag is helpful when 
+using automation software or scripted deployments. If {agent} is
+already installed on the host, the installation will terminate.
 
 `--help`::
 Show help for the `enroll` command.
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -193,6 +193,7 @@ elastic-agent enroll --fleet-server-es <string>
                      [--fleet-server-policy <string>]
                      [--fleet-server-port <uint16>]
                      [--force]
+                     [--non-interactive]
                      [--help]
                      [--tag <string>] 
                      [--url <string>] <3>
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/commands.asciidoc
```diff
@@ -282,6 +283,14 @@ Service token to use for communication with {es}.
 Force overwrite of current configuration without prompting for confirmation.
 This flag is helpful when using automation software or scripted deployments.
 
+Note: if the {agent} is already installed on the host, using --force will result in
+unpredictable behaviour with duplicate {agent}s appearing in {fleet}
```
dedemorton, line 287:
```suggestion
+
NOTE: If the {agent} is already installed on the host, using `--force` will result in
unpredictable behavior with duplicate {agent}s appearing in {fleet}.
```

docs/en/ingest-management/commands.asciidoc
```diff
@@ -282,6 +283,14 @@ Service token to use for communication with {es}.
...
+
+`--non-interactive`::
+The {agent} will be installed in a non-interactive mode. This flag is helpful when 
+using automation software or scripted deployments and will check if there is an {agent}
+already installed. If there is an {agent} on the host already, installation will abort.
```
dedemorton, line 292:
```suggestion
Install {agent} in a non-interactive mode. This flag is helpful when 
using automation software or scripted deployments. If {agent} is
already installed on the host, the installation will terminate.
```

docs/en/ingest-management/commands.asciidoc
dedemorton: Does this always happen in this situation? If not, you should say "may result".

docs/en/ingest-management/commands.asciidoc
dedemorton: we try to avoid saying abort.

dedemorton (commented):
Some editorial feedback. Thanks for opening this! We'll need to forward port it to `main` and `8.4`.

docs/en/ingest-management/commands.asciidoc
```diff
@@ -282,6 +284,14 @@ Service token to use for communication with {es}.
 Force overwrite of current configuration without prompting for confirmation.
 This flag is helpful when using automation software or scripted deployments.
 
+Note: if the {agent} is already installed on the host, using --force may result in
+unpredictable behaviour with duplicate {agent}s appearing in {fleet}
```
dedemorton, line 288:
```suggestion
NOTE: If the {agent} is already installed on the host, using `--force` may result in
unpredictable behaviour with duplicate {agent}s appearing in {fleet}.
```

dedemorton, line 288:
```suggestion
+
NOTE: If the {agent} is already installed on the host, using `--force` may result in
unpredictable behaviour with duplicate {agent}s appearing in {fleet}.
```

dedemorton: `NOTE` is required if you want this to render using the note formatting. Also added some markup for the flag name and a period at the end of the sentence.

Issue #1936, opened 2022-06-16
[Request] Custom ingest pipeline in Fleet


nchaulet:
## Description

Related Kibana issue  https://github.com/elastic/kibana/issues/133740
We are allowing user to write custom ingest pipeline for their Fleet managed datastream

It should probably be mentioned and documented here https://www.elastic.co/guide/en/fleet/current/data-streams.html 

We can now add a custom pipeline for an integration datastream by creating a pipeline named `{datastream.type}-{datastream.dataset}@custom`

## Collaboration

Contact Person: @nchaulet 

## Suggested Target Release

8.4

## Stakeholders

*(Please list any stakeholders for this issue.)*

joshdover:
cc @akshay-saraswat to add more information here about what we should document

amolnater-qasource:
Hi Team
We have revalidated the docs for Custom Ingest Pipeline at links:
- https://www.elastic.co/guide/en/fleet/current/data-streams.html
- https://www.elastic.co/guide/en/fleet/current/data-streams-pipeline-tutorial.html

**Screenshots:**
![6](https://user-images.githubusercontent.com/77374876/187405956-0b8b0cb2-610f-466e-ae1f-cf430536643f.JPG)

![7](https://user-images.githubusercontent.com/77374876/187405941-4e593613-3d56-4271-b5e8-b30cf42331b5.JPG)

Hence we are marking this as **QA:Validated**.
Thanks

Issue #748, opened 2021-05-28
Document how to use custom ingest pipeline


dedemorton:
Targeted for post-GA.

Copied from https://github.com/elastic/obs-dc-team/issues/147.

Issue #1226, opened 2021-11-04
[APM] Document custom ingest pipelines


bmorelli25:
### Related dev issue

* https://github.com/elastic/package-spec/issues/129

### Background

As of 8.0: The Fleet-managed ingest policy will be overwritten on reinstall/upgrade. Users can define their own custom ingest pipeline if needed, but they need to make sure that they also call the package-provided pipeline.

### Details

1. Fleet adds ingest pipelines for the APM integration using index templates that include pipeline index settings. 
2. Elasticsearch uses the data view in these index templates to match pipelines to APM data streams.
  As an example, the index template `traces-apm`, matches `traces-apm-*` data streams and assigns the `"traces-apm-7.16.0"` pipeline: 

  <img width="606" alt="Screen Shot 2021-10-29 at 1 15 07 PM" src="https://user-images.githubusercontent.com/5618806/139496414-89ea3078-3dd8-4c26-9766-80a299dd9694.png">

3. This Fleet-managed index template ⬆️ will be overwritten on reinstall/upgrade.

Assuming the above is correct, then I see three scenarios:

### Workarounds

1. A user can edit the Fleet-managed index template to point to a new pipeline, ensuring the following are true:
  a. ensure the package-provided pipeline is also called
  b. know that the index template will be overwritten on install/upgrade

2. A user can edit the default pipeline directly, like the `"traces-apm-7.16.0"` pipeline, ensuring the following are true:
  a. the package-provided pipelines are still included (first!)
  b. know that their edits will be overwritten on install/upgrade

### Documentaiton ideas

The following snippet was removed from https://github.com/elastic/apm-server/pull/6940.

```
[float]
[id="custom-pipelines{append-legacy}"]
=== Add a custom ingest pipeline

IMPORTANT: Fleet-managed index templates are overwritten on reinstall/upgrade.
Any customizations to index pipelines are not persisted when upgrading.
When editing {fleet}-managed index templates, ensure that the package-provided pipeline is still called.

// tutorial would be something like this:
. Find the pipeline **Stack Management** > **Index Management** > **Index Templates** > `traces-apm`.
. **Settings** >
+
[source,json]
----
{
  "index": {
    "default_pipeline": "traces-apm-{version}"
  }
}
----

// Edit default
. **Stack Management** > **Ingest Pipelines** > Search for +traces-apm-{version}+.
. **Manage** > **Edit** to edit the default pipeline.

// Add custom pipeline
[...]

PR #2094, opened 2022-08-17
Custom ingest pipelines

bmorelli25:
### Summary

- [x] Moves the custom ILM tutorial to its own sub-page of the Data Streams topic
- [x] Adds a custom ingest pipeline tutorial that follows the same outline as the custom ILM tutorial

Screen Shot 2022-08-16 at 10 52 58 PM

Preview the new tutorial by clicking **[here](https://observability-docs_2094.docs-preview.app.elstc.co/guide/en/fleet/master/data-streams-pipeline-tutorial.html)**.

### Related
* Closes https://github.com/elastic/observability-docs/issues/1936
* Closes https://github.com/elastic/observability-docs/issues/748
* Related to https://github.com/elastic/observability-docs/issues/1226

### DIFF
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -114,12 +114,13 @@ Want to customize your index lifecycle management? See <<data-streams-ilm-tutori
 [[data-streams-pipelines]]
 == Ingest pipelines
 
-{agent} integrations ship with a default ingest pipeline that preprocesses and enriches data before indexing.
-These pipelines should not be edited as it can break the functionality of the integration.
+{agent} integrations ship with a default {ref}/ingest.html[ingest pipeline]
+that preprocesses and enriches data before indexing.
+The default pipeline should not be directly edited as changes can easily break the functionality of the integration.
 
-Starting in version 8.4, all default ingest pipelines reference a non-existent and non-versioned `@custom` ingest pipeline.
-See <<data-streams-pipeline-tutorial>>.
-
+Starting in version 8.4, all default ingest pipelines call a non-existent and non-versioned `@custom` ingest pipeline.
+This pipeline can be edited to add custom data processing, new fields, data sanitization, and more.
+See <<data-streams-pipeline-tutorial>> to get started.
 
 // [discrete]
 [[data-streams-ilm-tutorial]]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/data-streams.asciidoc
```diff
@@ -240,3 +260,188 @@ or force a rollover using the {ref}/indices-rollover-index.html[{es} rollover AP
...
+
+. In {kib}, navigate to **Stack Management** -> **Ingest Pipelines** -> **Create pipeline** -> **New pipeline**.
+
+. Name your pipeline. We'll call this one, `add_field`.
+
+. Select **Add a processor**. Fill out the following information:
++
+** Processor: "Set"
+** Field: `test`
+** Value: `true`
```
bmorelli25, line 290: If anyone has a better idea for an example, I'm all ears.

bmorelli25:
@nchaulet / @joshdover here's a first draft of custom ingest pipeline documentation. I could use your help to:

* make sure I'm on the right track here
* ensure there aren't any important details missing
* flesh out a better example if this one is too simple



@dedemorton you can hold off on a tech writer review for now

docs/en/ingest-management/data-streams.asciidoc
```diff
@@ -108,9 +108,29 @@ By default, these data streams use an {ilm-init} policy that matches their data
 For example, the data stream `metrics-system.logs-*`,
 uses the metrics {ilm-init} policy as defined in the `metrics-system.logs` index template.
 
+Want to customize your index lifecycle management? See <<data-streams-ilm-tutorial>>.
+
 [discrete]
+[[data-streams-pipelines]]
+== Ingest pipelines
+
+{agent} integrations ship with a default {ref}/ingest.html[ingest pipeline]
```
nchaulet, line 117: Should we day `integrations datastreams ship with a default ...` as the relation is not 1 integration => 1 pipeline but 1 integration => n datastreams => 1 pipeline

bmorelli25: Good point 👍

bmorelli25:
I need to merge this now, but I welcome any additional feedback and will address it in a follow-up PR. Thanks!

felixbarny:
I think there's a leftover of outdated docs that could be confusing to users: https://www.elastic.co/guide/en/elasticsearch/reference/8.4/ingest.html#pipeline-custom-logs-configuration

Maybe we should replace this with a link to https://www.elastic.co/guide/en/fleet/current/data-streams-pipeline-tutorial.html?

But maybe there are some subtleties here. Not sure if the latter would work for the "custom logs" integration which doesn't have a default pipeline.

cc @joshdover

bmorelli25:
Catching up after a long vacation. Thanks for pointing this out, Felix! I'll connect with the ES docs team to figure out the best path forward and get this fixed.

AFAIK, the latter should work with all integrations. @joshdover can you verify?

nchaulet:
> @nchaulet can confirm, but my understanding is that it does work with custom logs

Just checked and it does not work with custom logs as we do not generate a correct index template for that stream

PR #2114, opened 2022-08-22
[synthetics] Standardize beta notes across synthetics docs

colleenmcginnis:
In our last synthetics/docs sync, @paulb-elastic pointed out that we weren't consistent with beta notes across synthetics docs. This PR adds a beta note (rather than inline text with a popup) at the beginning of each page related to Elastic Synthetics and/or browser monitors.


dedemorton (approved):
LGTM!

PR #2131, opened 2022-08-24
Consistently use attributes for links to other books

colleenmcginnis:
After the trouble I caused with some links that weren't using shared attributes, this PR updates existing hard-coded URLs to pages in other books to use shared attributes. I just searched for any uses of `https://www.elastic.co/guide/*` in this repo and replaced them with the equivalent from [`shared/attributes.asciidoc`](https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc) (if one exists).


bmorelli25 (approved):
Nice.

I'd recommend backporting this PR much further than 8.4. The problem with hard-coded links to current is that if any of these pages are removed in the future, it'll break every link in every version that they exist in. Some of these hard-coded links go back to 7.12 (well before your time). I'm happy to help take on some of that backporting pain since at least one of these is a link that I didn't catch in a review.

PR #2146, opened 2022-08-25
[synthetics] Add notes about API keys for private locations

colleenmcginnis:
Add notes about API keys for private locations. I'm not sure if the current phrasing is correct so feel free to suggest changes! 

Note: I also added more context to the `push` docs because I don't think it was clear where to generate the API key.


paulb-elastic (approved):
Thanks @colleenmcginnis these updates look great, LGTM

Not part of this PR (so might want a follow up), but mentioning what I put on slack that it might be worth adding some details about the permissions needed when enabling Monitor Management (maybe [here](https://github.com/elastic/observability-docs/pull/2146/files#diff-c993ea63593a86b248a0dcd0c6fea791b213115cebf6408923449bbff7c78f1cR11)) as this came up recently as a problem for a user knowing what permission to use. (cc @dominiqueclarke)

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -83,6 +83,7 @@ npx @elastic/synthetics push --auth <api-key> --url <kibana-url> --project <id|n
 
 `--auth`::
 API key used for {kibana-ref}/api-keys.html[{kib} authentication].
+If you are pushing to a <<private-locations,private location>>, you must use an API key generated in 8.4 or higher.
```
colleenmcginnis, line 86: @dominiqueclarke is this language accurate?

dominiqueclarke: Actually, we may want to be more specific with the requirements, similar to the other PR. For  generating an api key capable of interacting with private locations specifically, users must have write level Kibana permissions for Fleet and Integrations.

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -83,6 +83,13 @@ npx @elastic/synthetics push --auth <api-key> --url <kibana-url> --project <id|n
 
 `--auth`::
 API key used for {kibana-ref}/api-keys.html[{kib} authentication].
++
+If you are pushing to a <<private-locations,private location>>, you must use an API key generated in 8.4 or higher.
++
+To create an API key, you must be logged into {kib} as a user with the following privileges:
++
+* {ref}/security-privileges.html#privileges-list-cluster[Cluster privileges]: `monitor`, `read_ilm`, `read_pipeline`, and at least one of `manage_own_api_key`, `manage_security`, or `manage_api_key`
+* {ref}/security-privileges.html#privileges-list-indices[Indices privileges] for `synthetics-*`: `view_index_metadata`, `create_doc`, `auto_configure`
```
colleenmcginnis, line 92: @dominiqueclarke I'm not sure if the requirements for creating an API key are identical to those added in #2148. If they aren't can you suggest changes? Thanks!

dominiqueclarke: Hi @colleenmcginnis 

Sorry I missed this. The privileges are completely different, unfortunately, and nuanced.

To generate an api key, you need `manage_own_api_key`, `manage_security`, or `manage_api_key` cluster permissions.

```
* {ref}/security-privileges.html#privileges-list-cluster[Cluster privileges]: At least one of `manage_own_api_key`, `manage_security`, or `manage_api_key`
```

Plus AT LEAST Kibana Uptime all permissions if you want to use ONLY PUBLIC LOCATIONS

image

If you want to use PRIVATE LOCATIONS, you also need Fleet and Integrations Kibana permissions
image

colleenmcginnis: @dominiqueclarke does this look correct? 

Screen Shot 2022-09-08 at 3 16 58 PM

dominiqueclarke (approved):
LGTM

PR #2161, opened 2022-09-06
[DOCS] Add "attributes" to display {version}

lockewritesdocs:
An internal user indicated that the `{version}` variable isn't rendering on the page for [running an Elastic Agent in a container](https://www.elastic.co/guide/en/fleet/master/elastic-agent-container.html). This change adds the `subs="attributes"` parameter to several code examples, which allows the `{version}` variable to be substituted with the current version.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
@@ -78,7 +78,7 @@ Refer to <<agent-environment-variables>> for all available options.
 
 If you're running a self-managed cluster and want to run your own {fleet-server}, run the following command, which will spin up {agent} and {fleet-server} in a container:
 
-[source,terminal,subs="attributes"]
+[source,sh,subs="attributes"]
 ----
 docker run \
   --env FLEET_SERVER_ENABLE=true \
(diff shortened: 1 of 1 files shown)


dedemorton:
@lockewritesdocs I'm not sure if this is related to your callout problem, but there's another problem with turning on attribute substitution: some of these example blocks use curly braces to denote variable values. (If the doc build doesn't find an asciidoc attribute, it renders as plain text.) Elastic docs are a bit inconsistent with how we show variable values, but most docs use angle braces. So...if you turn on attribute substitution in these code blocks, you also need to change the variable values to use angle brackets. For example, `{enrollment-token}` would be `<enrollment-token>`.

dedemorton:
@lockewritesdocs I'm not sure if this is related to your callout problem, but there's another problem with turning on attribute substitution: some of these example blocks use curly braces to denote variable values. (If the doc build doesn't find an asciidoc attribute, it renders as plain text.) Elastic docs are a bit inconsistent with how we show variable values, but most docs use angle braces. So...if you turn on attribute substitution in these code blocks, you also need to change the variable values to use angle brackets. For example, `{enrollment-token}` would be `<enrollment-token>`.

lockewritesdocs:
> So...if you turn on attribute substitution in these code blocks, you also need to change the variable values to use angle brackets. For example, `{enrollment-token}` would be `<enrollment-token>`.

@dedemorton, thanks for pointing out this discrepancy 👍 I updated any variables that we're not replacing to use angle brackets (`<>`), and was also able to include substitutions in examples with callouts. Please review the latest changes when you can.

dedemorton (approved):
LGTM. Thanks for fixing this!

PR #2201, opened 2022-09-27
Adds Fleet/Agent 8.4.3 release notes

dedemorton:
Adds Elastic Agent release notes based on  https://github.com/elastic/elastic-agent/commits/5f09295a5dce74a20656ed5bc1d654b1fab06060 in BC2.

There are no Fleet release notes in the [Kibana relnotes PR](https://github.com/elastic/kibana/pull/141808), so I'm assuming no release notes for Fleet. Please confirm.

### DIFF
--- a/docs/en/ingest-management/release-notes/release-notes-8.4.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-8.4.asciidoc
@@ -28,21 +28,31 @@ Also see:
 [[release-notes-8.4.3]]
 == {fleet} and {agent} 8.4.3
 
-//Review important information about the {fleet} and {agent} 8.4.3 release.
+Review important information about the {fleet} and {agent} 8.4.3 release.
 
 [discrete]
 [[bug-fixes-8.4.3]]
 === Bug fixes
 
-There are no bug fixes for {fleet} or {agent} in this release.
+{fleet}::
+* No bug fixes for this release
 
-//{fleet}::
-//* TBD
-//Check https://github.com/elastic/kibana/pull/141808 for Fleet relnotes
+{agent}::
+* Use at least warning level for all status logs {agent-pull}1218[#1218]
+* Fix unintended reset of source URI when downloading components {agent-pull}1252[#1252]
+* Create separate status reporter for local only events so that degraded
+{fleet} checkins no longer affect health on successful {fleet} checkins {agent-issue}1157[#1157] {agent-pull}1285[#1285]
+* Add success log message after previous checkin failures {agent-pull}1327[#1327]
 
-//{agent}::
-//* TBD
-//Check build candidate commit history for Elastic Agent for changes.
+[discrete]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/release-notes/release-notes-8.4.asciidoc
```diff
@@ -22,11 +23,46 @@ Also see:
...
+
+{fleet}::
+* No bug fixes for this release
+
+{agent}::
+* Use at least warning level for all status logs {agent-pull}1218[#1218]
+* Fix unintended reset of source URI when downloading components {agent-pull}1252[#1252]
+* Create separate status reporter for local only events so that degraded
+{fleet} checkins no longer affect health on successful {fleet} checkins {agent-issue}1157[#1157] {agent-pull}1285[#1285]
+* Add success log message after previous checkin failures {agent-pull}1327[#1327]
```
cmacknz, line 45:
```suggestion
* Add success log message after previous checkin failures {agent-pull}1327[#1327]
* Fix Unix socket connection errors when using diagnostics command {agent-pull}2201[#2201]
```

There was no changelog entry for https://github.com/elastic/elastic-agent/pull/1034, so let's add this one.

cmacknz:
Do we include Fleet server changes in these notes as well? There are a few changes there we could include. This is the commit for BC2: https://github.com/elastic/fleet-server/commits/58b61b4f7db6e4e90749cd52755680ec233d02a5

This is the only notable change, which includes a changelog we could add here: https://github.com/elastic/fleet-server/commit/3afc7b5625f9b33bca7e8af1ba8fea7d52e07769

jen-huang (approved):
Ditto to check what @cmacknz mentioned, otherwise no notes from Fleet side.

dedemorton:
>Do we include Fleet server changes in these notes as well?

@cmacknz We don't at the moment because I was never asked to add this changelong and didn't know it existed. I think we should cover Fleet Server. To avoid confusion with existing release notes, I will just add it to the Fleet section for now, but we should consider having a separate section for Fleet Server in the future.

One thing, though, this changelog message doesn't seem very user friendly:

"LoadServerLimits will not overwrite specified limits when loading default/agent number specified values."

 Can we say this instead?

"Fleet Server no longer silently discards user-specified values for cache and server limits when defaults are loaded."

I'm not sure if I have all the technical details correct here. But you can see that passive voice makes the sentence ambiguous, and talking about code internals will probably be confusing to end users (the audience for the release notes).

We should focus on how this fix affects users. Again, I might be misunderstanding something here.... If my suggestion isn't correct, can you help me craft a better changelog entry?

cmacknz:
> "Fleet Server no longer silently discards user-specified values for cache and server limits when defaults are loaded."

This is much clearer than the previous changelog entry, thanks!

cmacknz (approved):

dedemorton:
@Mergifyio update

PR #2213, opened 2022-09-30
Explain why output settings might be grayed out in Fleet

dedemorton:
This came up in a discuss topic: https://discuss.elastic.co/t/endpoint-security-configuration/315486


nchaulet (approved):
Looks good to me, 
I am wondering if we could do a better job documenting the Kibana settings, like explaining that the defined output will not be editable in Kibana UI and removing it from Kibana config file will make editable in Kibana UI.

dedemorton:
>I am wondering if we could do a better job documenting the Kibana settings, like explaining that the defined output will not be editable in Kibana UI and removing it from Kibana config file will make editable in Kibana UI.

Good idea! I will open a PR in Kibana to make this clear.

PR #2224, opened 2022-10-05
Add docs about package verification

### LINKED ISSUE #2005: [Request][Fleet] Package signatures documentation
## Description

As part of https://github.com/elastic/kibana/issues/133822 we are adding package signature verification to Kibana. 

This is the same signing that we use for our code releases. e.g elastic binaries. Package signatures prevent attackers from modifying package content, when a package is downloaded, Kibana verifies the packages signature against Elastics public key meaning the package content can be truested. 

If a user tries to install a package which fails verification, they are presented with a modal which gives them the option to force install.

The original issue has some more details about why package signing is useful https://github.com/elastic/package-spec/issues/46

We need some content for the learn more link from the designs:

https://user-images.githubusercontent.com/6766512/172449124-5c50b54b-bb23-4de6-87a1-bfc69915743b.png
https://user-images.githubusercontent.com/6766512/172449239-c6d7e30a-5e3e-4fc6-b19a-2e74675782d2.png

## Collaboration
- The docs team will define with product team the structure and location, and the product team will provide the initial content

I am happy to have a stab at writing the document but I could maybe do with a chat with someone about the ideal structure/content

**Contact Person:** @hop-dev 

## Suggested Target Release

8.4

dedemorton:
Closes #2005 

Please review. Some of the UI-related details are fuzzy because the feature is not yet turned on in the build.

Remaining questions/issues:

- [x] @kpollich Which versions (if any) should I backport this to?
- [x] @elastic/ecosystem Will the docker image for the air gapped registry will have signatures?
- [x] @hop-dev Who is updating the docs at https://www.elastic.co/guide/en/kibana/master/fleet-settings-kb.html to show the package verification setting?
- [x] @dedemorton Need to verify this info by looking at the UI when the feature is turned on. (See if this is necessary.)
- [x] It sounds like we are blocked and shouldn't merge this until https://github.com/elastic/ingest-dev/issues/1193 is resolved.

### DIFF
--- a/docs/en/ingest-management/integrations/package-signatures.asciidoc
+++ b/docs/en/ingest-management/integrations/package-signatures.asciidoc
@@ -1,8 +1,6 @@
 [[package-signatures]]
 = Package signatures
 
-coming[8.5]
-
 All integration packages published by Elastic have package signatures that
 prevent malicious attackers from tampering with package content. When you
 install an Elastic integration, {kib} downloads the package and verifies the
@@ -14,9 +12,6 @@ IMPORTANT: By installing an unverified package, you acknowledge that you
 assume any risk involved.
 
 To force installation of an unverified package:
-
-//TODO: Need to verify this info when package verification is available
-//in the build.
 
 * When using the {integrations} UI, you'll be prompted to confirm that you want
 to install the unverified integration. Click **Install anyway** to force
@@ -71,14 +66,3 @@ verified integration packages will continue to show as verified until new
 packages are installed or existing ones are upgraded. If this happens, you can
 set the `xpack.fleet.packageVerification.gpgKeyPath` setting in the `kibana.yml`
 configuration file to use the new key. 
-
-//REVIEWERS: Who is updating the docs at
-// https://www.elastic.co/guide/en/kibana/master/fleet-settings-kb.html to show
-//the package verification setting?
-
-
-[discrete]
-== Open questions for reviewers:
-
-- Does this change affect the https://www.elastic.co/guide/en/fleet/current/air-gapped.html[air gapped docs]?
-
--- a/docs/en/ingest-management/integrations/package-signatures.asciidoc
+++ b/docs/en/ingest-management/integrations/package-signatures.asciidoc
@@ -1,12 +1,34 @@
 [[package-signatures]]
 = Package signatures
 
+coming[8.5]
+
 All integration packages published by Elastic have package signatures that
 prevent malicious attackers from tampering with package content. When you
 install an Elastic integration, {kib} downloads the package and verifies the
 package signature against a public key. If the package is unverified, you can
 choose to force install it. However, it's strongly recommended that you avoid
-installing unverified packages unless they are from trusted sources.
+installing unverified packages.
+
+IMPORTANT: By installing an unverified package, you acknowledge that you
+assume any risk involved.
+
+To force installation of an unverified package:
+
+//TODO: Need to verify this info when package verification is available
+//in the build.
+
+* When using the {integrations} UI, you'll be prompted to confirm that you want
+to install the unverified integration. Click **Install anyway** to force
+installation.
+
+* When using the {fleet} API, if you attempt to install an unverified package,
+you'll see a 400 response code with a verification failed message. To force
+installation, set the URL parameter `ignoreUnverified=true`. For more
+information, refer to <<fleet-api-docs>>.
+
+After installation, unverified {integrations} are flagged on the
+**Installed integrations** tab of the {integrations} UI.
 
 [discrete]
 [[why-verify-packages]]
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/integrations/package-signatures.asciidoc
```diff
@@ -1,4 +1,49 @@
...
+won't get flagged as "unverified"?
+- What actions, if any, should Kibana users take when they see that a package is
+unverified? If it's a package published by Elastic, should they try reinstalling
+or upgrading it?
+- Can anyone who has access to the integrations UI install unverified packages?
+Or does it require additional privileges?
+- Is there anything else missing from this topic that we need to cover? Like how
+to download a signing key and verify a package?
+- Does this change affect the air gapped docs at
+https://www.elastic.co/guide/en/fleet/current/air-gapped.html?
```
dedemorton, line 48:
```suggestion
- Does this change affect the https://www.elastic.co/guide/en/fleet/current/air-gapped.html[air gapped docs]?
```

docs/en/ingest-management/integrations/package-signatures.asciidoc
```diff
@@ -1,4 +1,48 @@
...
+corrupted or otherwise tampered with.
+
+[discrete]
+== Open questions for reviewers:
+
+- How do users verify Elastic packages when installing them through the API? Do
+they need to download Elastic's GPG key from
+https://artifacts.elastic.co/GPG-KEY-elasticsearch and store it on disk? What
+else?
+- What if a key has to be invalidated in the future? Do we need to document
```
hop-dev, line 31: > What if a key has to be invalidated in the future?

If the elastic signing key changed, then existing package installs would still show as verified (as they were verified at the time of install) but they would then show as unverified when new packages are installed or existing ones are upgraded. You are right that the user can update the ocnfig (see answer below) to make this work again. 

Some more info on this here: https://github.com/elastic/elastic-package/issues/576#issuecomment-1197867649

> Would users provide the file path in kibana.yml (through the xpack.fleet.packageVerification.gpgKeyPath setting)?

Yes this is what they would have to do for existing deployments, the key would be updated on kibana upgrade as well.

>  Do we need to worry about documenting this now?

Maybe a small nod to the `xpack.fleet.packageVerification.gpgKeyPath` setting? Not sure!

docs/en/ingest-management/integrations/package-signatures.asciidoc
```diff
@@ -1,4 +1,48 @@
...
+xpack.fleet.packageVerification.gpgKeyPath setting)? Do we need to worry about
+documenting this now?
+- Can users trust all packages that aren't flagged as "unverified," or are there
+some situations where they won't see the flag or be prompted to force install?
+- Is there some way for 3rd party contributors to sign their packages so they
+won't get flagged as "unverified"?
+- What actions, if any, should Kibana users take when they see that a package is
+unverified? If it's a package published by Elastic, should they try reinstalling
+or upgrading it?
+- Can anyone who has access to the integrations UI install unverified packages?
```
hop-dev, line 43: - Can anyone who has access to the integrations UI install unverified packages?

Currently yes

docs/en/ingest-management/integrations/package-signatures.asciidoc
```diff
@@ -1,4 +1,48 @@
...
+Or does it require additional privileges?
+- Is there anything else missing from this topic that we need to cover? Like how
+to download a signing key and verify a package?
+- Does this change affect the https://www.elastic.co/guide/en/fleet/current/air-gapped.html[air gapped docs]?
```
hop-dev, line 47: The docker image will have signatures when it is updated I believe @mtojek? the only complication is that kibana supports verification from 8.4 onwards but the registry may only have signatures from a later version? So if this doc went into 8.4 the 8.4 registry docker image wouldn't provide signatures and verification wouldnt be performed, but it wouldnt break anything

dedemorton: @mtojek Can you let me know if the docker image for the air gapped registry will have signatures? I need to know if the air gapped docs need to be updated, too.

dedemorton:
I've marked this PR as ready for review because I've taken it as far as I can without seeing the UI in a build. @hop-dev Are we still planning to turn this on for 8.5?

dedemorton:
@hop-dev Has this feature made it into 8.5?  I need someone to review the content.

@mtojek Can you please respond to the question about whether the air gapped registry will have signatures?

mrodm:
> @mtojek Can you please respond to the question about whether the air gapped registry will have signatures?

Currently, the docker images that are built for air-gapped envs and used in those Elastic Package registries do not contain the signatures. But we plan to include them soon. This change should be done after switching the DNS to start using package storage v2.

cc @jsoriano

dedemorton:
Added "blocked" label because I am blocked until https://github.com/elastic/ingest-dev/issues/1193 is merged.

@jen-huang While we're waiting for the DNS switch to v2, can someone please review the content here (and answer my questions) so that this content is ready to go when the feature goes live? Some insight into current plans would also be useful. Thanks!

jlind23:
@hop-dev @kpollich could you please review this content as asked by @dedemorton ?

kpollich (approved):
Thanks for this - docs look good. I have filed a separate PR updating the Fleet settings docs, and if the airgapped docs need an update I think it'd be fine to file a followup.

docs/en/ingest-management/integrations/package-signatures.asciidoc
mrodm: > @jsoriano @mrodm - will the `docker.elastic.co/package-registry/distribution:8.6.0` contain `signature_path` values in these responses?

@kpollich  Yes, with the new Package Storage V2 being launched, Jenkins jobs have been updated to generate the docker distribution images with the signatures. Thereby, the following docker distribution images will have the `signature_path` field for all packages.

Just tried locally one of the latest production images built and pushed in Jenkins and that field is already present:

```shell
 $ curl -s "http://localhost:8080/search?package=system" |jq -r '.[] | ."signature_path"'
/epr/system/system-1.20.4.zip.sig
```

dedemorton:
@kpollich A couple of quick follow-up questions before I merge this:
* Which versions (if any) should I backport this to?
* Is there an easy way to install an unverified package to confirm the details here? Or maybe that's not necessary since you have reviewed the info?

kpollich:
> Which versions (if any) should I backport this to?

We can backport this to 8.5.

> Is there an easy way to install an unverified package to confirm the details here? Or maybe that's not necessary since you have reviewed the info?

I don't think this is necessary. We ran through these scenarios in our testing/development cycle and everything documented here looks accurate. @hop-dev's previous review covered any of the functionality specifics, etc so I think we're good here.

PR #2236, opened 2022-10-10
Update view-integration-policies.asciidoc

redcinelli:
I feel it could be nice to reference the list of available integrations in the documentation.

### DIFF
--- a/docs/en/ingest-management/integrations/view-integration-policies.asciidoc
+++ b/docs/en/ingest-management/integrations/view-integration-policies.asciidoc
@@ -5,8 +5,6 @@
 <titleabbrev>View integration policies</titleabbrev>
 ++++
 
-:integrations_link: https://www.elastic.co/integrations/data-integrations
-
 An integration policy is created when you add an {integrations-docs}[integration] to an {agent}
 policy.
 
--- a/docs/en/ingest-management/integrations/view-integration-policies.asciidoc
+++ b/docs/en/ingest-management/integrations/view-integration-policies.asciidoc
@@ -7,7 +7,7 @@
 
 :integrations_link: https://www.elastic.co/integrations/data-integrations
 
-An integration policy is created when you add an {integrations_link}[integration] to an {agent}
+An integration policy is created when you add an {integrations-docs}[integration] to an {agent}
 policy.
 
 To view details about all the integration policies for a specific integration:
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/integrations/view-integration-policies.asciidoc
```diff
@@ -5,7 +5,9 @@
 <titleabbrev>View integration policies</titleabbrev>
 ++++
 
-An integration policy is created when you add an integration to an {agent}
+:integrations_link: https://www.elastic.co/integrations/data-integrations
+
+An integration policy is created when you add an {integrations_link}[integration] to an {agent}
```
dedemorton, line 10:
```suggestion
An integration policy is created when you add an {integrations-docs}[integration] to an {agent}
```

docs/en/ingest-management/integrations/view-integration-policies.asciidoc
```diff
@@ -5,7 +5,9 @@
 <titleabbrev>View integration policies</titleabbrev>
 ++++
 
-An integration policy is created when you add an integration to an {agent}
+:integrations_link: https://www.elastic.co/integrations/data-integrations
+
```
dedemorton, line 9:
```suggestion
```

dedemorton: Good idea, but let's point to the integrations documentation rather than the marketing page (I think the marketing page still has some links to Beats modules, which might be confusing to users.) Note that we already have a [shared attribute](https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc) defined for that book.

dedemorton (commented):
Suggested some changes.

Usually we open new PRs against main rather than a versioned branch. This change will need to be manually forward ported to main after merging. Do you want me to do that, or will you take care of that?

redcinelli:
Hi @dedemorton,

Thanks a lot for your review.
I am sorry I click on the ✏️  in the documentation and it led me right here. I'll be more careful next time.

Feel free to do it I am not sure I'll be as comfortable as you doing it.

dedemorton:
@redcinelli No worries. It's a common problem. I'll take care of the backports.

dedemorton (approved):
LGTM

Issue #2115, opened 2022-08-22
[Request] Add links to more detailed synthetics `request` examples


colleenmcginnis:
## Description

Add links to more detailed Elastic Synthetics `request` examples in [/synthetics-demo](https://github.com/elastic/synthetics-demo). Existing example code:

* https://github.com/elastic/observability-docs/issues/2049#issuecomment-1201343573 by @awahab07
* https://github.com/elastic/observability-docs/issues/2049#issuecomment-1202923144 by @bhoskins7
* https://github.com/elastic/observability-docs/pull/2071#pullrequestreview-1080263961 by @paulb-elastic 

## Collaboration

- [x] The product team will add example code to [/synthetics-demo](https://github.com/elastic/synthetics-demo)
- [ ] The docs team will add links in [Write a synthetic test > Make API requests](https://observability-docs_2071.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-create-test.html#synthetics-request-param) to the example code

**Contact Person:** TBD

## Suggested Target Release

post-8.4

## Stakeholders

* @paulb-elastic 
* @andrewvc 
* @awahab07
* @bhoskins7
* @vigneshshanmugam

paulb-elastic:
@colleenmcginnis this has been done hasn't it, can this be closed?

colleenmcginnis:
@paulb-elastic it looks like I didn't go back and add the link after the example was created. Fixed in https://github.com/elastic/observability-docs/pull/2243.

PR #2243, opened 2022-10-10
[synthetics] Link to `request` parameter example

colleenmcginnis:
Closes https://github.com/elastic/observability-docs/issues/2115

Adds a link to the [`request` parameter example](https://github.com/elastic/synthetics-demo/blob/main/advanced-examples/journeys/api-requests.journey.ts) in the [/synthetics-demo](https://github.com/elastic/synthetics-demo) repo.

### DIFF
--- a/docs/en/observability/synthetics-create-test.asciidoc
+++ b/docs/en/observability/synthetics-create-test.asciidoc
@@ -207,7 +207,7 @@ with the corresponding https://playwright.dev/docs/api/class-browsercontext[`Bro
 via <<elastic-synthetics-command, `--playwright-options`>> or in the
 <<synthetics-configuration, `synthetics.config.ts` file>>.
 
-See a full example using the `request` parameter in the https://github.com/elastic/synthetics-demo/blob/main/advanced-examples/journeys/api-requests.journey.ts[Elastic Synthetics demo repository].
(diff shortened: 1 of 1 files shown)


docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -207,6 +207,8 @@ with the corresponding https://playwright.dev/docs/api/class-browsercontext[`Bro
 via <<elastic-synthetics-command, `--playwright-options`>> or in the
 <<synthetics-configuration, `synthetics.config.ts` file>>.
 
+See a full example using the `request` parameter in the https://github.com/elastic/synthetics-demo/blob/main/advanced-examples/journeys/api-requests.journey.ts[Elastic Synthetics demo repository].
```
dedemorton, line 210: "full example using" is grammatically ambiguous...could be read as if the request parameter is used to see the example. Also the accessibility guidelines tell us to avoid using "see" because it's ableist (not sure I agree, but that's the guideline).

How about saying something like this (to match the comment in the example file):

For a full example that shows how to use the `request` object, refer to  the....

(Not sure whether you should use `request` parameter or `request` object but it would be good for the text here to match the comment in the example.)

paulb-elastic (approved):
LGTM

dedemorton (approved):
LGTM

PR #2269, opened 2022-10-13
Document Fleet API examples for creating agent policies

### LINKED ISSUE #2180: [Request][Fleet] Document agent policy/package policy API
## Description

We made some progress on Fleet API, to create an agent policy or a package policy, it will be a good improvments to reflect this in the guide, maybe we can have an how to guide on how to create a package policy via the API

https://github.com/elastic/kibana/pull/139658
https://github.com/elastic/kibana/pull/139045
https://github.com/elastic/kibana/pull/139420

We could document how to add an agent policy, a package policy, and retrieve the enrollment token to add agent

#### Authentication 

To make request to Kibana Fleet API user should be authenticated https://www.elastic.co/guide/en/kibana/master/api.html#api-authentication

#### Create an agent policy 

Could be done by calling `POST /api/fleet/agent_policies`

Screen Shot 2022-10-04 at 1 06 57 PM

curl example 
```
curl --request POST \
  --url 'http://localhost:5601/api/fleet/agent_policies?sys_monitoring=true' \
  --header 'Accept: */*' \
  --header 'Authorization: Basic xxxxx' \
  --header 'Cache-Control: no-cache' \
  --header 'Connection: keep-alive' \
  --header 'Content-Type: application/json' \
  --header 'kbn-xsrf: xxx' \
  --data '{
  "name": "Agent policy  1",
  "description": "",
  "namespace": "default",
  "monitoring_enabled": [
    "logs",
    "metrics"
  ]
}'
```

#### Create a package policy

You need to call `POST /api/fleet/package_policies` 
User may need to go to Fleet UI and look at the API reference tab to know what inputs are available for an integration

Screen Shot 2022-10-04 at 1 10 57 PM

Screen Shot 2022-10-04 at 1 07 46 PM

Example curl
```
curl --request POST \
  --url http://localhost:5601/api/fleet/package_policies \
  --header 'Authorization: Basic xxxx' \
  --header 'Content-Type: application/json' \
  --header 'kbn-xsrf: xx' \
  --data '{
  "name": "nginx-demo-123",
  "policy_id": "665303e0-434e-11ed-bc2a-95f65f653f79",
  "package": {
    "name": "nginx",
    "version": "1.5.1"
  },
  "inputs": {
    "nginx-logfile": {
      "streams": {
        "nginx.access": {
          "vars": {
            "tags": [
              "test"
            ]
          }
        },
        "nginx.error": {
          "vars": {
            "tags": [
              "test"
            ]
          }
        }
      }
    }
  }
}'
```

### Retrieve an enrollment token 

Screen Shot 2022-10-04 at 1 12 05 PM

## Collaboration
*(Choose the expected collaboration model and delete the others.)* 

- The product team will provide the initial content and the docs team will edit / review
- The docs team will define with product team the structure and location, and the product team will provide the initial content
- Other (please describe)

Contact Person: @nchaulet 

## Suggested Target Release

We planned to release this for 8.5

dedemorton:
Closes #2180 

Main preview link > https://observability-docs_2269.docs-preview.app.elstc.co/guide/en/fleet/master/fleet-api-docs.html

To preview other changes, go to the [diff preview](https://observability-docs_2269.docs-preview.app.elstc.co/diff) and look for files with actual changes. Please respond to my questions here.

Open questions:

- [x] What is correct: Fleet Kibana API or Kibana Fleet API? (I see both used)
- [x] The UI uses integration policy. Is there some difference between a package policy and an integration policy? I am assuming they are the same. If so, are there any plans to fix this naming inconsistency?
- [x] I changed the request examples to use ApiKey instead of basic auth because it seemed more like a best practice. Is that OK?
- [x] I changed the URLs in the examples to avoid using localhost and show something that looks more like what cloud users would see.  Is that OK?
- [x] The section on enrollment tokens needs more work. The text in the issue didn’t include a sample API call. The one I used works and returns a list of token info that I _assume_ represents all valid tokens. Please let me know if anything is wrong/missing. 
- [x] I  don't show the response for `package_policies` because it's super longggg and maybe less useful than the other responses I show. Is that OK?


kpollich:
> What is correct: Fleet Kibana API or Kibana Fleet API? (I see both used)

This is a great question. I think "Kibana Fleet API" is slightly more appropriate as we're describing the "Fleet API" served by "Kibana". It might also be better to just remove Kibana references altogether and simply refer to the "Fleet API" with a top-level note about how the Fleet API is exposed by Kibana?

> The UI uses integration policy. Is there some difference between a package policy and an integration policy? I am assuming they are the same. If so, are there any plans to fix this naming inconsistency?

We should use "integration policy" over "package policy" in public facing interfaces. "Package policy" is a legacy term and adds confusion when we talk about "Elastic Agent Integrations". We have an issue to track completely removing the "package policy" nomenclature wherever it appears here: https://github.com/elastic/kibana/issues/126337. It's worth prioritizing this before the Fleet API is considered GA, so I'll handle escalating that.

> I changed the request examples to use ApiKey instead of basic auth because it seemed more like a best practice. Is that OK?

I don't see a problem with this. Basic auth is a bit simpler but API keys are a better practice. 

> I changed the URLs in the examples to avoid using localhost and show something that looks more like what cloud users would see. Is that OK?

Fine by me 👍 

> The section on enrollment tokens needs more work. The text in the issue didn’t include a sample API call. The one I used works and returns a list of token info that I assume represents all valid tokens. Please let me know if anything is wrong/missing.

@nchaulet can you help with this?

> I don't show the response for package_policies because it's super longggg and maybe less useful than the other responses I show. Is that OK?

I'm happy with truncating some responses for sure. Maybe we can just `...` after a few lines?

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
```diff
@@ -156,6 +156,194 @@ integration, go to **Integrations**, select an {agent} integration, and click
 **API reference**.
 ****
 
+Example response (truncated for readability):
```
dedemorton, line 159: Decided to preserve a bit more of this response because it's pretty quick to scroll through it.

docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -104,16 +104,19 @@ The following table illustrates the {fleet} user actions available to different
...
 +
 [role="screenshot"]
 image::images/create-agent-policy.png[{fleet} in {kib}]
-
-. Click *Create agent policy*.
++
+. Create the agent policy:
+* To use the UI, click **Create agent policy**.
+* beta[] To use the {fleet} API, click **Preview API request** and run the
+request.
```
colleenmcginnis, line 119: What do you think of making it more explicit that the reader should choose one of the two options? (I know you are indicating that with the unordered list, but it might be worth reinforcing that with words, too.) I'm not sure what the right words are, but maybe something like... 

```suggestion
. Create the agent policy. There are two ways to choose from:
* To use the UI, click **Create agent policy**.
* beta[] To use the {fleet} API, click **Preview API request** and run the
request.
```

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
```diff
@@ -1,10 +1,450 @@
 [[fleet-api-docs]]
-= {fleet} APIs
+= {kib} {fleet} APIs
 
 experimental[]
```
colleenmcginnis, line 4: Should this be `beta`? Or should references to the Fleet API in [`agent-policies.asciidoc`](https://github.com/elastic/observability-docs/pull/2269/files#diff-977a9a6da7dcf823ccc6e8295230c9bd5a69b49c9cde8ba62059561cd3ba11a2R118) be `experimental`? Or are these not the same thing?

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
```diff
@@ -1,10 +1,450 @@
...
+----
+<1> Make a note of the policy ID. You'll need the policy ID to add integration
+policies.
+
+[discrete]
+[[create-integration-policy-api]]
+== Create integration policy
+
+To create a integration policy (also known as an package policy) and add it an 
+existing agent policy, call `POST /api/fleet/package_policies`.
```
colleenmcginnis, line 97:
```suggestion
To create a integration policy (also known as an package policy) and add it to an 
existing agent policy, call `POST /api/fleet/package_policies`.
```

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
```diff
@@ -1,10 +1,450 @@
...
+
+This cURL example creates a integration policy for Nginx and adds it to the
+agent policy created in the previous example:
```
colleenmcginnis, line 100:
```suggestion
This cURL example creates an integration policy for Nginx and adds it to the
agent policy created in the previous example:
```

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
```diff
@@ -1,10 +1,450 @@
...
+from the Dev Tools console. 
++
+. Go to **Integrations**, select an {agent} integration, and click
+**Add <Integration>**.
+. Configure the integration settings and select which agent policy to use.
+. Click **Preview API request**.
++
+If you're creating the integration policy for a new agent policy, the preview
+shows two requests: one to create the agent policy, and another to create the
+integration policy.
```
colleenmcginnis, line 150: Do you need a comma here? 

```suggestion
If you're creating the integration policy for a new agent policy, the preview
shows two requests: one to create the agent policy and another to create the
integration policy.
```

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
```diff
@@ -1,10 +1,450 @@
...
+
+
+[discrete]
+[[get-enrollment-token-api]]
+== Get enrollment tokens
+
+To get a list of valid enrollment tokens from {fleet}, call
+`GET /api/fleet/enrollment_api_keys`. 
+
+This cURL example returns list of enrollment tokens.
```
colleenmcginnis, line 354:
```suggestion
This cURL example returns a list of enrollment tokens.
```

docs/en/ingest-management/agent-policies.asciidoc
colleenmcginnis: Hmm. After reading the rest of the diff, I'm not sure if this is a good suggestion, but I'll leave it here for you to decide. 🙂

colleenmcginnis (commented):
Some small suggestions below!

docs/en/ingest-management/agent-policies.asciidoc
dedemorton: @colleenmcginnis Generally it's best to avoid introducing a list with the number of items that it contains, but I could say, "Do one of the following." I like to be a minimalist if something seems obvious, though. WDYT?

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
dedemorton: The API preview feature described in agent-policies.asciidoc is beta, not the API. I guess this is a bit ambiguous, but I'm not sure how to make it more clear.

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
dedemorton: Here I think the comma actually helps with readability, so I'm leaving it.

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
dedemorton: Great to have another pair of eyes on this. :-)

docs/en/ingest-management/agent-policies.asciidoc
colleenmcginnis: >Generally it's best to avoid introducing a list with the number of items that it contains

Ah yes, that makes sense. I'm fine with either adding "Do one of the following" or leaving it as is -- whichever you feel is better.

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
colleenmcginnis: I'm not sure if it really matters. The "consistency alarm" just went off in my brain when I saw them.

colleenmcginnis (approved):
Looks good from the docs side! 🙃

docs/en/ingest-management/agent-policies.asciidoc
dedemorton: At this point, I will err on the side of lazy and leave it as-is. :-)  We can always clean it up later.

docs/en/ingest-management/fleet/fleet-api-docs.asciidoc
dedemorton: This is a _good thing_! 🥇

jlind23:
@kpollich / @juliaElastic can we have your eyes on this?

juliaElastic:
Can we remove the `list` property from the response of get enrollment tokens query from the docs?
The `list` property is deprecated, and the `items` property contains the same data (the removal captured in [this](https://github.com/elastic/kibana/issues/121485) issue).

kpollich (approved):
Thanks @dedemorton this looks great!

Issue #2219, opened 2022-10-04
[Synthetics] Document how to import modules and their limitations


paulb-elastic:
## Description

The Synthetics documentation should include details about how [modules can be imported](https://github.com/elastic/synthetics/releases/tag/v1.0.0-beta.34) in Project Monitors.  There are certain limitations as described in the link above that we should include:

* Packages that use native bindings will not work
* Users cannot push bundles that are more than 800 Kilobytes

The main documentation that describes the format of journeys is shown [here](https://www.elastic.co/guide/en/observability/current/synthetics-create-test.html), so that _feels_ like the best place for it to go, although Project Monitors are referenced [here](https://www.elastic.co/guide/en/observability/current/synthetic-run-tests.html#synthetic-monitor-choose-project).

Maybe the [main docs page](https://www.elastic.co/guide/en/observability/current/synthetics-create-test.html) should describe how to use it, but clarify it only works for Project Monitors, with a link through to [that page](https://www.elastic.co/guide/en/observability/current/synthetic-run-tests.html#synthetic-monitor-choose-project)?

Happy to go with the docs team’s suggestion here.

## Collaboration

- The docs team will lead producing the content with support from @vigneshshanmugam

## Suggested Target Release

* 8.5.0

## Stakeholders

* @vigneshshanmugam
* @paulb-elastic 
* @andrewvc

PR #2277, opened 2022-10-13
[synthetics] Document how to import NPM packages and their limitations

colleenmcginnis:
Closes https://github.com/elastic/observability-docs/issues/2219

Documents how to import and use external NPM packages in journey code and the limitations when pushing a project that uses external packages.

cc @paulb-elastic

### DIFF
--- a/docs/en/observability/synthetics-command-reference.asciidoc
+++ b/docs/en/observability/synthetics-command-reference.asciidoc
@@ -94,9 +94,13 @@ it using a _different_ ID. Select `yes` to create duplicates of all monitors in 
 
 [NOTE]
 ====
-If the journey code in a project uses an external Node.js package,
-that package will be bundled with the journey code on `push`.
-However, you can't `push` bundles that are more than 800 Kilobytes.
+If the journey contains external NPM packages other than the `@elastic/synthetics`,
+those packages will be bundled along with the journey code when the `push` command is invoked.
+However there are some limitations when using external packages:
+	
+* Bundled journeys after compression should not be more than 800 Kilobytes.
+* Native node modules will not work as expected.
+  There will be platform inconsistency.
 ====
 
 `--auth`::
--- a/docs/en/observability/synthetics-command-reference.asciidoc
+++ b/docs/en/observability/synthetics-command-reference.asciidoc
@@ -92,6 +92,13 @@ Select `yes` to delete all monitors associated with the project ID being pushed.
 it using a _different_ ID. Select `yes` to create duplicates of all monitors in the project.
 ====
 
+[NOTE]
+====
+If the journey code in a project uses an external Node.js package,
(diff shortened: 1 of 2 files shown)


docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -92,6 +92,13 @@ Select `yes` to delete all monitors associated with the project ID being pushed.
 it using a _different_ ID. Select `yes` to create duplicates of all monitors in the project.
 ====
 
+[NOTE]
+====
+If the journey code in a project uses an external Node.js package,
+that package will be bundled with the journey code on `push`.
+However, you can't `push` bundles that are more than 800 Kilobytes.
+====
```
colleenmcginnis, line 100: I wasn't sure if I should also mention that "External packages that use native bindings will not work" in the command reference. Will it prevent the user from pushing the project or can they push it, but it won't work as expected?

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -92,6 +92,13 @@ Select `yes` to delete all monitors associated with the project ID being pushed.
 it using a _different_ ID. Select `yes` to create duplicates of all monitors in the project.
 ====
 
+[NOTE]
+====
+If the journey code in a project uses an external Node.js package,
+that package will be bundled with the journey code on `push`.
+However, you can't `push` bundles that are more than 800 Kilobytes.
```
vigneshshanmugam, line 99:
```suggestion
If the journey contains external NPM packages other than the `@elastic/synthetics`, those packages will be bundled along with the journey code when `push` command is invoked. However there are two limitations when using external packages

1. Bundled journey after compression should not be more than 800 Kilobytes.
2. Native node modules will not work as expected(platform inconsistency). 
```

vigneshshanmugam: Would replicate the same in the create test reference as well.

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -92,6 +92,17 @@ Select `yes` to delete all monitors associated with the project ID being pushed.
...
 
+[NOTE]
+====
+If the journey contains external NPM packages other than the `@elastic/synthetics`,
+those packages will be bundled along with the journey code when the `push` command is invoked.
+However there are some limitations when using external packages:
+	
+* Bundled journeys after compression should not be more than 800 Kilobytes.
+* Native node modules will not work as expected.
+  There will be platform inconsistency.
```
vigneshshanmugam, line 103:
```suggestion
* Native node modules will not work as expected due to platform inconsistency. 
```

vigneshshanmugam (approved):
LGTM

PR #2293, opened 2022-10-21
Clarify that Fleet Server certs are managed by cloud

### LINKED ISSUE #1864: TLS/SSL documentation about ESS fleet-servers in public docs
### Summary

Add information about how SSL/TLS certificates are managed in ESSs' Fleet server. An example addition would be:

"In ESS, the TLS certificates are managed by Elastic Cloud. There is no explicit need of creating a new Private Key and certificates. Any elastic agent just needs to have the truststore configured on the client side which will be connecting to Elastic Cloud's fleet servers".

### Details

- The documentation on self-managed fleet was clearly mentioned but, it seems like we missed upon the ESS managed Fleet documentation on the TLS/SSL part.
- This needs to be covered as many of them will be on boarded soon on ESS and will have same questions. Especially, customers who are used to configuring the SSL/TLS certificates for our product. 

### Related

* https://github.com/elastic/observability-docs/issues/1857

dedemorton:
Closes #1864

### DIFF
--- a/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
@@ -37,8 +37,7 @@ For more information about hosting {fleet-server} on {ece}, refer to
 
 NOTE: The TLS certificates used to secure connections between {agent} and
 {fleet-server} are managed by {ecloud}. You do not need to create a private key
-or generate certificates. To connect to {fleet-server}, {agent}s need to have
-the truststore configured on the client side.
+or generate certificates.
 
 [discrete]
 [[add-fleet-server-cloud-set-up]]
--- a/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
@@ -35,6 +35,11 @@ cover `*.cname`, not `*.*.cname`). This enables us to provide the URL for
 For more information about hosting {fleet-server} on {ece}, refer to
 {ece-ref}/ece-manage-integrations-server.html[Manage your {integrations-server}].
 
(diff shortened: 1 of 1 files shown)


dedemorton:
still looking for someone to review this please

docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
```diff
@@ -35,6 +35,11 @@ cover `*.cname`, not `*.*.cname`). This enables us to provide the URL for
 For more information about hosting {fleet-server} on {ece}, refer to
 {ece-ref}/ece-manage-integrations-server.html[Manage your {integrations-server}].
 
+NOTE: The TLS certificates used to secure connections between {agent} and
+{fleet-server} are managed by {ecloud}. You do not need to create a private key
+or generate certificates. To connect to {fleet-server}, {agent}s need to have
+the truststore configured on the client side.
```
joshdover, line 41: Do we need to mention the truststore part? From my understanding it's on by default, and I'm not even sure we have a way to disable it for Fleet-managed agents.

dedemorton: If that's the case, we should remove this sentence. I've had Elastic Agent running on my machine for so long I honestly can't remember if I had to do anything in the beginning to configure the truststore. :-) If users need to do anything, we should expand this description because it begs the question: how do I do that? let me know how to proceed...thanks!

dedemorton: @joshdover Do you want me to remove the last sentence here?

joshdover: Let's remove this. I don't think it's relevant for Fleet-managed Agents.

joshdover (approved):
Good after truststore part is removed.

PR #2298, opened 2022-10-26
Update package registry distributions according to Package Storage V2

mrodm:
Update references about the package registry distributions available once Package Storage V2 is used.

Snapshot and staging distributions (docker images) will no longer be updated. And, there will be a new distribution named "lite" with a small subset of commonly used packages.

### DIFF
--- a/docs/en/ingest-management/fleet/air-gapped.asciidoc
+++ b/docs/en/ingest-management/fleet/air-gapped.asciidoc
@@ -40,10 +40,10 @@ a selection of packages.
 There are different distributions available:
 
 * {version} (recommended): +docker.elastic.co/package-registry/distribution:{version}+ - Selection of packages from the production repository released with the {version} version of the {stack}.
-* lite {version}: +docker.elastic.co/package-registry/distribution:lite-{version}+ - Subset of the most commonly used packages from the production repository released with the {version} version of the {stack}.
+* lite-{version}: +docker.elastic.co/package-registry/distribution:lite-{version}+ - Subset of the most commonly used packages from the production repository released with the {version} version of the {stack}. This image includes the commonly used packages that are needed or when running Fleet in testing air-gapped environments.
 * production: `docker.elastic.co/package-registry/distribution:production` - Packages available in the production registry (https://epr.elastic.co).
-* lite: `docker.elastic.co/package-registry/distribution:lite` - Small subset of packages available in the production registry (https://epr.elastic.co).
-This image includes only the most commonly used packages, to be used when just common packages are needed or when trying Fleet in air-gapped environments.
+* lite: `docker.elastic.co/package-registry/distribution:lite` - Subset of the most commonly used packages available in the production registry (https://epr.elastic.co).
+This image includes the commonly used packages that are needed or when running Fleet in testing air-gapped environments.
 
 ifeval::["{release-state}"=="unreleased"]
 [WARNING]
--- a/docs/en/ingest-management/fleet/air-gapped.asciidoc
+++ b/docs/en/ingest-management/fleet/air-gapped.asciidoc
@@ -41,8 +41,7 @@ There are different distributions available:
 
 * {version} (recommended): +docker.elastic.co/package-registry/distribution:{version}+ - Selection of packages from the production repository released with the {version} version of the {stack}.
 * production: `docker.elastic.co/package-registry/distribution:production` - Packages available in the production registry (https://epr.elastic.co).
-* staging: `docker.elastic.co/package-registry/distribution:staging` - Packages available in the staging registry (https://epr-staging.elastic.co). These packages may be pending of validation.
-* snapshot: `docker.elastic.co/package-registry/distribution:snapshot` - Packages under development.
+* lite: `docker.elastic.co/package-registry/distribution:lite` - Small subset of packages available in the production registry (https://epr.elastic.co).
 
 ifeval::["{release-state}"=="unreleased"]
 [WARNING]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -41,8 +41,7 @@ There are different distributions available:
 
 * {version} (recommended): +docker.elastic.co/package-registry/distribution:{version}+ - Selection of packages from the production repository released with the {version} version of the {stack}.
 * production: `docker.elastic.co/package-registry/distribution:production` - Packages available in the production registry (https://epr.elastic.co).
-* staging: `docker.elastic.co/package-registry/distribution:staging` - Packages available in the staging registry (https://epr-staging.elastic.co). These packages may be pending of validation.
-* snapshot: `docker.elastic.co/package-registry/distribution:snapshot` - Packages under development.
+* lite: `docker.elastic.co/package-registry/distribution:lite` - Small subset of packages available in the production registry (https://epr.elastic.co).
```
jsoriano, line 44: Do we have versioned "lite" distributions?

jsoriano, line 44: Maybe we should add a note about when you would use this image.

jsoriano (approved):

docs/en/ingest-management/fleet/air-gapped.asciidoc
mrodm: I was thinking to also version that image to be offered along with the production one.
Something like: `docker.elastic.co/package-registry/distribution:lite-v8.x.y`

Related PR https://github.com/elastic/package-registry/pull/900

jsoriano (approved):

docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -40,9 +40,10 @@ a selection of packages.
 There are different distributions available:
 
 * {version} (recommended): +docker.elastic.co/package-registry/distribution:{version}+ - Selection of packages from the production repository released with the {version} version of the {stack}.
+* lite {version}: +docker.elastic.co/package-registry/distribution:lite-{version}+ - Subset of the most commonly used packages from the production repository released with the {version} version of the {stack}.
 * production: `docker.elastic.co/package-registry/distribution:production` - Packages available in the production registry (https://epr.elastic.co).
-* staging: `docker.elastic.co/package-registry/distribution:staging` - Packages available in the staging registry (https://epr-staging.elastic.co). These packages may be pending of validation.
-* snapshot: `docker.elastic.co/package-registry/distribution:snapshot` - Packages under development.
+* lite: `docker.elastic.co/package-registry/distribution:lite` - Small subset of packages available in the production registry (https://epr.elastic.co).
```
dedemorton, line 45: Shouldn't this description be consistent with the one that we provide for the versioned distro? I'm not sure we provide enough info here for users to decide between the versioned and non-versioned distro.

docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -40,9 +40,10 @@ a selection of packages.
...
+This image includes only the most commonly used packages, to be used when just common packages are needed or when trying Fleet in air-gapped environments.
```
dedemorton, line 46: Couple of thoughts: "to be used when just common packages are needed" seems a little vague, and I'm not sure I would say "common" if you mean "commonly used" because there's room for misinterpretation. Also I would change "trying" to something like "running" because "trying" makes it sound like it's difficult to do (which might be true, but let's stay positive in the docs. 😄

docs/en/ingest-management/fleet/air-gapped.asciidoc
mrodm: I'll update both definitions to be consistent

mrodm: About the difference between versioned and non-versioned distributions, it should be like in production. The packages that can be found in the versioned one are the ones that were  available at the moment of the Elastic stack X.Y release (e.g. 8.5.0). And the non-versioned contain the latest packages published.

Could we add something else to differentiate more those two distributions ?

jsoriano (approved):

docs/en/ingest-management/fleet/air-gapped.asciidoc
```diff
@@ -40,9 +40,9 @@ a selection of packages.
 There are different distributions available:
 
 * {version} (recommended): +docker.elastic.co/package-registry/distribution:{version}+ - Selection of packages from the production repository released with the {version} version of the {stack}.
+* lite-{version}: +docker.elastic.co/package-registry/distribution:lite-{version}+ - Subset of the most commonly used packages from the production repository released with the {version} version of the {stack}. This image is a good candidate to start using Fleet on air-gapped environments.
```
dedemorton, line 43:
```suggestion
* lite-{version}: +docker.elastic.co/package-registry/distribution:lite-{version}+ - Subset of the most commonly used packages from the production repository released with {stack} {version}. This image is a good candidate to start using {fleet} in air-gapped environments.
```

dedemorton (commented):
Minor edits, otherwise LGTM.

docs/en/ingest-management/fleet/air-gapped.asciidoc
mrodm: @dedemorton  Should I update also the item related to production versioned (line 42?) with this change ?

```
* {version} (recommended): +docker.elastic.co/package-registry/distribution:{version}+ - Selection of packages from the production repository released with {stack} {version}.
```

dedemorton: Yes please!

dedemorton (approved):
LGTM

PR #2315, opened 2022-11-03
adjust for removed kibana security settings

robb3rt:
The setting `xpack.security.enabled` [has been removed since 8.0.0](https://github.com/elastic/kibana/pull/111681) so following these instructions will result in errors from the Kibana side. This file needs to be adjusted for versions of 8.0.0 and upwards


dedemorton (approved):
LGTM. Thank you for fixing this! I'll add the necessary backport labels and make sure it gets backported to all branches 8.0+.

PR #2335, opened 2022-11-07
Add GKE docs section

ChrsMark:
Add documentation page for GKE environments.

Part of https://github.com/elastic/observability-dev/issues/2408.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
@@ -14,7 +14,7 @@ To learn how to run {agent}s in a containerized environment, see:
 
 * <<running-on-kubernetes-standalone>>
 
-* <<running-on-gke-standard-managed-by-fleet>>
+** <<running-on-gke-standard-managed-by-fleet>>
 
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
 
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
@@ -14,5 +14,7 @@ To learn how to run {agent}s in a containerized environment, see:
 
 * <<running-on-kubernetes-standalone>>
 
+* <<running-on-gke-managed-by-fleet>>
+
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
 
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
@@ -14,7 +14,7 @@ To learn how to run {agent}s in a containerized environment, see:
 
 * <<running-on-kubernetes-standalone>>
 
-* <<running-on-gke-managed-by-fleet>>
+* <<running-on-gke-standard-managed-by-fleet>>
 
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
 
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -39,6 +39,7 @@ Refer to:
 * <<running-on-kubernetes-managed-by-fleet>>
 * <<running-on-kubernetes-standalone>>
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
+* <<running-on-kubernetes-gke-standard-managed-by-fleet>>
 --
 
 [discrete]
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -40,7 +40,7 @@ Refer to:
 * <<running-on-kubernetes-standalone>>
 * <<running-on-gke-standard-managed-by-fleet>>
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
-* <<running-on-kubernetes-gke-standard-managed-by-fleet>>
+* <<running-on-gke-standard-managed-by-fleet>>
 --
 
 [discrete]
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -38,6 +38,7 @@ Refer to:
 * <<elastic-agent-container>>
 * <<running-on-kubernetes-managed-by-fleet>>
 * <<running-on-kubernetes-standalone>>
+* <<running-on-gke-managed-by-fleet>>
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
 * <<running-on-kubernetes-gke-standard-managed-by-fleet>>
 --
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -38,7 +38,6 @@ Refer to:
 * <<elastic-agent-container>>
 * <<running-on-kubernetes-managed-by-fleet>>
 * <<running-on-kubernetes-standalone>>
-* <<running-on-gke-standard-managed-by-fleet>>
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
 * <<running-on-gke-standard-managed-by-fleet>>
 --
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -38,7 +38,7 @@ Refer to:
 * <<elastic-agent-container>>
 * <<running-on-kubernetes-managed-by-fleet>>
 * <<running-on-kubernetes-standalone>>
-* <<running-on-gke-managed-by-fleet>>
+* <<running-on-gke-standard-managed-by-fleet>>
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
 * <<running-on-kubernetes-gke-standard-managed-by-fleet>>
 --
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -30,7 +30,7 @@ at the time of writing these docs.
 [discrete]
 == Prerequisites
 
-Kube-state-metrics is not deployed by default in Kubernetes GKE!
+`kube-state-metrics` is not deployed by default in Kubernetes GKE!
 
 Data_streams with the `state_` prefix require `kube-state-metrics` to be present.
 
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -37,7 +37,7 @@ Data_streams with the `state_` prefix require `kube-state-metrics` to be present
 In order to install kube-state-metrics follow the instructions for its deployment are available
 https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
 Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
-`kube-system` namespace, which will be the service to use in our configuration of agent
+`kube-system` namespace, which will be the service to use in agent configuration
 service within the cluster.
 
 [discrete]
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -51,7 +51,7 @@ the https://docs.elastic.co/en/integrations/kubernetes[Kubernetes integration].
 
 
 [discrete]
-== Enroll Elastic Agent to Fleet
+== 2. Enroll Elastic Agent to Fleet
 
 With {fleet}, each agent enrolls in a policy defined in {kib} and stored in
 {es}. The policy specifies how to collect observability data from the services
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -102,7 +102,7 @@ Refer to <<agent-environment-variables>> for all available options.
 ====
 
 [discrete]
-== Deploy kubernetes manifests
+== 3. Deploy kubernetes manifests
 
 On Kubernetes, we suggest deploying {agent} as a https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/[DaemonSet]
 to ensure that there is a running instance on each node of the cluster.
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -37,8 +37,7 @@ Data_streams with the `state_` prefix require `kube-state-metrics` to be present
 In order to install `kube-state-metrics` follow the instructions for its deployment, that are available
 https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
 Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
-`kube-system` namespace, which will be the service to use in agent configuration
-service within the cluster.
+`kube-system` namespace, which will be the service to use in agent configuration.
 
 [discrete]
 == 1. Configure a Fleet policy
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -1,4 +1,4 @@
-[[running-on-kubernetes-gke-standard-managed-by-fleet]]
+[[running-on-gke-standard-managed-by-fleet]]
 = Run {agent} on GKE standard managed by {fleet}
 
 Use {agent} https://www.docker.elastic.co/r/beats/elastic-agent[Docker images] on Kubernetes to
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -41,7 +41,7 @@ Generally `kube-state-metrics` runs a `Deployment` and is accessible via a servi
 service within the cluster.
 
 [discrete]
-== Configure a Fleet policy
+== 1. Configure a Fleet policy
 
 In order for the Agents to enable the proper inputs, they need to be enrolled to the proper policy.
 For achieving Kubernetes observability, one need to create a policy and enable the Kubernetes integration.
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -34,7 +34,7 @@ at the time of writing these docs.
 
 Data_streams with the `state_` prefix require `kube-state-metrics` to be present.
 
-In order to install kube-state-metrics follow the instructions for its deployment are available
+In order to install `kube-state-metrics` follow the instructions for its deployment, that are available
 https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
 Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
 `kube-system` namespace, which will be the service to use in agent configuration
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -16,7 +16,7 @@ endif::[]
 [discrete]
 == Important Notes:
 
-On managed Kubernetes solutions like GKE, {agent} has no access several data sources. Find below the list of
+On managed Kubernetes solutions like GKE, {agent} has no access to several data sources. Find below the list of
 the non available data:
 
 1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -24,7 +24,7 @@ components are not available. Consequently metrics are not available for `kube-s
 In this regard, the respective **dashboards** will not be populated with data.
 2. **Audit logs** are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
 
-Also note that the documentation page only applies to Standard GKE and not Autopilot. Autopilot support is a work in progress
+Also note that the documentation page only applies to Standard GKE and not Autopilot. **Autopilot support** is a work in progress
 at the time of writing these docs.
 
 [discrete]
--- a/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
@@ -20,7 +20,8 @@ On managed Kubernetes solutions like GKE, {agent} has no access to several data 
 the non available data:
 
 1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
-components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`.
+components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`
+components.
 In this regard, the respective **dashboards** will not be populated with data.
 2. **Audit logs** are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
 
--- a/docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
@@ -14,23 +14,31 @@ released, so no Docker image is currently available for this version.
 endif::[]
 
 [discrete]
+== Important Notes:
+
+On managed Kubernetes solutions like GKE, {agent} has no access several data sources. Find below the list of
+the non available data:
+
+1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
+components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`.
+In this regard, the respective dashboards will not be populated with data.
+2. Audit logs are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
+
+Also note that the documentation page only applies to Standard GKE and not Autopilot. Autopilot support is a work in progress
+at the time of writing these docs.
+
+[discrete]
 == Prerequisites
 
 Kube-state-metrics is not deployed by default in Kubernetes GKE!
 
-Data_streams with the `state_` prefix require ` `kube-state-metrics` library to be present
+Data_streams with the `state_` prefix require `kube-state-metrics` to be present.
 
-. In order to install kube-state-metrics follow
-the instructions for its deployment are available https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
+In order to install kube-state-metrics follow the instructions for its deployment are available
+https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
 Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
 `kube-system` namespace, which will be the service to use in our configuration of agent
 service within the cluster.
-
-`Kube-state-metrics` is not deployed by default in Kubernetes GKE. In order to install `kube-state-metrics` follow
-the instructions for its deployment https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
-Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
-`kube-system` namespace, which will be the service to use in our configuration.
-
 
 [discrete]
 == Kubernetes deploy manifests
@@ -41,9 +49,9 @@ to be monitored. The {agent} connects to a trusted {fleet-server} instance
 to retrieve the policy and report agent events.
 
 We recommend using {fleet} management because it makes the management and
-upgrade of agents considerably easier.
+upgrade of agents' integrations considerably easier.
 
-On Kubernetes, deploy {agent} as a https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/[DaemonSet]
+On Kubernetes, we suggest deploying {agent} as a https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/[DaemonSet]
 to ensure that there is a running instance on each node of the cluster.
 These instances are used to retrieve metrics from the host, such as system metrics, container stats,
 and metrics from all the services running on top of Kubernetes.
@@ -67,7 +75,17 @@ curl -L -O https://raw.githubusercontent.com/elastic/elastic-agent/{branch}/depl
 ------------------------------------------------
 
 [discrete]
-== Settings
+== Configure a Fleet policy
+
+In order for the Agents to enable the proper inputs, they need to be enrolled to the proper policy.
+For achieving Kubernetes observability, one need to create a policy and enable the Kubernetes integration.
+
+Refer to <<create-a-policy>> to learn how to enable
+the https://docs.elastic.co/en/integrations/kubernetes[Kubernetes integration].
+
+
+[discrete]
+== Enroll Elastic Agent to Fleet
 
 {agent} is enrolled to a running {fleet-server} using `FLEET_URL` parameter.
 The `FLEET_ENROLLMENT_TOKEN` parameter is used to connect {agent} to a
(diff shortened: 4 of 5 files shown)


ChrsMark:
@gizas @mlunadia can you have a look please and let me know what you think?

My main concern here is that we replicate a lot of content which will end up being extra maintenance load for us in the future. However I don't think we have an efficient way to abstract some parts so maintaining the pages individually seems to be the way to go.

dedemorton:
>My main concern here is that we replicate a lot of content which will end up being extra maintenance load for us in the future.

I haven't done a side-by-side comparison to see how much duplication is here, but asciidoc *does* support [tagged regions](https://docs.asciidoctor.org/asciidoc/latest/directives/include-tagged-regions/#tagging-regions) that allow you to pull in chunks of content.

Extensive use of tagged regions can make the source files harder to read and update, but it might be worth the effort if you're duplicating a lot of content.

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+- name: KIBANA_HOST
+  value: ""
+- name: KIBANA_FLEET_USERNAME
+  value: ""
+- name: KIBANA_FLEET_PASSWORD
+  value: ""
+------------------------------------------------
+<1> To learn how to create a policy, refer <<create-a-policy-no-ui>>.
+
+
```
dedemorton, line 90: You need to set the `type` attribute somewhere before these includes because the IDs use `type` as a variable to ensure uniqueness in case the setting is included in multiple places.

`:type: somevalue`

ChrsMark:
Thanks @dedemorton for the suggestion!
Thinking of this again and checking the content I believe is better to go with separated sections without importing common parts. We can revisit this in the future if we see it brings issues. 

In addition, I would combine maintaining these pages part of a smoke test procedure in the end of each release so as to verify and ensure that all providers are covered with the latest changes.

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+endif::[]
+
+[discrete]
+== Prerequisites
+
+All data_streams with the `state_` prefix require `hosts` field pointing to `kube-state-metrics`
+service within the cluster.
+
+Kube-state-metrics is not deployed by default in Kubernetes GKE. In order to install kube-state-metrics follow
+the instructions for its deployment are available https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
```
tetianakravchenko, line 23:
```suggestion
the instructions for its deployment https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+However, version {version} of {agent} has not yet been
+released, so no Docker image is currently available for this version.
+
+endif::[]
+
+[discrete]
+== Prerequisites
+
+All data_streams with the `state_` prefix require `hosts` field pointing to `kube-state-metrics`
```
tetianakravchenko, line 19: I think that at this point of the documentation it might be not yet clear about which data_streams is it going, and what is a `hosts`.
maybe it could be better to just say what is required and keep rest as an clarification, smth like:
```
- `kube-state-metrics` is running and accessible within the cluster.

`kube-state-metrics` is not deployed by default in Kubernetes GKE. In order to install `kube-state-metrics` follow
the instructions for its deployment https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+Refer to <<agent-environment-variables>> for all available options.
+
+****
+====
+
+
+[discrete]
+== Deploy
+
+If planning to deploy `state_*` datasets of Kubernetes package,
```
tetianakravchenko, line 123: I think it can be removed as it was already mentioned. Maybe we should adjust origin doc page as well

tetianakravchenko (commented):
Should be mentioned `Enable explicitly "Collect Kubernetes metrics from kube-state-metrics"` and provide the kube-state-metrics service as a host?
Or we assume that we enable it by default, and no need to mention it here?

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+
+Everything is deployed under the `kube-system` namespace by default. To change
+the namespace, modify the manifest file.
+
+To download the manifest file, run:
+
+["source", "sh", subs="attributes"]
+------------------------------------------------
+curl -L -O https://raw.githubusercontent.com/elastic/elastic-agent/{branch}/deploy/kubernetes/elastic-agent-managed-kubernetes.yaml
```
tetianakravchenko, line 59: Maybe instead of adding this command and explaining Settings, we should mention to copy the manifest from the UI?

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
+[[running-on-kubernetes-gke-standard-managed-by-fleet]]
+= Run {agent} on GKE standard managed by {fleet}
```
tetianakravchenko, line 2: should this documentation be more an `Observe GKE standard with elastic-agent managed by fleet`, then the run fleet?

gizas: So also the title should not have kubenretes word. I think is duplication
[running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc] -> [running-on-gke-standard-managed-by-fleet.asciidoc]

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
gizas, line 19:
```suggestion
Kube-state-metrics is not deployed by default in Kubernetes GKE!

Data_streams with the `state_` prefix require ` `kube-state-metrics` library to be present

. In order to install kube-state-metrics follow
the instructions for its deployment are available https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
`kube-system` namespace, which will be the service to use in our configuration of agent
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+[discrete]
+== Prerequisites
+
+All data_streams with the `state_` prefix require `hosts` field pointing to `kube-state-metrics`
+service within the cluster.
+
+Kube-state-metrics is not deployed by default in Kubernetes GKE. In order to install kube-state-metrics follow
+the instructions for its deployment are available https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
+Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
+`kube-system` namespace, which will be the service to use in our configuration.
```
gizas, line 25: [configuration]
Maybe we can point to line of manifest where kube-state metrics service is ?

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+
+[discrete]
+== Kubernetes deploy manifests
```
gizas, line 29: I think all this section can be removed and go to elastic-agent doc. So to keep it common for all 

In here I would like something like list of steps until someone can have GKE monitored with our solution

1. Login to GKE and and retrieve credentials for specific GKE cluster 
<gcloud .....> example command
2. Verify your cluster locally by running kubectl command 
3. Login to kibana and go to specific menu and 
3a. Install Integration
3b Create Agent Policy
3c Bind Policy to Agent
4. Apply  manifest to your cluster
5. Verify that your agent is running --> Can we show logs of the agent pod?
6. Verify that your policy is applied  --> Any indication there? Probably from UI
7. Verify logs/ metrics in your kibana discovery 
7a. Explain where logs come from and where metrics come from and should be under which data-view

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+To download the manifest file, run:
+
+["source", "sh", subs="attributes"]
+------------------------------------------------
+curl -L -O https://raw.githubusercontent.com/elastic/elastic-agent/{branch}/deploy/kubernetes/elastic-agent-managed-kubernetes.yaml
+------------------------------------------------
+
+[discrete]
+== Settings
```
gizas, line 63: This section can go on the steps above where we install agent.

Can you make a comment for FLEET_ENROLLMENT_TOKEN vs FLEET_ENROLLMENT_USERNAME/PASSWORD?

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+The size and the number of nodes in a Kubernetes cluster can be fairly large at times,
+and in such cases the Pod that will be collecting cluster level metrics might face performance
+issues due to resources limitations. In this case users might consider to avoid using the
+leader election strategy and instead run a dedicated, standalone {agent} instance using
+a Deployment in addition to the DaemonSet.
+
+[discrete]
+== Not available data
+
+On managed Kubernetes solutions like GKE, {agent} has no access to collect metrics from the https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
```
gizas, line 167: Make those bullets to be different to the eye

I am wondering if we should move this section first after Prerequisities

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+components, like `kube-scheduler` and `kube-controller-manager`, that are scheduled on Kubernetes master nodes.
+
+Audit logs are available only on Kubernetes master nodes as well and hence cannot be collected by {agent}.
```
gizas, line 170:
```suggestion
Audit logs are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+The size and the number of nodes in a Kubernetes cluster can be fairly large at times,
+and in such cases the Pod that will be collecting cluster level metrics might face performance
+issues due to resources limitations. In this case users might consider to avoid using the
+leader election strategy and instead run a dedicated, standalone {agent} instance using
+a Deployment in addition to the DaemonSet.
+
+[discrete]
+== Not available data
+
```
gizas, line 166: Can we add here or to another section details for the orchestratot.cloud.name with add_field ?
Just an example oof the processor with one liner to explain what problem it solves

Also for autopilot: GKE autopilot installations do now allow agent installation. At the moment several changes are needed to the provided manifest in order to be installed. Do we plan to have another version of such manifest online?

ChrsMark: This PR is only for GKE standard. Autopilot will be covered separately once we have all the details figured out regarding the requirements etc.

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
ChrsMark: Only `FLEET_URL` and `FLEET_ENROLLMENT_TOKEN` are required. I will remove the rest of them. FYI those are just copied from the original doc page so we need to revisit this one too :)

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
ChrsMark: This steps are already part of the original "Run Agent on k8s" documentation page. Is this what you mean by "elastic-agent doc"?

To my mind I wouldn't remove the steps and redirect to another basic page since here we need a page that would be just enough for the GKE users without the need to jump to other pages and combine information. With this the experience is more simple and robust -> "I want to run Agent on GKE I follow the whole page and that works".

Also I don't like to have different verification steps for GKE compared to basic k8s.
Let me know what you think.

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
ChrsMark: Not sure about this. Are we planing to do the same for other providers too? And what about the basic "running on k8s" guide?

ChrsMark:
@gizas @tetianakravchenko thanks for your comments here! Reading those carefully I realize that we lack a common understanding of what these docs should be about. Maybe I have taken this to the wrong direction but let me try to explain my perspective here.

First of all we decided to have a different page for GKE users so as to make it easier for them to navigate inside our docs and ensure that each GKE specific detail is handled and mentioned explicitly. So in this regard this doc page should follow the same approach as the [basic one](https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc) does with only difference that makes it GKE (standard) specific. Note that I mention GKE standard and not Autopilot cause this should require another additional page which will be added in the future once we verify the details.

So to my mind this page should not redirect to other pages and should be a standalone one so as the users to be able to achieve the installation by just following steps of this page.

Also the page should not follow a different approach or have a different goal compared to https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc. This means that the steps should be quite similar as well as the verification process. I would like to diverge from the original one. Of course if we want to revisit the original page we can do by the baseline of the flow of these docs should be quite similar as much as possible. This would also help us maintaining those with less effort.

As you can see I tried to adhere by this approach and I completely followed the flow of the basic page (https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc). 

My inspiration comes from projects that do something similar like https://istio.io/latest/docs/setup/platform-setup/ or https://docs.cilium.io/en/v1.9/gettingstarted/k8s-install-managed/ and https://docs.cilium.io/en/v1.9/gettingstarted/k8s-install-sandbox/.

Would that make sense? Let me know what are your thoughts here.

tetianakravchenko:
> @gizas @tetianakravchenko thanks for your comments here! Reading those carefully I realize that we lack a common understanding of what these docs should be about. Maybe I have taken this to the wrong direction but let me try to explain my perspective here.

> First of all we decided to have a different page for GKE users so as to make it easier for them to navigate inside our docs and ensure that each GKE specific detail is handled and mentioned explicitly. So in this regard this doc page should follow the same approach as the [basic one](https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc) does with only difference that makes it GKE (standard) specific. Note that I mention GKE standard and not Autopilot cause this should require another additional page which will be added in the future once we verify the details.

@ChrsMark I think I understand your point and it totally makes sense.
If we think about those documentation pages (run agent on GKE standard/EKS,etc) as a "installation guide" - they should be almost the same, just changing name of the cloud provider (for Autopilot it will be completely different). From this point, I agree it should follow the same approach as [origin page](https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc). 

From my understanding what we want to highlight that to observe kubernetes cluster (with help of kubenetes integration package) we require kube_state_metrics installed (it is not required just to run an agent, right?), additionally for some cloud providers should be added `add_fields` processor, in order to make kubernetes dashboards working, audit logs are not available - this information is only relevant for the observing k8s cluster.
@ChrsMark does it makes sense?
 
@gizas @mlunadia any thoughts on this discussion?

ChrsMark:
> From my understanding what we want to highlight that to observe kubernetes cluster (with help of kubenetes integration package) we require kube_state_metrics installed (it is not required just to run an agent, right?), additionally for some cloud providers should be added `add_fields` processor, in order to make kubernetes dashboards working, audit logs are not available - this information is only relevant for the observing k8s cluster. @ChrsMark does it make sense?

To my mind it should be "Install on k8s and monitor k8s" at the same context. We assume the same for the "Run standalone Agent on k8s" guide where we enable/define the k8s integration by default at https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml#L46.

The same was in place for Metricbeat too, where we had the modules defined in the k8s manifests.

For managed mode it should be the same. Since the mention of installing the k8s integration explicitly is missing from the https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc I suggest that we just add it there too when we explain about the policies and follow the same for the providers' docs.

Things like `add_fields` processor need etc totally fit to a provider specific page like the EKS one.

tetianakravchenko:
> For managed mode it should be the same. Since the mention of installing the k8s integration explicitly is missing from the https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc I suggest that we just add it there too when we explain about the policies and follow the same for the providers' docs.

so you would suggest to proceed with this PR, and open an issue to describe installation of k8s integration in origin page, including configuration and align provider specific pages accordingly, correct?

mlunadia:
Our world is going to keep moving further from a one-size-fits-all driven by Vanilla K8s into having to create CSP-specific user journeys.
This said we need to make it consistent as possible and find a mechanism that ties our CSP agnostic journey with each customised one. 
This will be reflected in the same way in docs. 
@ChrsMark how can I access a preview of the proposed docs for GKE end to end?

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,177 @@
...
+Data_streams with the `state_` prefix require ` `kube-state-metrics` library to be present
+
+. In order to install kube-state-metrics follow
+the instructions for its deployment are available https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
+Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
+`kube-system` namespace, which will be the service to use in our configuration of agent
+service within the cluster.
+
+`Kube-state-metrics` is not deployed by default in Kubernetes GKE. In order to install `kube-state-metrics` follow
```
tetianakravchenko, line 29: info is duplicated here, could you remove it?

ChrsMark:
@mlunadia in general you can see a preview of the docs through the links from https://github.com/elastic/observability-docs/pull/2335#issuecomment-1305257520. However those seem to not work now. 

@dedemorton do we have a way to rebuild the preview docs pages?

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
gizas: Makes sense. Then lets update cetral main page

ChrsMark:
We had an offline chat with @gizas and concluded to move this forward with a desired format which will be as complete as possible. This means that we will need to revisit the https://github.com/elastic/observability-docs/blob/main/docs/en/ingest-management/elastic-agent/running-on-kubernetes-managed-by-fleet.asciidoc docs to adjust them accordingly to follow the flow we introduce here.

See the current state of this PR after [dfdf3b8](https://github.com/elastic/observability-docs/pull/2335/commits/dfdf3b839c8448cf48bfca16ac2a3dcb3e5bfe68) for the proposed structure of the docs.

If we conclude on this we can follow the same in other providers too, but @gizas @tetianakravchenko please have another look  to the updated format and let me know what you think :).

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+[discrete]
+== Important Notes:
+
+On managed Kubernetes solutions like GKE, {agent} has no access several data sources. Find below the list of
+the non available data:
+
+1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
+components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`.
+In this regard, the respective dashboards will not be populated with data.
+2. Audit logs are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
```
gizas, line 25:
```suggestion
2. **Audit logs** are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+[discrete]
+== Important Notes:
+
+On managed Kubernetes solutions like GKE, {agent} has no access several data sources. Find below the list of
+the non available data:
+
+1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
+components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`.
+In this regard, the respective dashboards will not be populated with data.
```
gizas, line 24:
```suggestion
In this regard, the respective **dashboards** will not be populated with data.
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+In addition, one of the Pods in the DaemonSet will constantly hold a _leader lock_ which makes it responsible for
+handling cluster-wide monitoring.
+Find more information about leader election configuration options at <<kubernetes_leaderelection-provider, leader election provider>>.
+This instance is used to retrieve metrics that are unique for the whole
+cluster, such as Kubernetes events or
+https://github.com/kubernetes/kube-state-metrics[kube-state-metrics].
+
+
+Everything is deployed under the `kube-system` namespace by default. To change
```
gizas, line 67:
```suggestion
Default namespace of deployment is `kube-system` . To change
```

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+[discrete]
+== Kubernetes deploy manifests
+
+With {fleet}, each agent enrolls in a policy defined in {kib} and stored in
+{es}. The policy specifies how to collect observability data from the services
+to be monitored. The {agent} connects to a trusted {fleet-server} instance
+to retrieve the policy and report agent events.
+
+We recommend using {fleet} management because it makes the management and
+upgrade of agents' integrations considerably easier.
```
gizas, line 52: >With {fleet}, each agent enrolls in a policy defined in {kib} and stored in
{es}. The policy specifies how to collect observability data from the services
to be monitored. The {agent} connects to a trusted {fleet-server} instance
to retrieve the policy and report agent events.
We recommend using {fleet} management because it makes the management and
upgrade of agents' integrations considerably easier.

I think all this part should go first to the section `Configure a Fleet Policy`

gizas: Also the rest section of `Kubernetes deploy manifests` should go and be merged with the `Deploy` section

What do you think?

gizas: Also last comment on this:
Installation Steps can be the [discrete] header here and under that to have:
1. Configure a Fleet policy
2. Enroll Elastic Agent to Fleet
3. Deploy kubernetes manifests

Those I dont think should be discrete and I would put them as sub headers with the numbers as well. To so that are steps in a sequence

gsantoro:
hey @ChrsMark , should I create a similar doc for AKS? I wasn't aware that we were writing cloud provider docs until @tetianakravchenko mentioned this morning in a chat. I might have missed the memo

ChrsMark:
> hey @ChrsMark , should I create a similar doc for AKS? I wasn't aware that we were writing cloud provider docs until @tetianakravchenko mentioned this morning in a chat. I might have missed the memo

Hey! This is part of https://github.com/elastic/observability-dev/issues/2408 and most probably applies to all the relevant cloud providers' tasks. I guess its up to everyone of us to decide when is the best time to write this docs page, but I think we have now decided to the form of these pages so you could follow the same format. I would suggest verifying this with @gizas too :).

dedemorton:
@ChrsMark 
>@dedemorton do we have a way to rebuild the preview docs pages?

The preview is rebuilt every time you push a new commit. But if you want to force a build, just add a comment with:  `@elasticmachine run elasticsearch-ci/doc`.

dedemorton:
To get the preview working, you'll have to fix these errors by ensuring that these IDs are unique: 

```
06:41:50 INFO:build_docs:asciidoctor: WARNING: elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc: line 118: id assigned to block already in use: env-{type}-fleet-url
06:41:50 INFO:build_docs:asciidoctor: WARNING: elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc: line 128: id assigned to block already in use: env-{type}-fleet-enrollment-token
```

Don't have time to look right now to figure out the problem, but keep in mind that the value of `:type:` remains in scope for the entire book unless it's set to something else or cleared.

docs/en/ingest-management/elastic-agent/running-on-kubernetes-gke-standard-managed-by-fleet.asciidoc
dedemorton: I like the idea of making the titles more user goal oriented. I wonder what users are likely to search for when they go to google or our docs?

ChrsMark:
@mlunadia docs are now available for preview at https://github.com/elastic/observability-docs/pull/2335#issuecomment-1305257520.

@gizas also the PR is ready for another round of review :).

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -38,7 +38,9 @@ Refer to:
 * <<elastic-agent-container>>
 * <<running-on-kubernetes-managed-by-fleet>>
 * <<running-on-kubernetes-standalone>>
+* <<running-on-gke-standard-managed-by-fleet>>
 * {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] -- for {eck} users
+* <<running-on-gke-standard-managed-by-fleet>>
```
tetianakravchenko, line 43: this line is mentioned twice in this list (line 41 & 43), is it expected?

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+
+However, version {version} of {agent} has not yet been
+released, so no Docker image is currently available for this version.
+
+endif::[]
+
+[discrete]
+== Important Notes:
+
+On managed Kubernetes solutions like GKE, {agent} has no access several data sources. Find below the list of
```
tetianakravchenko, line 19:
```suggestion
On managed Kubernetes solutions like GKE, {agent} has no access to several data sources. Find below the list of
```

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+the non available data:
+
+1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
+components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`.
```
tetianakravchenko, line 23:
```suggestion
components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager` datasets.
```
?

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+at the time of writing these docs.
+
+[discrete]
+== Prerequisites
+
+Kube-state-metrics is not deployed by default in Kubernetes GKE!
+
+Data_streams with the `state_` prefix require `kube-state-metrics` to be present.
+
+In order to install kube-state-metrics follow the instructions for its deployment are available
```
tetianakravchenko, line 37:
```suggestion
In order to install `kube-state-metrics` follow the instructions for its deployment, that are available
```
? or just remove `are available`

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+In this regard, the respective **dashboards** will not be populated with data.
+2. **Audit logs** are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
+
+Also note that the documentation page only applies to Standard GKE and not Autopilot. Autopilot support is a work in progress
+at the time of writing these docs.
+
+[discrete]
+== Prerequisites
+
+Kube-state-metrics is not deployed by default in Kubernetes GKE!
```
tetianakravchenko, line 33:
```suggestion
`kube-state-metrics` is not deployed by default in Kubernetes GKE!
```

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
ChrsMark: You are right... 🤦🏼‍♂️ thank you!

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
ChrsMark: Here I refer to the services not the data_streams.

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+
+On managed Kubernetes solutions like GKE, {agent} has no access to several data sources. Find below the list of
+the non available data:
+
+1. Metrics from https://kubernetes.io/docs/concepts/overview/components/#control-plane-components[Kubernetes control plane]
+components are not available. Consequently metrics are not available for `kube-scheduler` and `kube-controller-manager`.
+In this regard, the respective **dashboards** will not be populated with data.
+2. **Audit logs** are available only on Kubernetes master nodes as well, hence cannot be collected by {agent}.
+
+Also note that the documentation page only applies to Standard GKE and not Autopilot. Autopilot support is a work in progress
```
gizas, line 27:
```suggestion
Also note that the documentation page only applies to Standard GKE and not Autopilot. **Autopilot support** is a work in progress
```

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+Data_streams with the `state_` prefix require `kube-state-metrics` to be present.
+
+In order to install `kube-state-metrics` follow the instructions for its deployment, that are available
+https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
+Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
+`kube-system` namespace, which will be the service to use in agent configuration
+service within the cluster.
+
+[discrete]
+== Configure a Fleet policy
```
gizas, line 44:
```suggestion
== 1. Configure a Fleet policy
```

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+
+|===
+
+Refer to <<agent-environment-variables>> for all available options.
+
+****
+====
+
+[discrete]
+== Deploy kubernetes manifests
```
gizas, line 105:
```suggestion
== 3. Deploy kubernetes manifests
```

gizas (approved):

docs/en/ingest-management/elastic-agent/running-on-gke-standard-managed-by-fleet.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+
+`kube-state-metrics` is not deployed by default in Kubernetes GKE!
+
+Data_streams with the `state_` prefix require `kube-state-metrics` to be present.
+
+In order to install `kube-state-metrics` follow the instructions for its deployment, that are available
+https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment[here].
+Generally `kube-state-metrics` runs a `Deployment` and is accessible via a service called `kube-state-metrics` on
+`kube-system` namespace, which will be the service to use in agent configuration
+service within the cluster.
```
tetianakravchenko, line 41:
```suggestion
`kube-system` namespace, which will be the service to use in agent configuration.
```

docs/en/ingest-management/elastic-agent/install-elastic-agent-in-container.asciidoc
```diff
@@ -14,5 +14,7 @@ To learn how to run {agent}s in a containerized environment, see:
 
 * <<running-on-kubernetes-standalone>>
 
+* <<running-on-gke-standard-managed-by-fleet>>
```
tetianakravchenko, line 17: to simplify readability, we could make it looks like a sublist of `Run Elastic Agent on Kubernetes managed by Fleet`
Screenshot 2022-11-14 at 17 41 22

```suggestion
** <<running-on-gke-standard-managed-by-fleet>>
```
wdyt?

PR #2344, opened 2022-11-08
Fix mini toc in level 3 topics

dedemorton:
~Removing the `[discrete]` tags allows the build to create the "On this page" links for level 3 topics in the nav, but it makes the content harder to move around. If we move the topic up a level, we'll need to add the discrete tags back in. And if we move a topic down a level, we'll need to remove the discrete tags to make sure "On this page" gets populated.~

~@gtback Is this a bug or expected behavior? I feel like we ran into this problem with other topics and fixed it, so maybe the fix just needs to be applied to level 3 topics?~

Here is an example of a level 3 topic that does not have "On this page" links: https://www.elastic.co/guide/en/fleet/current/upgrade-elastic-agent.html

To fix this problem, we have decided to bump the heading levels on affected pages instead of removing the discrete tags.

TODO: 
- [x] Remove `discrete` tags from other level 3 topics (if we decide this is expected behavior and do not plan to fix it):
  Here's the list of topics that have missing mini TOCs:
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/add-fleet-server-cloud.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/add-fleet-server-on-prem.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/add-fleet-server-mixed.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/elastic-agent-container.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/running-on-kubernetes-managed-by-fleet.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/running-on-kubernetes-standalone.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/agent-environment-variables.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/monitor-elastic-agent.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/data-streams-ilm-tutorial.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/data-streams-pipeline-tutorial.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/dynamic-input-configuration.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/kubernetes_leaderelection-provider.html
  - [x] https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/kubernetes-provider.html

### DIFF
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -96,7 +96,7 @@ These templates are loaded when the integration is installed, and are used to co
 
 [discrete]
 [[data-streams-ilm]]
-= {ilm} ({ilm-init})
+= Index lifecycle management ({ilm-init})
 
 Use the {ref}/index-lifecycle-management.html[index lifecycle
 management] ({ilm-init}) feature in {es} to manage your {agent} data stream indices as they age.
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -21,7 +21,7 @@ makes sense to your use case or company.
 
 [discrete]
 [[data-streams-naming-scheme]]
-== Data stream naming scheme
+= Data stream naming scheme
 
 {agent} uses the Elastic data stream naming scheme to name data streams.
 The naming scheme splits data into different streams based on the following components:
@@ -78,14 +78,14 @@ All of the juicy details are available in {ref}/data-streams.html[{es} Data stre
 
 [discrete]
 [[data-streams-data-view]]
-== {data-sources-cap}
+= {data-sources-cap}
 
 When searching your data in {kib}, you can use a {kibana-ref}/data-views.html[{data-source}]
 to search across all or some of your data streams.
 
 [discrete]
 [[data-streams-index-templates]]
-== Index templates
+= Index templates
 
 An index template is a way to tell {es} how to configure an index when it is created.
 For data streams, the index template configures the stream's backing indices as they are created.
@@ -96,7 +96,7 @@ These templates are loaded when the integration is installed, and are used to co
 
 [discrete]
 [[data-streams-ilm]]
-== {ilm} ({ilm-init})
+= {ilm} ({ilm-init})
 
 Use the {ref}/index-lifecycle-management.html[index lifecycle
 management] ({ilm-init}) feature in {es} to manage your {agent} data stream indices as they age.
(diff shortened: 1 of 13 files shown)


gtback:
> @gtback Is this a bug or expected behavior? I feel like we ran into this problem with other topics and fixed it, so maybe the fix just needs to be applied to level 3 topics?

I'm not sure. @bmorelli25 and @colleenmcginnis know this code better than me.

It seems like the page in the PR preview has the "on this page" section as expected, though: https://observability-docs_2344.docs-preview.app.elstc.co/guide/en/fleet/master/upgrade-elastic-agent.html

dedemorton:
>It seems like the page in the PR preview has the "on this page" section

Yah, I think that page does not have any `discrete` tags, so it works as expected. It sounds like fixing this bug would be a bit of trouble, so I will juse remove the `discrete` tags and hope that if someone moves this content up in the hierarchy, they will notice that the TOC structure breaks.

dedemorton:
I'm having second thoughts about removing the `discrete` tags because there are more topics affected than I realized. If someone moves the topic up to a different level in the nav at some point, the structure is going to break pretty badly. WDYT?

bmorelli25:
I'm not convinced that there's actually a bug here. I think the content is just nested so deep that it's pushing the heading levels too high for the OTP. I finally had a chance to play around with this, and I'd fix this by changing the heading level, not by removing the `[discrete]` tags.

For example:

```diff
[discrete]
[[upgrade-an-agent]]
- == Upgrade a single {agent}
+ = Upgrade a single {agent}
```

Updating all five headings in `upgrade-elastic-agent.asciidoc` allows us to keep the discrete tags and fix the OTP.

dedemorton:
@bmorelli25 Yah I think the asciidoc(tor) way of chunking content is part of the problem here (no decoupling of navigation structure from heading levels coded in the files) so that we get skipped heading levels in nested topics. IMO 3 levels of nesting in the TOC should _not_ be too deep for our tools to handle. ;-P  I like your fix, though, because it shouldn't break to move things around.

docs/en/ingest-management/data-streams.asciidoc
```diff
@@ -96,7 +96,7 @@ These templates are loaded when the integration is installed, and are used to co
 
 [discrete]
 [[data-streams-ilm]]
-== {ilm} ({ilm-init})
+= {ilm} ({ilm-init})
```
bmorelli25, line 99: Can we change this while we're here? This is bugging me. Sorry, I can't upload a screenshot for some reason. It's a capitalization issue.

```suggestion
= Index lifecycle management ({ilm-init})
```

bmorelli25 (approved):
I clicked through every changed page and LGTM. Just one comment, but it's unrelated to the changes here so up to you if you want to fix.

PR #2346, opened 2022-11-08
actions: use github actions

v1v:
It does exactly what it says on the tin

Already tested in some other projects, see https://github.com/elastic/security-docs/pull/2391


bmorelli25 (approved):

v1v:
Tested it out with https://github.com/elastic/observability-docs/pull/2354#issuecomment-1308651935

PR #2350, opened 2022-11-08
[DOC] Typos

lucabelluccini:
The list was not correctly rendered as this is `asciidoc`, not markdown.

image

I think we should use `.` for numbered lists.


dedemorton (approved):
Good catch! LGTM.

PR #2356, opened 2022-11-10
Correcting the air gap comment

nimarezainia:
in the capabilities table the air gap comment was not updated to reflect the capabilities.


dedemorton (approved):
Good catch! LGTM.

PR #2359, opened 2022-11-14
Fix installation layout docs

cleydyr:
After https://github.com/elastic/elastic-agent/pull/398, the fleet.yml file may not be present in the installation directory. This PR fixes the documentation by removing mentions to the existence of `fleet.yml` in the installation directory and adding notes about the ephemerality of that file where necessary.

### DIFF
--- a/docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
+++ b/docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
@@ -272,7 +272,8 @@ fleet:
   agent:
     id: ""
 ----
-NOTE: Since version 8.3 of the Elastic Agent, after being read, the `fleet.yml` file will be encrypted and renamed to `fleet.enc`.
+
+NOTE: When {agent} runs, the `fleet.yml` file gets encrypted and renamed to `fleet.enc`.
 
 [discrete]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -272,6 +272,7 @@ fleet:
   agent:
     id: ""
 ----
+NOTE: Since version 8.3 of the Elastic Agent, after being read, the `fleet.yml` file will be encrypted and renamed to `fleet.enc`.
```
dedemorton, line 275:
```suggestion

NOTE: When {agent} runs, the `fleet.yml` file gets encrypted and renamed to `fleet.enc`.
```

I don't know if this is accurate ^^. I'm trying to simplify the statement because it's not clear what we mean by "after being read." I think you mean when Elastic Agent runs for the first time, it reads the file, then the file gets encrypted. 

I don't think we need to mention the version number; we can just backport this PR to 8.3.

cleydyr: Looks good to me. :)

dedemorton (approved):
LGTM

PR #2364, opened 2022-11-15
Update ES docs managed by fleet + standalone + autodiscovery

constanca-m:
The main changes of this PR are:

- The elastic agent container
- Elastic agent managed by fleet
- Elastic agent standalone + autodiscovery

I tried to keep the common things like kube-state-metrics, deploy manifests etc in common docs files so it is easier to change them all at once if something changes in the future.

I also deleted the images that were not being used and moved the others to the directory that was using them. I didn't get any errors on that, so I think they are in the right place.


constanca-m:
Just letting you know that the commit has some changes of things that were edited in parallel: every asciidoc with manage-monitors, uptime and synthetics.

PR #2371, opened 2022-11-16
[Fleet] Update logstash client certificate generation

nchaulet:
## Description

Update confusing instructions in logstash client certificate generation. 

In Fleet we force mutual TLS when using the logstash output, but the client certificate is used to verify that the agent is part of the Fleet but not to authenticate a specific agent, the client certificate is shared between all the agent using the logstash output so it should not contains IP or DNS.


dedemorton (approved):
LGTM (assuming this is tested/works). Don't forget backports. :-)

nimarezainia (commented):
looks good.

PR #2395, opened 2022-11-22
Correction on step 5

constanca-m:
Step 5 was pointing to the git manifest, instead of the locally modified one.


dedemorton (approved):
LGTM. Probably needs backport labels though.

PR #2418, opened 2022-11-29
docs: add `0.0.0.0` tip

bmorelli25:
### Summary

This PR updates the APM quick start to:
* Include a tip explaining the port Kubernetes and Docker users should use when setting up the APM integration.
* Links to the common problems section from the bottom of the quick start guide for users who aren't seeing any data flowing into the APM app.

Unfortunately, I think this is the best we can do until https://github.com/elastic/apm-server/issues/2107 lands and provides a nice home for this information. The `0.0.0.0` tip has existed in our [common problems](https://www.elastic.co/guide/en/apm/guide/current/common-problems.html#no-data-indexed) guide since 2018, but clearly, people aren't looking for it there. Hopefully this helps.

* Closes https://github.com/elastic/apm-server/issues/9214

### DIFF
--- a/docs/en/observability/ingest-traces.asciidoc
+++ b/docs/en/observability/ingest-traces.asciidoc
@@ -130,6 +130,8 @@ For a list of indices users need access to, refer to
 [role="screenshot"]
 image::images/kibana-apm-sample-data.png[{apm-app} with data]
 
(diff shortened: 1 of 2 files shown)


dedemorton (approved):
LGTM

PR #2445, opened 2022-12-02
[synthetics] Document migrating from the Synthetics Integration

### LINKED ISSUE #2257: [Synthetics] Document migrating from Synthetics Integration to Private Locations/Project Monitors
## Description

Building on top of https://github.com/elastic/observability-docs/issues/2149 and that we’ll be removing ZIP URLs in the future (via [kibana#137508](https://github.com/elastic/kibana/issues/137508), [kibana#137505](https://github.com/elastic/kibana/issues/137505) and [kibana#137501](https://github.com/elastic/kibana/issues/137501)), this requirement is to provide some information on how users can migrate from using the _legacy_ Synthetics Integration to set up monitors, over to Private Locations / Project Monitors.

## Collaboration
- The docs team will lead producing the content (with support from any of the team)

## Suggested Target Release
- This isn't needed to coincide with the 8.5 release, but there’s no need to wait for the 8.6 release.  This might be a good candidate to work on once all the content needed for the 8.5 release is done, before 8.6 work begins.

## Stakeholders
- @paulb-elastic @andrewvc @dominiqueclarke

colleenmcginnis:
Closes #2257 

@paulb-elastic @andrewvc @dominiqueclarke this is an early draft of a migration guide from the Elastic Synthetics integration to Project Monitors or the Uptime app. I'm open to any and all suggestions! 

The preview for this page is here: https://observability-docs_2445.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-migrate-from-integration.html

Based on Paul's answers to my initial questions (thank you!):

>One thing to mention is that they will have created a default location and schedule as part of init so don't have to set these when migrating them over. 

It isn't totally clear to me (yet) whether they can repurpose the directory they used to zip and add via the integration or if they should create a new project from scratch using `init` and then copy over just the journey files. Let me know what you think.


andrewvc (commented):
It looks great as is, flawless AFAICT.

They should be able to run `npm @elastic/synthetics init <directory/path>` to "upgrade" their legacy projects to use the new projects settings. Even though the old monitors will not have `monitor.use` they will use the default settings that get written to the config file by the `init` command. CC @vigneshshanmugam

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+
+| *Where you configure monitors*
+a| You could configure monitors using:
+
+* A user interface in {kib} (all monitors)
+* Code in an external, version-controlled project (browser monitors _only_)
+a| You can configure monitors using:
+
+* A user interface in {kib} (all lightweight monitors, browser monitors via inline script only)
+* Code in an external, version-controlled project (all monitors)
```
paulb-elastic, line 47: Same as above, maybe _all monitor types_?

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+[discrete]
+[[synthetics-migrate-integration-projects]]
+= How to use projects
+
+// Is this true?
+If you already have an external project you were adding via a ZIP URL
+you can use the same project, but you will have to make some changes:
+
+. Update to the latest version of `@elastic/synthetics`.
+. Configure monitors with code instead of in the UI:
```
paulb-elastic, line 110: I'm not sure I understand this.  As this is in the section about ZIP URLs, won't that mean they will have already defined them in code, as opposed to the UI?

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+| You had to run monitors on your infrastructure.
+a| You can run monitors on both:
+
+* Your infrastructure using <<synthetics-private-location,private locations>>
+* Elastic's global managed infrastructure
+
+| *Where you configure monitors*
+a| You could configure monitors using:
+
+* A user interface in {kib} (all monitors)
```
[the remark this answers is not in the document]
colleenmcginnis:
```suggestion
* A user interface in {kib} (all monitor types)
```

docs/en/observability/synthetics-migrate-integration.asciidoc
colleenmcginnis:
```suggestion
* Code in an external, version-controlled project (all monitor types)
```

docs/en/observability/synthetics-migrate-integration.asciidoc
colleenmcginnis: Yes I see that this is confusing. With the integration you _had_ to configure all the non-journey details (like the name of the monitor, the frequency at which it runs, throttling options, screenshots, etc) in the UI right? And with project monitors now you configure all those details in the code itself instead of in Kibana. If that's correct, do you have any thoughts on how to make that clearer?

paulb-elastic: Ah yes, of course, thanks for the clarification.

I wonder if it would help to give some context here, such as `Configure monitors (such as the frequency) with code instead of in the UI` (or is this too wordy?)

To @andrewvc's [point](https://github.com/elastic/observability-docs/pull/2445#pullrequestreview-1205581794), it's also worth adding a bullet between the current (1) and (2) that describes the need to `npm @elastic/synthetics init...` the folder too.

paulb-elastic:
@colleenmcginnis thanks for your [points](https://github.com/elastic/observability-docs/pull/2445#issuecomment-1339763193)

* Although we put the examples in the `journeys` folder, they don't have to go there, `push` will include `*journey.[ts|js]` files in the root of the project as well as other sub folders
* I just tried an `init` into a folder with a duplicate journey and `init` actually prompts to overwrite, so I think this is covered
  image
* Overwriting `synthetics.config.ts`, `package.json` and `package-lock.json`
  * Yes, I can see these are all overwritten by `init` too
  * Should we enhance `init` to abort in this instance rather than overwrite?  Or at least give them the option like we do with journey files?  (I guess the challenge here is if we don't, the `init` folder might be in a _faulty_ configuration from the start)
  * Keen to get thoughts from @andrewvc and @vigneshshanmugam

colleenmcginnis:
>Should we enhance init to abort in this instance rather than overwrite? Or at least give them the option like we do with journey files? (I guess the challenge here is if we don't, the init folder might be in a faulty configuration from the start)

I totally understand how this would be a problem. Hopefully the fact that it gets overwritten isn't a big deal if they are working in an environment with version control, but I guess I'm wondering if it's worth adding a note to this doc telling them that they will have to manually move over existing configuration settings and/or dependencies/scripts in their `package.json`.

vigneshshanmugam:
> Overwriting synthetics.config.ts, package.json and package-lock.json

The reason for overrding these are files are 
1. Synthetics config - With new init, there is a possibility that Kibana url, locations, space might have changed so we have to replace them on the config file as well. Since the whole push flow is based on this file being present and being correct, skipping is not an option. 
2. `package.json and lock file` - These are dependent on the package manager the user is using, so forgetting to override it might be risky than being beneficial. 

Also some things about project scaffolding,
1. `init` is a one time command, we should not expect or encourage users to call init multiple times to set up Synthetics projects. 
2. Scaffolding a project must always get us closer to a clean state.

paulb-elastic:
Thanks @vigneshshanmugam

As it looks like we _have_ to overwrite, should we at least stop with a prompt (_"You have x and y files which will be overwritten, proceed?"_), so the user knows and can quickly take a copy of them.  I would also suggest maybe renaming the originals (`*.pre-init` or something) so user can still look at what was in there if they needed (_"We have renamed your x to y and a to b..."_).

Maybe I am over-worrying as ZIP URLs are probably not that widely used, I just feel a little nurvous overwriting user's files with no way back for them.

paulb-elastic:
There's a similar conversation over at https://github.com/elastic/synthetics/pull/675#pullrequestreview-1210217633

vigneshshanmugam:
> Maybe I am over-worrying as ZIP URLs are probably not that widely used, I just feel a little nurvous overwriting user's files with no way back for them.

> @vigneshshanmugam does this mean we should not recommend that users migrating from an early version of synthetics used via ZIP URLs to project monitors use init to "upgrade" their project?

We could definitely add a warning or skip the package manager step (Installing dependencies and so on), that would help if the customers are migrating from ZIP based monitors to project monitors. 

What do you both think about this?

paulb-elastic:
> add a warning

I think a prompt/warning (with option to abort) would be useful

> skip the package manager step

I'm not sure how much of a problem this would be to get to the _clean state_ you've been referring to - I'd defer to your advice here (especially if we do the prompt)

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,167 @@
...
+
+[discrete]
+[[synthetics-migrate-integration-projects]]
+= How to use projects
+
+If you already have an external project you were adding via a ZIP URL
+you can use the same project, but you will have to make some changes.
+
+First, upgrade the existing project to use the latest version of 
+`@elastic/synthetics`:
```
colleenmcginnis, line 109: **Option 1: Warn users about the impact of running `init` on an existing project**

```suggestion
If you already have an external project you were adding via a ZIP URL
you can use the same project, but you will have to make some changes.

First, upgrade the existing project to use the latest version of 
`@elastic/synthetics`.

[IMPORTANT]
====
*Upgrading your project using these instructions will overwrite your existing synthetics configuration file (`synthetics.config.ts` or `synthetics.config.js`) and `package.json` file.*

If you have customized either of these files in your existing project, be sure you
have the current version saved in version control or create a fresh project using
`npm @elastic/synthetics init` with a path to a new directory instead of the path to
existing project. Then, manually transfer:
* Any synthetics configuration options to the new synthetics configuration file.
* Any dependencies, scripts, or other configurations to the new `package.json` file.
====
```

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,167 @@
...
+. Upgrade your existing project to use new project settings:
+.. Run `npm @elastic/synthetics init <path-to-existing-project>`.
+.. Respond to all prompts provided by the CLI.
+. Review updated files and directories, including:
+.. `journeys/` will contain sample journey code.
+   Move existing journey files into this directory.
+.. `synthetics.config.ts` will contain updated default settings
+   needed for the upgraded project.
+.. `package.json` will contain updated NPM settings for your project.
+.. `.github/` will contain sample workflow files to use with GitHub Actions.
```
colleenmcginnis, line 122: **Option 2: Instruct users to create a fresh project instead of "upgrading"**

```suggestion
. Run `npm install -g @elastic/synthetics@latest` to install
  the latest version of the CLI.
. Create a new project:
.. Run `npm @elastic/synthetics init <new-directory>`.
.. Respond to all prompts provided by the CLI.
. Review the generated files and directories, including:
.. `journeys/` will contain sample journey code.
   Move existing journey files into this directory.
.. `synthetics.config.ts` will contain updated default settings
   needed for the upgraded project. Manually copy over previous configuration options that are compatible with the latest version of Elastic Synthetics.
.. `package.json` will contain updated NPM settings for your project. Manually copy over previous dependencies, scripts, and other pacakage manager configuration options.
.. `.github/` will contain sample workflow files to use with GitHub Actions.
```

colleenmcginnis (commented):
>What do you both think about this?

I don't have a strong opinion here. Either way, it sounds like we will have to warn users about avoiding overwriting any customizations they made in their existing `synthetics.config.ts/js` file _outside_ the CLI. I guess for me the question is which approach should we take in the docs?

* Option 1: Warn users about the impact of running init on an existing project
* Option 2: Instruct users to create a fresh project instead of "upgrading"

Option 2 is probably safer, but likely adds more friction to the process. I don't have a good sense whether the risk of overwriting an existing config is big enough to justify defaulting to recommending a process with more friction.

Let me know if I'm overthinking. 🙂

vigneshshanmugam:
> Either way, it sounds like we will have to warn users about avoiding overwriting any customizations they made in their existing synthetics.config.ts/js file outside the CLI

From the users perspective or from ours, its really not useful to do go through the init process if the user is not overriding the Synthetics config file since all of the question prompts would eventually get saved in to the file.

paulb-elastic:
@vigneshshanmugam you make a good case that these have to be overwritten, so I am happy that we have to do that for it to be a working folder.

However, if you were a user of this, and in terms of _least surprise_, which of these do you think would be best:
* Just overwrite on `init`, having given warnings about it in the docs (Colleen's option 1)
  * My worry with this one is that it might still be a suprise to the user unless they've paid close attention to the docs
* Document users should create a new `init` folder and move things over (Colleen's option 2)
  * There's nothing to stop users still doing option 1, just we wouldn't doc this as the recommended approach
* Option 1 but make it clear to the user
  * If possible, `init` detects it's an existing folder and gives them a prompt (_"Continuing will overwrite the ... files, are you sure?"_)
  * Still document option 1 as the recommended approach
  * This still overwrites the files getting to a _clean_ state but at least the user is aware and can abort if they need to make a copy
  * Would require a change to Synthetics (seems relatively simple, but keen to get your view Vignesh)

colleenmcginnis:
>There's nothing to stop users still doing option 1, just we wouldn't doc this as the recommended approach

For what it's worth, I had no idea I could run `init` on an existing project before someone told me I could while working on this PR. From my perspective, running `init` on an existing project to "upgrade" (rather than initializing a fresh project) is not very intuitive. If others feel the same way, I would _think_ the risk would be low of users doing option 1 without being told it's an option.

vigneshshanmugam:
Great summary Paul. Running `init` multiple times might be necessary when we add new features like `adding lightweight monitors` or upgrading experience if we want to support new locations and so on. 

> Option 1 but make it clear to the user

Given that, I would like to resort to Option 3 that you have here https://github.com/elastic/observability-docs/pull/2445#issuecomment-1348352903 which is based on the Option 1. 

- The agent would allow users to override the files, but prompts on overwrite. This gets the clean state, plus also keeps the push command from not breaking if the user tries to run it after scaffolding a existing project. 
- As you stated, this would require a change in the agent, but should be minimal.

colleenmcginnis:
@paulb-elastic it looks like the changes discussed above are not going to be ready for the 8.6 launch ([this issue](https://github.com/elastic/synthetics/issues/680) mentions 8.7). Should I hold this PR until that is ready or use [one of these options](https://github.com/elastic/observability-docs/pull/2445#pullrequestreview-1214203891) for the 8.6 launch and update when the changes make it into `@elastic/synthetics`?

paulb-elastic:
Thanks for picking this up this discrepancy @colleenmcginnis 

Because the Synthetics Agent doesn't strictly follow the stack release, https://github.com/elastic/synthetics/issues/680 went onto the 8.7 board as our _active_ board.

For example, we will cut a new version of the Synthetics Agent to coincide with the 8.6 release, but won't do that until nearer the release date.  I would think we **can** get https://github.com/elastic/synthetics/issues/680 into the next Synthetics release, which will be available for 8.6 (@vigneshshanmugam do you agree that this can be included in the beta.40 release?).

I'll move https://github.com/elastic/synthetics/issues/680 onto the 8.6 board for now to help clarify that (and we'll need to change approach if @vigneshshanmugam doesn't think we can get it in for beta.40).

colleenmcginnis:
Thanks @paulb-elastic! I subscribed to that synthetics issue so I'll know when it makes it into a release.

vigneshshanmugam:
Will try to do this in the agent as part of the planned 8.6 release update (v1.0.0-beta.40)

vigneshshanmugam:
Its done - https://github.com/elastic/synthetics/pull/688

colleenmcginnis:
@paulb-elastic can you or someone from your team take a last look and suggest any final changes or approve? I will wait to merge until I see that @elastic/synthetics v1.0.0-beta.40 was released and verify that https://github.com/elastic/synthetics/pull/688 was included. Thanks!

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,172 @@
...
+.. `journeys/` will contain sample journey code.
+Move existing journey files into this directory.
+.. `synthetics.config.ts` will contain updated default settings
+needed for the upgraded project.
+.. `package.json` will contain updated NPM settings for your project.
++
+[NOTE]
+====
+If there is already a `package.json` file in the directory, it will only
+modify the existing file if no `test` or `push` script is defined.
```
paulb-elastic, line 130: I know this came from https://github.com/elastic/synthetics/pull/688 but I am not sure it provides enough information, or is clear in describing the behaviour, or what a user should do.

Similarly, I don't think we refer to this beaviour anywhere else like https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html

@vigneshshanmugam can you provide clarity here?

paulb-elastic (changes requested):
In general, LGTM, except for my comment on using `test` or `push`.  I think this should be a bit clearer, and maybe include details of this behaviour on https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html too.

docs/en/observability/synthetics-migrate-integration.asciidoc
vigneshshanmugam: We need to modify it as, If there is already existing `package.json`, 
1. Synthetics agent will not create a new `package.json` file
2. Instead only adds the `@elastic/synthetics` library to the dependencies if its not already present, and modifies the `test/push` script if not present already.

docs/en/observability/synthetics-migrate-integration.asciidoc
```diff
@@ -0,0 +1,172 @@
...
+====
```
colleenmcginnis, line 131: Is this clearer @paulb-elastic @vigneshshanmugam ?

```suggestion
[NOTE]
====
If there is already a `package.json` file in the directory when you run `init`,
the synthetics agent will _not_ create a new `package.json` file.
Instead it will modify the existing `package.json` file to:

* Add `@elastic/synthetics` library to the dependencies if it's not already present.
* Add a `test` and `push` script if they are not already present.
====
```

vigneshshanmugam: Looks good to me 👍🏽

vigneshshanmugam (approved):
LGTM

paulb-elastic (approved):
LGTM

PR #2479, opened 2023-01-09
Add unique input IDs to all agent policy examples.

cmacknz:
Starting in 8.6 inputs in agent policies require unique IDs. Prior to 8.6 a unique ID for each input isn't required but should be encouraged as some integrations rely on the id values being unique and stable. Agent policies generated by Fleet have included unique IDs per input for several releases now. This change only affects standalone agents.

When there is only a single instance of a particular input type (for example when there is one `logfile` input) then the policy will still work but the id will be the empty string `""`. Adding a second instance of the input without an ID will trigger an error starting in 8.6. Editing the first input with the empty ID to have a named ID may have unexpected consequences like log duplication, so it is safer to use an explicit unique ID from the start.

- Closes https://github.com/elastic/elastic-agent/issues/1991


dedemorton (approved):
LGTM. Thank you for updating the docs!

Issue #2458, opened 2022-12-19
[REQUEST]: Fix Fleet supported outputs table


lucabelluccini:
### Description

The page https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#supported-outputs-beats-and-agent contains a "feature" table regarding supported outputs by Beats, Fleet-Agent & Standalone Agent.

We should probably add a `*` on all the outputs except Elasticsearch to highlight the fact APM Integration ONLY supports Elasticsearch output.
In the current state, users might be misled into thinking APM Integration can be used with any supported output type.

Final users will realize they cannot use Logstash output only when creating the policy and trying to add the APM Integration - which will be denied.

### Resources

This is derived from https://github.com/elastic/apm-server/issues/7025#issuecomment-1031785020, the docs https://www.elastic.co/guide/en/apm/guide/current/upgrade-to-apm-integration.html#apm-integration-upgrade-limitations and an internal discussion with the APM team

This should go in couple with https://github.com/elastic/apm-server/issues/9848

### Collaboration

Other (please describe below in the Point of contact section).

### Point of contact.

**Main contact:** @simitt 

**Stakeholders:** @lucabelluccini 

Obviously feel free to add/remove anyone

dedemorton:
Looks like this was closed by https://github.com/elastic/observability-docs/pull/2491

PR #2491, opened 2023-01-11
Excluding APM server from the output section of compatibility page

nimarezainia:
Addressing https://github.com/elastic/observability-docs/issues/2458

@dedemorton I couldn't find a global variable for APM server (like {apm-server} ), so just called it out.

### DIFF
--- a/docs/en/ingest-management/beats-agent-comparison.asciidoc
+++ b/docs/en/ingest-management/beats-agent-comparison.asciidoc
@@ -67,7 +67,7 @@ To find out if an integration is GA, see the
 The following table shows the outputs supported by the {agent} in {version}:
 
 
-NOTE: {elastic-defend} has a different output matrix.
+NOTE: {elastic-defend} and APM server have a different output matrix.
 
 [options,header]
 |===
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/beats-agent-comparison.asciidoc
```diff
@@ -67,7 +67,7 @@ To find out if an integration is GA, see the
 The following table shows the outputs supported by the {agent} in {version}:
 
 
-NOTE: {elastic-defend} has a different output matrix.
+NOTE: {elastic-defend} and APM server have a different output matrix.
```
dedemorton, line 70:
```suggestion
NOTE: {elastic-defend} and APM Server have a different output matrix.
```

dedemorton: They capitalize server. Not sure why we don't have a shared attribute for this, but it looks like they hard code it in the APM docs, so let's go with that.

dedemorton (approved):
LGTM. I'll apply my suggestion and merge this. Thanks!

dedemorton:
@nimarezainia Which versions do you want this backported to?

PR #2510, opened 2023-01-13
spell updates are required

simranvaseer-qasource:
As per my understanding, `status` is misspelled as stats at line no. 140


dedemorton:
Closing because this is not a typo. The word "stats" does not mean "status" in this context. "Stats" is a short form of the word "statistics" and used commonly when describing statistics gathered about system components.

PR #2513, opened 2023-01-16
Correction to the processor document for agent

nimarezainia:
just a small clarification. thanks

### DIFF
--- a/docs/en/ingest-management/processors/processors.asciidoc
+++ b/docs/en/ingest-management/processors/processors.asciidoc
@@ -69,8 +69,8 @@ inputs:
           - add_locale: null
 ----
 
-You can define processors that apply to all inputs defined in the configuration,
-or specific inputs.
+You can define processors that apply to a specific input defined in the configuration. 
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/processors/processors.asciidoc
```diff
@@ -69,8 +69,8 @@ inputs:
           - add_locale: null
 ----
 
-You can define processors that apply to all inputs defined in the configuration,
-or specific inputs.
+You can define processors that apply to a specific input defined in the configuration. 
+Applying a processor to all the inputs on a global basis is currently not supported
```
dedemorton, line 73:
```suggestion
Applying a processor to all the inputs on a global basis is currently not supported.
```

dedemorton (commented):
Just a minor edit...

Is there an issue open to track support for global processors? If so, you should add a note there to remind us to remove this restriction when that functionality is supported. Otherwise there is a good chance we'll forget to remove this sentence.

dedemorton (approved):

PR #2515, opened 2023-01-17
Fix app log ingest doc

SylvainJuge:
- fix indent in filebeat config
- avoid .json suffix for plaintext log file


bmorelli25 (approved):
Thank you!

PR #2518, opened 2023-01-17
max number of policies

nimarezainia:
currently tested to 500

### DIFF
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -278,4 +278,4 @@ will still forward data to the cluster.
 [[agent-policy-scale]]
 == Scale of Policies Supported
 
-Maximum number of Elastic Agent policies supported in a single {fleet} is 500. You may experience slowness in response when a higher number of policies are configured. 
+Maximum number of Elastic Agent policies supported by {fleet} is 500. You may experience slowness in response when a higher number of policies are configured. 
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -278,4 +278,4 @@ will still forward data to the cluster.
 [[agent-policy-scale]]
 == Scale of Policies Supported
 
-Maximum number of Elastic Agent policies supported by {fleet} is 500. You may experience slowness in response when a higher number of policies are configured. 
+A single instance of {fleet} supports a maximum of 500 {agent} policies. If more policies are configured, UI performance might be impacted.
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -273,3 +273,9 @@ Make sure that the {agent}s assigned to this policy all have connectivity to the
 that you added. Lack of connectivity will prevent the {agent}
 from checking in with the {fleet-server} and receiving policy updates, but the agents
 will still forward data to the cluster.
+
+[discrete]
+[[agent-policy-scale]]
+== Scale of Policies Supported
+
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -273,3 +273,9 @@ Make sure that the {agent}s assigned to this policy all have connectivity to the
 that you added. Lack of connectivity will prevent the {agent}
 from checking in with the {fleet-server} and receiving policy updates, but the agents
 will still forward data to the cluster.
+
+[discrete]
+[[agent-policy-scale]]
+== Scale of Policies Supported
```
dedemorton, line 279: I have a few thoughts about this section:

* "Scale of policies supported" doesn't seem intuitive unless you think users will search on this phrase. Maybe replace with "Policy limits" or "Policy scaling recommendations"?
* I think it would be better to have this section appear earlier in the topic, maybe under the section about policy benefits?

docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -273,3 +273,9 @@ Make sure that the {agent}s assigned to this policy all have connectivity to the
...
+
+Maximum number of Elastic Agent policies supported in a single {fleet} is 500. You may experience slowness in response when a higher number of policies are configured.
```
dedemorton, line 281: I have a couple questions before I can edit this part:

* What do we mean by a "single Fleet"? Are we talking about one instance of Kibana with Fleet? My concern here is that users will conflate a "single Fleet" with the number of Fleet Servers they have deployed. Can you think of a better way to describe this?
* How does the user experience the slow response that you describe? Are you talking about UI performance? Or is there some delay in the data being collected, parsed, and indexed into Elasticsearch?

dedemorton (changes requested):
I have a few questions before I can edit this properly.

docs/en/ingest-management/agent-policies.asciidoc
nimarezainia: agree on both points yes. it is a recommendation indeed.

docs/en/ingest-management/agent-policies.asciidoc
nimarezainia: @dedemorton I modified it to just say "fleet". Slowness in response is in UI. If you feel that is too much information we can drop it. thanks.

docs/en/ingest-management/agent-policies.asciidoc
dedemorton, line 279:
```suggestion
== Policy scaling recommendations
```

docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -273,3 +273,9 @@ Make sure that the {agent}s assigned to this policy all have connectivity to the
 that you added. Lack of connectivity will prevent the {agent}
 from checking in with the {fleet-server} and receiving policy updates, but the agents
 will still forward data to the cluster.
+
+[discrete]
+[[agent-policy-scale]]
+== Scale of Policies Supported
+
+Maximum number of Elastic Agent policies supported by {fleet} is 500. You may experience slowness in response when a higher number of policies are configured.
```
dedemorton, line 281:
```suggestion
The maximum number of {agent} policies supported by {fleet} is 500. If more than 500 policies are configured, {fleet} UI performance might be affected.
```
"You may experience slowness" is grammatically vague (the user does not experience slowness).

dedemorton (changes requested):

docs/en/ingest-management/agent-policies.asciidoc
nimarezainia: sounds good.

dedemorton (approved):
LGTM

dedemorton:
@nimarezainia Please add any missing backport labels. Thanks!

PR #2525, opened 2023-01-18
Add new known issue to 8.6 agent release notes

cmacknz:
Add https://github.com/elastic/elastic-agent/issues/2086 as a known issue in the release notes.

Also removes an entry about fixing duplicated events on upgrade that was not actually fixed in 8.6.

### DIFF
--- a/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
@@ -128,7 +128,6 @@ The 8.6.0 release adds the following new and notable features.
 
 {agent}::
 * {agent} now uses the locally bound port (8221) when running {fleet-server} instead of the external port (8220) {agent-pull}1867[#1867]
-* Correctly preserve the {filebeat} registry and other persistent input states during upgrades to eliminate event duplication after upgrades {agent-pull}1701[#1701] {agent-issue}836[#836]
 // end 8.6.0 relnotes
 
 
--- a/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
@@ -85,7 +85,7 @@ Be aware that the live query results shown in {kib} may be delayed by up to 5 mi
 ====
 
 [[known-issue-issue-2086]]
-.Beats started by agent can be configured without a valid output
+.Beats started by agent may fail with `output unit has no config` error
 [%collapsible]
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -84,6 +84,22 @@ For more information, refer to {agent-issue}2066[#2066].
...
+[[known-issue-issue-2086]]
+.Beats started by agent can be configured without a valid output
+[%collapsible]
+====
+*Details* +
+A known issue in {agent} may lead to Beat processes being started without a
+valid output. The problem can be corrected by triggering a restart of {agent}
+or the affected Beats. Beats managed by {agent} can be restarted by changing the
+{agent} log level or the output section of the {agent} policy.
+For more information, refer to {agent-issue}2086[#2086].
```
dedemorton, line 96:
```suggestion
A known issue in {agent} may lead to Beat processes being started without a
valid output. To correct the problem, trigger a restart of {agent}
or the affected Beats. For Beats managed by {agent}, you can trigger a restart by changing the
{agent} log level or the output section of the {agent} policy.
For more information, refer to {agent-issue}2086[#2086].
```
Just trying to remove a bit of the passive voice.

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -84,6 +84,22 @@ For more information, refer to {agent-issue}2066[#2066].
 Be aware that the live query results shown in {kib} may be delayed by up to 5 minutes.
 ====
 
+[[known-issue-issue-2086]]
+.Beats started by agent can be configured without a valid output
```
dedemorton, line 88:
```suggestion
.Beats started by agent may fail with `output unit has no config` error
```
I think the original could be misread because the passive voice makes it unclear as to what is doing the configuring (user or system). I suggest including the error message because that's what users might search on.

dedemorton (approved):
Offered a couple edits to clarify the text.

dedemorton (changes requested):
Sorry didn't mean to approve previously

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
cmacknz: Thanks, this is much clearer

dedemorton (approved):
LGTM. Thanks for adding this!

PR #2452, opened 2022-12-14
Application logging docs

bmorelli25:
### Summary

cc @SylvainJuge 

[**Link to preview**](https://observability-docs_2452.docs-preview.app.elstc.co/guide/en/observability/master/application-logs.html)

### Related issues

- Replaces https://github.com/elastic/apm-server/pull/9751
- Related https://github.com/elastic/apm-agent-java/pull/2916
- Closes https://github.com/elastic/observability-docs/issues/1295
- Related https://github.com/elastic/observability-docs/issues/2362

### Related links

- https://github.com/elastic/apm/blob/main/specs/agents/log-sending.md
- https://www.elastic.co/guide/en/apm/guide/8.5/log-correlation.html
- https://www.elastic.co/guide/en/apm/agent/java/current/log-correlation.html
- https://www.elastic.co/guide/en/apm/agent/dotnet/current/log-correlation.html
- https://www.elastic.co/guide/en/apm/agent/python/current/log-correlation.html
- https://www.elastic.co/guide/en/apm/agent/nodejs/current/log-correlation.html
- https://www.elastic.co/guide/en/apm/agent/ruby/current/log-correlation.html
- https://www.elastic.co/guide/en/ecs-logging/overview/master/intro.html

### To do in a future PR:

* Add a step 3 to all methods
* Change imports to ECS logging book
* Fix filebeat.yaml sample files: some config options are out of date and others aren't explained well
* Update links to Java and Ruby agents
*

PR #2542, opened 2023-01-24
docs: Finish Application logging docs

bmorelli25:
### Summary

This PR is a follow-up to https://github.com/elastic/observability-docs/pull/2452. Its main purpose is to update the Filebeat configuration with additional details and updated configuration options.

### DIFF
--- a/docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc
+++ b/docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc
@@ -17,8 +17,10 @@ filebeat.inputs:
       overwrite_keys: true <2>
       add_error_key: true <3>
       expand_keys: true <4>
+  fields:
+    service.name: your_service_name <5>
 
-processors: <5>
+processors: <6>
   - add_host_metadata: ~
   - add_cloud_metadata: ~
   - add_docker_metadata: ~
@@ -28,7 +30,8 @@ processors: <5>
 <2> Values from the decoded JSON object overwrite the fields that {filebeat} normally adds (type, source, offset, etc.) in case of conflicts.
 <3> {filebeat} adds an "error.message" and "error.type: json" key in case of JSON unmarshalling errors.
 <4> {filebeat} will recursively de-dot keys in the decoded JSON, and expand them into a hierarchical object structure.
-<5> Processors enhance your data. See {filebeat-ref}/filtering-and-enhancing-data.html[processors] to learn more.
+<5> The `service.name` of the service you're collecting logs from. This is required for <<log-correlation>>.
+<6> Processors enhance your data. See {filebeat-ref}/filtering-and-enhancing-data.html[processors] to learn more.
 endif::ecs-logs[]
 ifdef::plaintext[]
 . Configure filebeat.yaml file to start collecting log data.
--- a/docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc
+++ b/docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc
@@ -14,18 +14,21 @@ filebeat.inputs:
   paths: /path/to/logs.json
   parsers:
     - ndjson:
-      keys_under_root: true
-      overwrite_keys: true
-      add_error_key: true
-      expand_keys: true
+      overwrite_keys: true <2>
+      add_error_key: true <3>
+      expand_keys: true <4>
 
-processors:
+processors: <5>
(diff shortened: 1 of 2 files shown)


docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc
```diff
@@ -14,18 +14,21 @@ filebeat.inputs:
   paths: /path/to/logs.json
   parsers:
     - ndjson:
-      keys_under_root: true
-      overwrite_keys: true
-      add_error_key: true
-      expand_keys: true
+      overwrite_keys: true <2>
+      add_error_key: true <3>
+      expand_keys: true <4>
```
bmorelli25, line 19: Removed `keys_under_root` as that is now the default. This can be changed with `target`, but I didn't think it was necessary to list it here as we're not suggesting changing the default.

SylvainJuge: @bmorelli25 do you know which version introduced this change and does it applies to other keys that are injected into the document ?
In the ready-to-use examples that I currently have (initially built for 8.5.3, but now upgraded to 8.6.0):
- `fields_under_root`: https://github.com/elastic/apm-contrib/blob/b99ee6225f01bad97901396d63a2f29d31a57f2a/apm-agent-java/log-ingest/01-filebeat.yml#L20 (for plain-text logs)
- `keys_under_root` https://github.com/elastic/apm-contrib/blob/b99ee6225f01bad97901396d63a2f29d31a57f2a/apm-agent-java/log-ingest/02-filebeat.yml#L13 (for nd-json logs)

SylvainJuge:
In order to have correlation between log documents and APM services, we need to (at least) set the value of `service.name` (and maybe few other fields but I don't know yet which ones).
This is required for the following strategies:
- plain-text with filebeat
- ECS logging with filebeat:
  - `service.name` can be set in ECS logging configuration
  - when not set at ECS logging configuration level and deployed with the Java agent, then the Java agent will set `service.name` with the value from its own configuration.
  
When using the other strategies (ECS reformatting and Log sending), then the ECS logging library in the agent is configured with the `service.name` from the agent configuration, hence making it implicit.
  
Without this, the end-user will only be able to see logs documents in the Logs Stream UI, not in APM.

SylvainJuge:
The `service.name` field is currently the main way to correlate the logs, which is properly described in https://github.com/elastic/kibana/issues/148788 description:

> Logs annotated with a matching `service.name` are displayed . Log lines without `service.name` will be matched using `container.id` and then `host.name` (in that order).

I don't think that we need to document the `container.id` and `host.name` fallback as the `container.id` is very unlikely to be set by the end user, but it can be implicitly provided when using the APM agents.

dedemorton (approved):
Changes LGTM from an editorial perspective.

docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc
bmorelli25: For nd_json logs, `keys_under_root` doesn't do anything and never has; a documentation error made it appear necessary. In 8.2, the documentation was updated to show `target` instead of `keys_under_root`. If `target` is not set, fields go to the root of the event by default ([source](https://github.com/elastic/beats/issues/29422#issuecomment-1105361562)).

For plain text logs, you are correct. `fields_under_root` is needed to store custom fields as top-level fields in the output document instead of being grouped under a `fields` sub-dictionary ([source](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-filestream.html#fields-under-root-filestream)).

bmorelli25:
Thanks! 

For `plaintext`, I added a step on log correlation. For `ECS logging with filebeat`, it looks like all of the ECS guides mention log correlation, so I'm not sure we need to add it to these docs.

SylvainJuge (approved):

SylvainJuge:
For the plain-text logs, I think that adding the `service.name` extra field configuration would be relevant here : https://github.com/elastic/observability-docs/blob/ce3ebd1dd6ec052242580ae2bed24943f246cb2d/docs/en/observability/tab-widgets/filebeat-logs/content.asciidoc?plain=1#L43

While there is a mention of `service.name` in the "Application logs" section, this is quite far away and for plaintext it will always be required (unlike ECS logging where it can be set either through logger configuration or directly by the APM agent), hence usually not required in the filebeat configuration.

PR #2559, opened 2023-01-27
Fix broken automation

bmorelli25:


bmorelli25:
I guess I can't test this unless I merge... so gonna do that and see what happens.

Issue #2278, opened 2022-10-13
[Request] Make tasks consistent in the Fleet/Agent docs


dedemorton:
## Description

In the Fleet and Elastic Agent Guide, we are wildly inconsistent wrt how we describe navigating through the menus, pages, and tabs to accomplish specific tasks. This is an artifact of how we develop content. We need to make an editorial pass through all the tasks and clean up the steps to make sure:

- The steps are consistent.
- The user has enough context to get them to the correct page to perform the task.
- We are consistent with how Kibana documents similar tasks.

## Collaboration
- The docs team will lead producing the content

jillguyonnet:
Linking https://github.com/elastic/observability-docs/pull/2566, now closed, regarding the possible confusion between the `Management` section in Kibana's main menu and the `Management` section in Kibana's homepage.

PR #2566, opened 2023-01-31
Make location of Fleet UI more explicit

jillguyonnet:
### Context
New users could be confused between the Management section in Kibana's sidebar and the Management section in the homepage.

### Change
Replace "In Kibana" with "In Kibana sidebar" for `Elasticsearch Service` and `Self-managed` tabs.

### Screenshot
image

### DIFF
--- a/docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
+++ b/docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
@@ -6,7 +6,7 @@ deployment.
 
 To confirm that an {integrations-server} is available in your deployment:
 
-. In {kib}, go to **Management > {fleet}**.
+. In {kib} sidebar, go to **Management > {fleet}**.
 . On the **Agents** tab, look for the **{ecloud} agent policy**. This policy is
 managed by {ecloud}, and contains a {fleet-server} integration and an Elastic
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/tab-widgets/add-fleet-server/content.asciidoc
```diff
@@ -6,7 +6,7 @@ deployment.
 
 To confirm that an {integrations-server} is available in your deployment:
 
-. In {kib}, go to **Management > {fleet}**.
+. In {kib} sidebar, go to **Management > {fleet}**.
```
dedemorton, line 9: A couple of thoughts:

* Our guidelines recommend that we avoid using the term "sidebar". To follow the guidelines, we should probably say something like, "In {kib}, open the main menu, then click {fleet}." (not sure we need to mention the management container)
image

* We should make this change globally rather than in just one topic.
* There's already [an issue open to make the tasks more consistent](https://github.com/elastic/observability-docs/issues/2278). You could close this PR and add a comment to that issue instead so that we fix all occurrences of the problem.

dedemorton (changes requested):
See my comments. Let's close this PR and add to the existing issue so that the change gets into all the docs.

jillguyonnet:
Thank you for your review and insights @dedemorton 🙏 Closing this PR and adding a comment to the existing issue.

PR #2693, opened 2023-02-23
Updating the fleet server scalability resources

nimarezainia:
Based on our recent testing results, modifying the fleet server resource consumption values. Also removing the various parameter values that have been confusing to the users. Instead keying everything from the number of agents configured.

### DIFF
--- a/docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
+++ b/docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
@@ -139,47 +139,12 @@ on the number of agents required by your deployment:
 === Resource requirements by number of agents
 |===
 | Number of Agents | Memory       | vCPU            | {es} Cluster size
-| 50               | 1 GB         | Up to 8.5 vCPU  | 480 GB disk \| 16 GB RAM \| up to 5 vCPU
-| 5,000            | 2 GB         | Up to 8.5 vCPU  | 960 GB disk \| 32 GB RAM \| 5 vCPU
-| 7,500            | 4 GB         | Up to 8.5 vCPU  | 1.88 TB disk \| 64 GB RAM \| 9.8 vCPU
-| 10,000           | 8 GB         | Up to 8.5 vCPU  | 3.75 TB disk \| 128 GB RAM \| 19.8 vCPU
-| 15,000           | 16 GB        | 8.5 vCPU        | 7.5 TB disk \| 256 GB RAM \| 39.4 vCPU
-| 25,000           | 16 GB        | 8.5 vCPU        | 7.5 TB disk \| 256 GB RAM \| 39.4 vCPU
-| 50,000           | 32 GB        | 16.9 vCPU       | 11.25 TB disk \| 384 GB RAM \|59.2 vCPU
+| 2,000            | 2 GB         | 2.5 vCPU        | 480 GB disk (hot tier) \| 16 GB RAM \| 5 vCPU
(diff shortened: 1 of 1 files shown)


kilfoyle (approved):
LGTM! 🚀 
@nimarezainia I pushed a commit to remove the `<<recommend-settings-scaling-agents>>` link, which was broken due to the removed section. The CI checks should pass now. 🤞

docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
```diff
@@ -132,54 +132,18 @@ The following tables provide resource requirements and scaling guidelines based
 on the number of agents required by your deployment:
 
 * <<resource-requirements-by-number-agents>>
-* <<recommend-settings-scaling-agents>>
 
 [discrete]
 [[resource-requirements-by-number-agents]]
 === Resource requirements by number of agents
 |===
 | Number of Agents | Memory       | vCPU            | {es} Cluster size
```
joshdover, line 140: Can we also clarify that Memory and vCPU are for Fleet Server?

docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
```diff
@@ -132,54 +132,18 @@ The following tables provide resource requirements and scaling guidelines based
...
-| 15,000           | 16 GB        | 8.5 vCPU        | 7.5 TB disk \| 256 GB RAM \| 39.4 vCPU
-| 25,000           | 16 GB        | 8.5 vCPU        | 7.5 TB disk \| 256 GB RAM \| 39.4 vCPU
-| 50,000           | 32 GB        | 16.9 vCPU       | 11.25 TB disk \| 384 GB RAM \|59.2 vCPU
+| 2,000            | 2 GB         | 2.5 vCPU        | 480 GB disk (hot tier) \| 16 GB RAM \| 5 vCPU
+| 5,000            | 4 GB         | 2.5 vCPU        | 480 GB disk (hot tier) \| 16 GB RAM \| 5 vCPU
+| 10,000           | 8 GB         | 2.5 vCPU        | 480 GB disk (hot tier) \| 16 GB RAM \| 5 vCPU
+| 15,000           | 8 GB         | 2.5 vCPU        | 480 GB disk (hot tier) \| 16 GB RAM \| 5 vCPU
+| 25,000           | 8 GB         | 2.5 vCPU        | 480 GB disk (hot tier) \| 16 GB RAM \| 5 vCPU
+| 50,000           | 8 GB         | 8 vCPU          | 480 GB disk (hot tier) \| 128 GB RAM x2 \| 64 vCPU x2
+| 100,000          | 8 GB x2      | 8 vCPU x2       | 480 GB disk (hot tier) \| 256 GB RAM x2 \| 128 vCPU x2
```
joshdover, line 147: I'm not sure how much value there is in showing the disk size at all since Fleet isn't dependent on it. We should also make clear that you need the RAM and vCPUs to be on the hot tier. I'd recommend instead:
- Change the header of the column to be `{es} Hot tier capacity`
- Only include the RAM and vCPU requirements

docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
nimarezainia: yep good point done.

docs/en/ingest-management/fleet/fleet-server-scaling.asciidoc
nimarezainia: done

PR #2717, opened 2023-02-27
Add a known issue for missing elastic-agent signatures on MacOS in 8.6.1

cmacknz:
Add a known issue explaining that the elastic-agent executable was unsigned in 8.6.1, which will cause upgrades to any other version to fail on MacOS Ventura. The failed upgrades can lead to non-functional Elastic Defend deployments.


docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -196,6 +196,40 @@ Affected {fleet-server} instances must be reinstalled to fix the problem.
...
+Signature size=9068
+Authority=Developer ID Application: Elasticsearch, Inc (2BT3HPN62Z)
+Authority=Developer ID Certification Authority
+Authority=Apple Root CA
+Timestamp=Feb 24, 2023 at 4:33:02 AM
+...
+----
+
+*Impact* +
+Any {agent} deployed to MacOS Ventura that was upgraded to version 8.6.1 prior to February 27th 2023 must be reinstalled.
```
lucabelluccini, line 228: Can we clarify:
- The user will need to re-download the binary
- How to reinstall once we have the binary? How the user should handle the "leftovers" of the unsigned 8.6.1 installation? Do they have to unenroll first? If so, how?

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -196,6 +196,40 @@ Affected {fleet-server} instances must be reinstalled to fix the problem.
 Upgrade to 8.6.2 to fix the problem. For more information, refer to {fleet-server-issue}2328[#2328].
 ====
 
+[[known-issue-issue-unsigned-8.6.1]]
+.The initial release of {agent} 8.6.1 for MacOS contained an unsigned elastic-agent executable. On MacOS Ventura, upgrading from 8.6.1 to any other version will fail and can disable Elastic Defend protections.
+[%collapsible]
+====
+*Details* +
+The initial release of {agent} version 8.6.1 for MacOS contained an unsigned elastic-agent executable and a correctly signed endpoint-security
+executable. The endpoint-security exectuable implements the endpoint protection functionality of the Elastic Defend integration.
```
lucabelluccini, line 205:
```suggestion
executable. The endpoint-security executable implements the endpoint protection functionality of the Elastic Defend integration.
```

dedemorton: Good catch. I think I need new glasses. :-)

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
cmacknz: I will clarify that they need to re-download the binary first. Good point.

The full sequence needs to be:

1. Download a correctly signed version of the agent.
2. Unenroll the affected agent because the re-installation will generate a new agent ID.
3. Run `elastic-agent uninstall` to remove the previously installed artifacts.
4. Run `elastic-agent install` to install the properly signed binaries. They can enroll at install time or by running `elastic-agent enroll` separately later.

cmacknz: Updated, let me know if the new steps are clear enough.

lucabelluccini: Sorry my pedantic questions but as I've not experienced this myself I'm trying to imagine what are the questions a user might have.

At step 2 & 3, will the binary run even if unsigned (to unenroll & uninstall)?
Otherwise LGTM! Thank you for the super clear release note message.

cmacknz: Yes, the binary runs when it is unsigned it just cannot interact with signed applications like endpoint security. 

This problem has no obvious effects until you trigger an action that causes the unsigned elastic-agent application to modify the endpoint-security application, like upgrading to 8.6.2. In addition because of signature caching most of the problems only show up after the first reboot after upgrading to 8.6.1.

Issue #2238, opened 2022-10-10
[Request] Add info on default port assignments


colleenmcginnis:
Related #2178 #2009 

## Description

Figure out where to put information on default port assignments when using Fleet Server on-prem. This was included in the draft provided in #2009, but didn't make it into #2178. 

```asciidoc
== Use default port assignments

When {es} or {fleet-server} are deployed on-premises, communication between certain
components will take place over well defined, pre-allocated ports.
In most cases the operators may need to allow access to these ports.

[options,header]
|====
| Component Communication | Default Port
| Elastic Agent → {fleet-server} | 8220
| Elastic Agent → {es} | 9200
| Elastic Agent → Logstash | 5044
| Elastic Agent → {fleet} | 5601
| {fleet-server} → {fleet} | 5601
| {fleet-server} → {es} | 9200
|====
```

## Collaboration

The docs team will lead producing the content (@colleenmcginnis).

**Contact Person:** @nimarezainia

## Suggested Target Release

This is a general improvement that is not tied to a release.

## Stakeholders

@dedemorton

nimarezainia:
@nrichers moving this to the ingest platform docs planning so it can be addressed as part of our backlog.

nimarezainia:
@dedemorton Could this table be includes in this section: https://www.elastic.co/guide/en/fleet/master/add-a-fleet-server.html#deployed-on-prem  ??

mdbirnstiehl:
@nimarezainia do we see this table as a prerequisite to deploying an on-premise Fleet Server? Will users need to have these ports available before they deploy an on-prem Fleet Server? If this is the case, maybe it would make sense [here](https://www.elastic.co/guide/en/fleet/master/add-fleet-server-on-prem.html#add-fleet-server-on-prem-prereq)?

nimarezainia:
@mdbirnstiehl yes they would be prerequisites. The operator would have to open these ports for both on-prem and cloud deployments. For cloud however they are generally opened for internet access. I don't want to make this just an on-prem specific information.

PR #2725, opened 2023-02-28
Add default port assignment information

mdbirnstiehl:
## Description

Closes [Issue 2238](https://github.com/elastic/observability-docs/issues/2238). 

Adding the default ports to the prerequisite sections of the **Add fleet server** pages. 

I also added some sections in the prereqs for clarity and introductory material for that. 

@nimarezainia you mentioned needing these prerequisites for for both on-prem and cloud deployments, so I have added the table to both pages as well as the mixed deployment page.

### DIFF
--- a/docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
@@ -45,9 +45,19 @@ For more information about hosting {fleet-server} on {ece}, refer to
 [[add-fleet-server-on-prem-prereq]]
 = Prerequisites
 
+Before deploying, you need to:
+
+* Obtain or generate a Cerfiticate Authority (CA) certificate.
+* Ensure components have access to the ports needed for communication.
+
+[discrete]
+[[add-fleet-server-on-prem-cert-prereq]]
+== CA certificate
+
 // tag::cert-prereq[]
+
 Before setting up {fleet-server} using this approach, you will need a
-Certificate Authority (CA) certificate to configure Transport Layer Security (TLS)
+CA certificate to configure Transport Layer Security (TLS)
 to encrypt traffic between the {fleet-server}s and the {stack}.
 
 If your organization already uses the {stack}, you may already have a CA certificate. If you do not have a CA certificate, you can read more
@@ -56,6 +66,28 @@ about generating one in <<secure-connections>>.
 NOTE: This is not required when testing and iterating using the *Quick start* option, but should always be used for production deployments.
 
 // end::cert-prereq[]
+
+[discrete]
+[[default-port-assignments-on-prem]]
+== Default port assignments
+
+//tag::default-port-prereq[]
+
+When {es} or {fleet-server} are deployed, components communicate over well-defined, pre-allocated ports.
+You may need to allow access to these ports. See the following table for default port assignments:
+
+|===
+| Component communication | Default port
+
+| Elastic Agent → {fleet-server} | 8220
+| Elastic Agent → {es} | 9200
+| Elastic Agent → Logstash | 5044
+| Elastic Agent → {fleet} | 5601
+| {fleet-server} → {fleet} | 5601
+| {fleet-server} → {es} | 9200
+|===
+
+// end::default-port-prereq[]
 
 [discrete]
 [[add-fleet-server-on-prem-hosts]]
--- a/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
@@ -39,6 +39,8 @@ NOTE: The TLS certificates used to secure connections between {agent} and
 {fleet-server} are managed by {ecloud}. You do not need to create a private key
 or generate certificates.
 
+include::add-fleet-server-on-prem.asciidoc[tag=default-port-prereq]
+
 [discrete]
 [[add-fleet-server-cloud-set-up]]
 = Setup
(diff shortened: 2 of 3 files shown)


nimarezainia (approved):

kilfoyle (approved):
LGTM! 🚀

docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
```diff
@@ -57,6 +67,28 @@ NOTE: This is not required when testing and iterating using the *Quick start* op
...
+
+| Elastic Agent → {fleet-server} | 8220
+| Elastic Agent → {es} | 9200
+| Elastic Agent → Logstash | 5044
+| Elastic Agent → {fleet} | 5601
+| {fleet-server} → {fleet} | 5601
+| {fleet-server} → {es} | 9200
+|===
+
+// end::default-port-prereq[]
```
karenzone, line 90: I like the way you set this up for reuse. 

Questions to help expand my understanding: 
How did you decide which topic should contain the source content?  Is basing out of on prem and reusing in other topics a standard approach?

dedemorton:
@mdbirnstiehl I think you need a slightly different table for Elastic Cloud. The last time I checked, the port for Fleet Server on cloud was either 443 or 9243.

mdbirnstiehl:
> @mdbirnstiehl I think you need a slightly different table for Elastic Cloud. The last time I checked, the port for Fleet Server on cloud was either 443 or 9243.

Thanks, @dedemorton, I'll verify.

dedemorton:
Also check the other ports because I'm looking at 8.6.1, and the Kibana port is 9243.

docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
mdbirnstiehl: I wish I could give a better response than this, but there was already a tag in the on-prem docs that was acting as source material for cloud, so I just kept with that.

mdbirnstiehl:
@nimarezainia Just want to verify that the ports in this table are accurate for Cloud as well as On-Prem. I've used the same table for both, but DeDe raised the issue that it might not be accurate for Cloud. 

image

nimarezainia:
> @nimarezainia Just want to verify that the ports in this table are accurate for Cloud as well as On-Prem. I've used the same table for both, but DeDe raised the issue that it might not be accurate for Cloud.
> 
> image

@mdbirnstiehl these are look correct. They only apply to on-prem deployments, for cloud everything is on 443 as it connects to the internet.

dedemorton:
@nimarezainia We're trying to nail down the exact details to put in the docs. Can you confirm that this is accurate? I'm speculating about a few of the ports based on my cloud deployment, but not 100% sure.

[Deploy on Elastic Cloud](https://observability-docs_2725.docs-preview.app.elstc.co/guide/en/fleet/master/add-fleet-server-cloud.html)

Component communication | Default port
-- | --
Elastic Agent → Fleet Server | 443 or 9243
Elastic Agent → Elasticsearch | 9243
Elastic Agent → Logstash | 5044
Elastic Agent → Fleet | 9243
Fleet Server → Fleet | 9243
Fleet Server → Elasticsearch | 9243

[Deploy on premises and self-managed](https://observability-docs_2725.docs-preview.app.elstc.co/guide/en/fleet/master/add-fleet-server-on-prem.html)

Component communication | Default port
-- | --
Elastic Agent → {fleet-server} | 8220
Elastic Agent → {es} | 9200
Elastic Agent → Logstash | 5044
Elastic Agent → {fleet} | 5601
{fleet-server} → {fleet} | 5601
{fleet-server} → {es} | 9200


[Deploy Fleet Server on-premises and Elasticsearch on Cloud](https://github.com/elastic/observability-docs/edit/main/docs/en/ingest-management/fleet/add-fleet-server-mixed.asciidoc)

Component communication | Default port
-- | --
Elastic Agent → Fleet Server | 8220
Elastic Agent → Elasticsearch | 9243
Elastic Agent → Logstash | 5044
Elastic Agent → Fleet | 9243
Fleet Server → Fleet | 9243
Fleet Server → Elasticsearch | 9243

nimarezainia:
@dedemorton including @cmacknz and @joshdover to confirm also:

the table for **Self Managed** is correct (I am %99.9 certain. My info is a year old though but these ports wouldn't have changed)

The table for Elastic Cloud, anything connecting to cloud assets over the internet (connecting to Fleet, Fleet Server and ES) would be over 443 (https). 9243 is the same as 443. @cmacknz which port do we use? could you confirm please.

the table for fleet server onprem also looks good to me barring the question regarding the 9243 port.

cmacknz:
> The table for Elastic Cloud, anything connecting to cloud assets over the internet (connecting to Fleet, Fleet Server and ES) would be over 443 (https). 9243 is the same as 443. @cmacknz which port do we use? could you confirm please.

The default endpoints for Elastic Cloud use port 443, this is what I get when I click on "Copy Endpoint" in a deployment and is also what the URL in a Cloud ID decodes to.

I believe we expose both 443 and 9243, both will work. 443 is more likely to be accepted by default in Firewall configurations so I'd favour documenting that as the primary port.

https://www.elastic.co/guide/en/cloud-enterprise/2.1/ece-getting-started-connect.html#ece-getting-started-connect is at least one existing reference to us supporting both 443 and 9243.

joshdover:
One thing that is worth noting with the ES output is that if you _don't_ specify the port as 443, then Agent will default to 9200 which is a bit different than typical handling of `https` URLs.

nimarezainia:
> One thing that is worth noting with the ES output is that if you _don't_ specify the port as 443, then Agent will default to 9200 which is a bit different than typical handling of `https` URLs.

I didn't know this.
@dedemorton @mdbirnstiehl could we add this info where the table for Elastic Cloud resides?

mdbirnstiehl:
I've made updates according to this conversation. @nimarezainia @dedemorton, if you wouldn't mind taking another look, we can get this done and merged!

dedemorton (approved):
LGTM

nimarezainia (approved):

PR #2731, opened 2023-03-08
Enhance service correlation doc

SylvainJuge:
- logs general correlation fields + filebeat setup
- invite users to set service version + env

### Checklist
- [x] ~~should we also recommend setting `service.version` and `environment` in RUM agent too ? That would be consistent with other agents, especially if the environment filter is applied globally.~~ Not for now, the per-agent config is buried and only used in k8s tutorial.
- [x] validate config for Go agent
- [x] validate config for Java agent
- [x] validate config for .net agent
- [x] validate config for Node.js agent
- [x] validate config for Python agent
- [x] validate config for Ruby agent


bmorelli25 (approved):
Thank you! This looks great. Should this be backported to 8.7?

PR #2732, opened 2023-03-08
[Synthetics] Suggest not to use browser types in 7x

paulb-elastic:
Synthetics has changed significantly in 8x and is not a good experience in 7x (and is, and always will be beta in 7x).  This PR makes the suggestion not to use browser type monitors in 7x.

It's the same copy on various pages.


paulb-elastic:
Thanks for the screenshot @colleenmcginnis, I wondered why there was no preview generated
@andrewvc @florent-leborgne you both ok with this going into the 7x docs?

PR #2734, opened 2023-03-08
Clarify limitations around using field names in agent processor configs

dedemorton:
We document this limitation in the intro to processors:

"Cannot process data after it’s been converted to the Elastic Common Schema (ECS) because the conversion is performed by Elasticsearch ingest pipelines. This means that your processor configuration cannot refer to fields that are created by ingest pipelines or Logstash because those fields are created after the processor runs, not before."

But in the spirit of "every page is the first page" we should call out this restriction in any processor docs that require users to refer to field names. Let me know if you think I should add this note to _every_ page.

This is related to https://github.com/elastic/integrations/issues/5481, but not a fix.

### DIFF
--- a/docs/en/ingest-management/processors/processor-add_cloud_metadata.asciidoc
+++ b/docs/en/ingest-management/processors/processor-add_cloud_metadata.asciidoc
@@ -44,6 +44,8 @@ refer to <<provider-specific-examples>>.
 [discrete]
 == Configuration settings
 
+include::processors.asciidoc[tag=processor-limitations]
+
 [options="header"]
 |===
 | Name | Required | Default | Description
--- a/docs/en/ingest-management/processors/processor-add_cloudfoundry_metadata.asciidoc
+++ b/docs/en/ingest-management/processors/processor-add_cloudfoundry_metadata.asciidoc
@@ -53,6 +53,8 @@ settings.
 [discrete]
 == Configuration settings
 
+include::processors.asciidoc[tag=processor-limitations]
+
 [options="header"]
 |===
 | Name | Required | Default | Description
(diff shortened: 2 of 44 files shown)


ycombinator:
I think it's a good idea to include this note in the documentation of every processor, but I'm torn about it's _location_ on each processor's documentation page.  As it stands with this PR, the note will be displayed right under the processor name heading, but before we introduce the processor with a short description of what it does.  I wonder if we should move this note to below that short description or perhaps even further down on the page?  

In other words, I can't decide what the appropriate level of prominence this note deserves on every processor page.  I'd like to think that _most_ users understand that processors that are documented as part of Beats documentation run inside Beats, and therefore before Logstash or ES Ingest Pipelines "see" the data for further processing.  If that's true, then I don't think this note ought to be very prominently placed on every processor's page.

dedemorton:
>As it stands with this PR, the note will be displayed right under the processor name heading, but before we introduce the processor with a short description of what it does.

Fair point. I was taking the easy path here to do it quickly. I think maybe putting this under the Configuration section would be OK.

dedemorton:
@ycombinator Just wanted to point out that the docs I'm updating here are in the Elastic Agent guide. I don't have plans to go back and add this to the Beats documentation.

dedemorton:
@ycombinator Let me know if the new location works? (I overwrote my previous commit and force pushed because it was easier.)

ycombinator:
@dedemorton Yes, I think the note works under Configuration but I wonder if it should only be added to processors that work with existing event fields, e.g. `copy_fields`, `dissect`, `timestamp`, etc.  I don't know that the note makes sense for processors that don't work with existing event fields, e.g. `drop_event`, `add_tags`, `add_locale`, etc.

dedemorton:
@ycombinator Agreed, but I thought from your previous comment that you wanted it in all the docs. I can see some value to that, so I'm going to push a commit that uses some conditional logic to apply different notes depending on whether the processor works with fields.

PR #2746, opened 2023-03-09
Add post-processing step to tidy up API docs section headers

kilfoyle:
This adds a _perl_ step into the API generation README instructions and also updates the API docs output, so that the section headers have spaces between words. It's not perfect: We still have one header reading as `Enrollment APIKeys` rather than `Enrollment API Keys`, but it's better than before. The original `ElasticPackageManagerAPM` now renders as `Elastic Package Manager EPM`.

Ideally we don't want to introduce somewhat arcane code into our docs processes, but this is just a one line command, and these generated docs are essentially a stop-gap fix while we wait for a better implementation.

Note: I used perl so that the command is the same across operating systems.

**Original:**
Screenshot 2023-03-09 at 11 06 12 AM

**Fixed up:**
Screenshot 2023-03-09 at 11 07 16 AM

### DIFF
--- a/docs/en/ingest-management/fleet/api-generated/README.md
+++ b/docs/en/ingest-management/fleet/api-generated/README.md
@@ -6,6 +6,8 @@ Open API specifications (OAS) exist in JSON for Fleet, though they are experimen
 
 A preview of the API specifications can be added to the Fleet and Elastic Agent Guide by using the following process:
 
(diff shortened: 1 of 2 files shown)


kpollich (approved):
Looks great on my end! Thanks for this. 🚀

Issue #2543, opened 2023-01-24
[REQUEST]: The new Inativity Timeout needs to be documented


nimarezainia:
### Description

8.7 brings in a new feature: https://github.com/elastic/kibana/issues/143455

I have created the documentation for this feature:

https://docs.google.com/document/d/1sWXzljzvuIccpreHou-xD7STo6nSpy6vWhDVzdVI6jw/edit?usp=sharing

### Resources

https://docs.google.com/document/d/1sWXzljzvuIccpreHou-xD7STo6nSpy6vWhDVzdVI6jw/edit?usp=sharing

### Collaboration

The product team will provide the initial content.

### Point of contact.

**Main contact:** @nimarezainia 

**Stakeholders:**
@kylepol

nimarezainia:
@dedemorton all the data is in the document attached. I placed this on the radar column but needs to be prioritized for 8.7

dedemorton:
@nimarezainia Might be better to add your details (and link) to the existing issue here: https://github.com/elastic/observability-docs/issues/2352. Otherwise feels like we are double tracking....

nimarezainia:
@dedemorton i can't add the doc which is private to elastic to a public issue. Well technically I can but will point the reader to something they can't access.

dedemorton:
Yah I understand. The duplicate issues become kind of confusing though. I've started adding the link and just explaining that it's internal only. I don't know if there is a company wide standard for how to do this, but double tracking just seems inefficient and error-prone.

bmorelli25:
Closing this as it looks to be done.

Issue #2352, opened 2022-11-08
[Fleet] Replace "Unenrollment timeout" with "Agent inactivity timeout"


kpollich:
In https://github.com/elastic/kibana/issues/143455, we'll be replacing the concept of Fleet's "Unenrollment timeout" with a less destructive "Agent inactivity timeout". This is a placeholder issue to capture the docs work required to update our public facing documentation with

- The wording change for this setting
- Functionality changes associated with the new concept of "inactive agents" vs the previous "automatic unenrollment" process

Will flesh this out with more information as the feature is worked on. 🙂

jlind23:
@kpollich anything missing here? Looks like @hop-dev PR landed recently so we should be good to go.

hop-dev:
@kpollich Added a summary to the issue description 👍

jlind23:
@ollyhowell Not sure what the process is but could you please make sure that someone picks it up?

jlind23:
Removing @hop-dev to have someone from docs taking it over from there.

dedemorton:
There's a related issue with a google doc containing proposed changes from the PM team here: https://github.com/elastic/observability-docs/issues/2543

jlind23:
@nimarezainia Any reason why a new one has been created? If not, I would be happy to merge both.

nimarezainia:
@dedemorton @jlind23 I created the docs issue and put the link to documentation there. I can't link private docs to a public issue, unless there's a better way.

bmorelli25:
Hi @kpollich is this still blocked?

jlind23:
@bmorelli25 this is not blocked anymore!

dikshachauhan-qasource:
Hi @kpollich 

 We have created **08** testcases for this feature under our Fleet test suite at links:
- [Validate that new filter options ‘Inactive’ and ‘Unenrolled’ agents are available on the ‘Status’ dropdown on the Agent list page.](https://elastic.testrail.io/index.php?/cases/view/168962)
- [Validate that "Unenrollment Timeout" field is now deprecated and replaced setting as "Agent inactivity timeout" is available on the Policy Settings page](https://elastic.testrail.io/index.php?/cases/view/168963)
- [Validate that agent gets "Inactive" after specified time limit mentioned at "Inactivity timeout" field under Policies>settings tab is crossed.](https://elastic.testrail.io/index.php?/cases/view/168964)
[...]

PR #2747, opened 2023-03-09
Add inactivity timeout documentation

mdbirnstiehl:
Closes issues [2352](https://github.com/elastic/observability-docs/issues/2352) and [2543](https://github.com/elastic/observability-docs/issues/2543).

### Summary

I've added the **Set inactivity timeout** section and updated information on statuses and links regarding setting the unenrollment timeout. 

@nimarezainia In the [suggested documentation](https://docs.google.com/document/d/1sWXzljzvuIccpreHou-xD7STo6nSpy6vWhDVzdVI6jw/edit), there was some additional information on agent statuses and lifecycle. Some of that information is [here](https://www.elastic.co/guide/en/fleet/8.6/monitor-elastic-agent.html#view-agent-status), which I have linked to from the **Set inactivity timeout** doc. Maybe we can update the information there to explain that better. Let me know if this is ok.

I am also wondering if we still want to document the unenrollment timeout as deprecated since it's still in the UI. Not with a full page on setting it, but a brief mention somewhere.

### DIFF
--- a/docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
+++ b/docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
@@ -1,9 +1,5 @@
 [[set-inactivity-timeout]]
-= Set inactivity timeout for ephemeral hosts
-
-++++
-<titleabbrev>Set inactivity timeout</titleabbrev>
-++++
+= Set inactivity timeout
 
 [[why-set-inactivity-timeout]]
 == Why set an inactivity timeout?
@@ -14,10 +10,10 @@ declutter the {fleet} UI.
 
 Once {fleet-server} receives a check-in from an inactive {agent}, it returns to healthy status. 
 
-For example, if an employee is on holiday with their laptop off while they're away, 
+For example, if an employee is on holiday with their laptop off, 
 the {agent} will transition to offline then inactive once the inactivity timeout limit is reached. 
 This prevents the inactive {agent} from cluttering the {fleet} UI, and,
-onnce the employee returns, the {agent} checks in and returns to healthy status with valid API keys to use. 
+once the employee returns, the {agent} checks in and returns to healthy status with valid API keys. 
 
 If an {agent} is no longer valid, you can manually <<unenroll-elastic-agent,unenroll>> inactive {agent}s to revoke the API keys. 
 Unenrolled agents need to be re-enrolled to be operational again.
--- a/docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
+++ b/docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
@@ -1,0 +1,41 @@
+[[set-inactivity-timeout]]
+= Set inactivity timeout for ephemeral hosts
+
+++++
+<titleabbrev>Set inactivity timeout</titleabbrev>
+++++
+
+[[why-set-inactivity-timeout]]
+== Why set an inactivity timeout?
+
+The inactivity timeout moves {agent}s to *inactive* status after being offline for a set amount of time. 
+Inactive {agent}s are still valid {agent}s, but are removed from the main {fleet} UI allowing you to better manage {agent}s and 
+declutter the {fleet} UI.
+
+Once {fleet-server} receives a check-in from an inactive {agent}, it returns to healthy status. 
+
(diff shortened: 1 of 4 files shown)


docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
```diff
@@ -0,0 +1,35 @@
...
+The inactivity timeout moves {agent}s to inactive status after being offline for a set amount of time. 
+Inactive {agent}s are still valid {agent}s, but are removed from the main {fleet} UI allowing you to better manage {agent}s and 
+declutter the {fleet} UI.
+
+Once {fleet-server} receives a check-in from an inactive {agent}, it returns to healthy status. 
+
+For example, if an employee is on holiday with their laptop off, 
+the {agent} will transition to offline then inactive once the inactivity timeout limit is reached. 
+This prevents the inactive {agent} from cluttering the {fleet} UI.
+Once the employee returns, the {agent} checks in and returns to healthy status with valid API keys.
```
dedemorton, line 16:
```suggestion
When {fleet-server} receives a check-in from an inactive {agent}, it returns to healthy status. 

For example, if an employee is on holiday with their laptop off, 
the {agent} will transition to offline then inactive when the inactivity timeout limit is reached. 
This prevents the inactive {agent} from cluttering the {fleet} UI.
When the employee returns, the {agent} checks in and returns to healthy status with valid API keys. 

```
Changing once > when because I've always been told by editors to avoid "once" because it might cause ambiguity during translation (not that we translate, but....). After or when are preferred.

docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
```diff
@@ -0,0 +1,35 @@
...
+If an {agent} is no longer valid, you can manually <<unenroll-elastic-agent,unenroll>> inactive {agent}s to revoke the API keys. 
+Unenrolled agents need to be re-enrolled to be operational again.
+
+For more on {agent} statuses, see <<view-agent-status, view agent status>>.
+
+
+[[setting-inactivity-timeout]]
+== Set the inactivity timeout
+
+Set the inactivity timeout in the {agent} policy to the amount of time after which you want an offline {agent} become inactive.
```
dedemorton, line 27:
```suggestion
Set the inactivity timeout in the {agent} policy to the amount of time after which you want an offline {agent} to become inactive.

```

docs/en/ingest-management/fleet/set-inactivity-timeout.asciidoc
```diff
@@ -0,0 +1,35 @@
+[[set-inactivity-timeout]]
+= Set inactivity timeout
+
+[[why-set-inactivity-timeout]]
+== Why set an inactivity timeout?
```
dedemorton, line 5:
```suggestion
```
Not sure if this ID is referenced somewhere else, but I don't think we need this heading here. It just sort of repeats the topic title as a question. Also the "why" is a bit buried in the paragraph. I would just delete this.

dedemorton (commented):
Just a few minor comments. Looks great!

dedemorton (approved):
LGTM. BTW, I fact checked myself on "once" vs "when". Sounds like once is "more emphatic." The translation ambiguity is probably not as much of a problem these days because machine translation has gotten so much better. But the edits work and are consistent with other topics in this book, so it's all good. :-)

nimarezainia:
> @nimarezainia In the [suggested documentation](https://docs.google.com/document/d/1sWXzljzvuIccpreHou-xD7STo6nSpy6vWhDVzdVI6jw/edit), there was some additional information on agent statuses and lifecycle. Some of that information is [here](https://www.elastic.co/guide/en/fleet/8.6/monitor-elastic-agent.html#view-agent-status), which I have linked to from the **Set inactivity timeout** doc. Maybe we can update the information there to explain that better. Let me know if this is ok.

@mdbirnstiehl the state changes mentioned in the doc are valid post 8.7 so IMO they should completely replace what you pointed to i the docs. that is, there are newer states introduced and slight modifications to the transitions between each.

Do you think we could just use the diagram that shows the transitions? I feel that a pictorial presentation has a lot more impact. lmk if you want me to redraw.

nimarezainia (approved):

PR #2774, opened 2023-03-21
Update upgrade-elastic-agent.asciidoc

xtruthx:
In the document there is written that the "Upgrade Agent" function in Fleet is grayed out if the kibana server version is lower than the Agent version. The kibana version must be higher than the agent version.

Kibana version must be higher so that option is shown. The original description is not really clear and confusing. IMHO

### DIFF
--- a/docs/en/ingest-management/fleet/upgrade-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/fleet/upgrade-elastic-agent.asciidoc
@@ -69,7 +69,7 @@ that you can upgrade.
 image::images/upgrade-single-agent.png[Menu for upgrading a single {agent}]
 +
 The **Upgrade agent** option is grayed out if an upgrade is unavailable or
-the {kib} version is lower than the agent version.
+the {kib} version is not higher or equal than the agent version.
 
 . In the Upgrade agent window, select an upgrade version and click
 **Upgrade agent**.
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet/upgrade-elastic-agent.asciidoc
```diff
@@ -69,7 +69,7 @@ that you can upgrade.
 image::images/upgrade-single-agent.png[Menu for upgrading a single {agent}]
 +
 The **Upgrade agent** option is grayed out if an upgrade is unavailable or
-the {kib} version is lower than the agent version.
+the {kib} version is not higher or equal than the agent version.
```
kilfoyle, line 72:
```suggestion
the {kib} version is below the agent version.
```
I think `below` is equivalent but a easier to parse than `not higher or equal than`.

xtruthx: IMHO `below`  is not enough. In your case the kibana version can be equal the agent version and the option is grayed out.

Suggestion:
`the {kib} version is below or equal to the agent version.`

kilfoyle, line 72:
```suggestion
the {kib} version is below or equal to the agent version.
```

kilfoyle: Sounds good to me! Thanks.

kilfoyle:
Thanks @xtruthx! I see it passed the CLA check now. I'll merge your suggested change and hopefully the other CI checks will all turn green.

Issue #2603, opened 2023-02-09
[Synthetics] Document that not all Playwright capabilities are available in Synthetics


paulb-elastic:
### Description

We’ve had some confusion (internally and externally) when folk take a Playwright script example and it fails to run in Synthetics.  This has been because we do not support the Playwright Test features (as we have our own runner in Synthetics).

For example, https://playwright.dev/docs/api/class-locatorassertions documents things like [`toHaveText`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-text) assertions.  However these will not work in Synthetics as they are provided by Playwright Test.

I've marked this for 8.7, although it doesn't specifically relate to that release, and it wouldn't matter if it wasn’t immediately ready on release day, but came out shortly after.

### Resources

I think the Project Team (namely @vigneshshanmugam) will need to provide the content, with help from the Docs team on the best place to put it.

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** @vigneshshanmugam 

**Stakeholders:**
* @vigneshshanmugam 
* @andrewvc 
* @paulb-elastic

colleenmcginnis:
>This has been because we do not support the Playwright Test features (as we have our own runner in Synthetics).

>I think the Project Team (namely @vigneshshanmugam) will need to provide the content, with help from the Docs team on the best place to put it.

@vigneshshanmugam can you help point me in the right direction here?

Based on browsing through the /synthetics repo, it looks like maybe we support _most_ of the Playwright [GenericAssertions](https://playwright.dev/docs/api/class-genericassertions) in [`expect.ts`](https://github.com/elastic/synthetics/blob/main/src/core/expect.ts), but we might not support any of the [LocatorAssertions](https://playwright.dev/docs/api/class-locatorassertions), [PageAssertions](https://playwright.dev/docs/api/class-pageassertions), or [SnapshotAssertions](https://playwright.dev/docs/api/class-snapshotassertions). Is that true? Is there somewhere else I should be looking?

vigneshshanmugam:
> Based on browsing through the /synthetics repo, it looks like maybe we support most of the Playwright [GenericAssertions](https://playwright.dev/docs/api/class-genericassertions) in [expect.ts](https://github.com/elastic/synthetics/blob/main/src/core/expect.ts)

[...]

PR #2792, opened 2023-03-24
[synthetics] Document that not all Playwright functionality is available in Synthetics

colleenmcginnis:
Closes https://github.com/elastic/observability-docs/issues/2603

## Approach

After some discussion in #2603, I've approached this more like _documenting the relationship between Synthetics and Playwright_ rather than just documenting that not all Playwright functionality is available in Synthetics. Here's what I've done so far:

* Added `expect` to the existing [Syntax overview](https://observability-docs_2792.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-create-test.html#synthetics-syntax) section to highlight that this functionality is built into Synthetics. 
* Added a new [Playwright syntax](https://observability-docs_2792.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-create-test.html#synthetics-playwright) section to the existing [Add steps](https://observability-docs_2792.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-create-test.html#synthetics-create-step) section. This new section:
  * Explains the relationship between Synthetics and Playwright using a few examples of commonly used Playwright capabilities (for finding elements on a page and simulating user interaction).
  * Calls out Playwright capabilities that are definitely _not_ supported in Synthetics (for example, videos).
  * Calls out functionality that is available in Playwright, but is not supported/recommended when using Playwright with Synthetics because there is an alternative built into Synthetics (for example, `expect` and `request`).
* Split the content on using `expect` in a `step` into its own section, [Make assertions](https://observability-docs_2792.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-create-test.html#synthetics-make-assertions) in order to provide more detail.
  * The current draft also includes a reference of all the `expect` methods, but I'm not sure if this is necessary based on the discussion in [#2603](https://github.com/elastic/observability-docs/issues/2603) (since this information could be included in a future generated reference). I'm interested in @vigneshshanmugam and @paulb-elastic's thoughts on this.

## To do

- [x] Get feedback on the approach from @vigneshshanmugam and @paulb-elastic
- [x] If needed, fill in missing `Description...` placeholders
- [ ] Full review from @vigneshshanmugam and/or @paulb-elastic
- [ ] Address feedback and publish

### DIFF
--- a/docs/en/observability/synthetics-create-test.asciidoc
+++ b/docs/en/observability/synthetics-create-test.asciidoc
@@ -224,149 +224,35 @@ expected value. See https://jestjs.io/docs/expect[Jest expect docs] for more inf
 
 [discrete]
 [[synthetics-assertions-methods]]
-== Methods
+== Supported `expect` methods
 
-[cols="1,2"]
-|===
-| *`not()`*
-| If you know how to test something, `.not` lets you test its opposite.
+:jest: https://jestjs.io/docs
 
-| *`resolves()`*
-| Use resolves to unwrap the value of a fulfilled promise so any other matcher can be chained. If the promise is rejected the assertion fails.
+* {jest}/expect#not[`not ()`]
+* {jest}/expect#resolves[`resolves ()`]
+* {jest}/expect#rejects[`rejects ()`]
+* {jest}/expect#tobevalue[`toBe (expected)`]
+* {jest}/expect#tobeclosetonumber-numdigits[`toBeCloseTo (expected, numDigits?)`]
+* {jest}/expect#tobedefined[`toBeDefined ()`]
+* {jest}/expect#tobefalsy[`toBeFalsy ()`]
+* {jest}/expect#tobegreaterthannumber\--bigint[`toBeGreaterThan (expected)`]
+* {jest}/expect#tobegreaterthanorequalnumber\--bigint[`toBeGreaterThanOrEqual (expected)`]
+* {jest}/expect#tobeinstanceofclass[`toBeInstanceOf (expected)`]
+* {jest}/expect#tobelessthannumber\--bigint[`toBeLessThan (expected)`]
+* {jest}/expect#tobelessthanorequalnumber\--bigint[`toBeLessThanOrEqual (expected)`]
+* {jest}/expect#tobenull[`toBeNull ()`]
+* {jest}/expect#tobetruthy[`toBeTruthy ()`]
+* {jest}/expect#tobeundefined[`toBeUndefined ()`]
+* {jest}/expect#tobenan[`toBeNaN ()`]
+* {jest}/expect#tocontainitem[`toContain (expected)`]
+* {jest}/expect#tocontainequalitem[`toContainEqual (expected)`]
+* {jest}/expect#toequalvalue[`toEqual (expected)`]
+* {jest}/expect#tohavelengthnumber[`toHaveLength (expected)`]
+* {jest}/expect#tohavepropertykeypath-value[`toHaveProperty (keyPath, value?)`]
+* {jest}/expect#tomatchregexp\--string[`toMatch (expected)`]
+* {jest}/expect#tomatchobjectobject[`toMatchObject (expected)`]
+* {jest}/expect#tostrictequalvalue[`toStrictEqual (expected)`]
 
-| *`rejects()`*
-| Unwraps the reason of a rejected promise so any other matcher can be chained. If the promise is fulfilled the assertion fails.
-
-| *`toBe(expected)`*
-a| Checks that a value is what you expect. It uses `===` to check strict equality. Don't use `toBe` with floating-point numbers.
-
-*Parameters*
-
-* *`expected`* (_any_) +++<span style="color:#f04e98">Description...</span>+++
-
-| *`toBeCloseTo (expected, numDigits?)`*
-a| Using exact equality with floating point numbers is a bad idea. Rounding means that intuitive things fail.
-
-*Parameters*
-
-* *`expected`* (_number_) The expected value.
-* *`numDigits`* (_number_?) The number of digits.
-
-| *`toBeDefined ()`*
-| Ensure that a variable is not undefined.
-
-| *`toBeFalsy ()`*
-| +++<span style="color:#f04e98">Description...</span>+++.
-
-| *`toBeGreaterThan (expected)`*
-a| +++<span style="color:#f04e98">Description...</span>+++.
-
-*Parameters*:
-
-* *`expected`* (_any_) +++<span style="color:#f04e98">Description...</span>+++
-
-| *`toBeGreaterThanOrEqual (expected)`*
-a| +++<span style="color:#f04e98">Description...</span>+++.
-
-*Parameters*:
-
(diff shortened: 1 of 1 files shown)


paulb-elastic:
@colleenmcginnis the build seems to have finished but the [links](https://github.com/elastic/observability-docs/pull/2792#issuecomment-1483248724) aren't working (which would be useful in this context), thanks.

colleenmcginnis:
>@colleenmcginnis the build seems to have finished but the https://github.com/elastic/observability-docs/pull/2792#issuecomment-1483248724 aren't working (which would be useful in this context), thanks.

We're experiencing some issues with the preview generation process. I'll let you know when it's back up!

colleenmcginnis:

colleenmcginnis:

colleenmcginnis:
@paulb-elastic all the preview links in the original post should work now! Let me know if you have any issues.

docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -147,16 +222,151 @@ See the https://playwright.dev/docs/api/class-page#page-get-attribute[`page.getA
 <2> Use the assertion library provided by the Synthetics agent to look for the
 expected value. See https://jestjs.io/docs/expect[Jest expect docs] for more information.
 
-[[synthetics-create-test-script-recorder]]
-[NOTE]
-====
-If you want to generate code by interacting with a web page directly, you can use the *Synthetics Recorder*.
+[discrete]
+[[synthetics-assertions-methods]]
+== Methods
```
vigneshshanmugam, line 227: We use the expect library from the Jest framework which itself is a standalone package. We could also point users to the website - https://jestjs.io/docs/expect. TLDR: Playwright used the same library before and recently enhanced it to support custom assertions which we clearly dont support leading to the whole confusion. 

One more idea would be to actually include only the subset of methods that we support and link the docs directly to the expect library instead. 

Thoughts?

colleenmcginnis: >We could also point users to the website - https://jestjs.io/docs/expect. 

At first glance it looks like not all the methods listed in the Jest docs are included in [`expect.ts`](https://github.com/elastic/synthetics/blob/main/src/core/expect.ts). Does that mean not all the Jest methods are supported in Synthetics? If so, could linking to the Jest site just cause the same kind of confusion we're seeing with linking to Playwright assertions? 

>One more idea would be to actually include only the subset of methods that we support and link the docs directly to the expect library instead.

When you say "link the docs directly to the expect library" do you mean the Jest site or somewhere else?

vigneshshanmugam: > If so, could linking to the Jest site just cause the same kind of confusion we're seeing with linking to Playwright assertions?

Yeah thats a possibility, We didnt expose that as some of the matchers in expect does not make sense for Synthetics. 

> When you say "link the docs directly to the expect library" do you mean the Jest site or somewhere else?

Yeah directly to the jest site for each supported methods. 

I don't have any preference on whether we should add them here vs referencing the users directly there and clearly documentating that we support only partial of them. Thoughts @paulb-elastic @andrewvc

colleenmcginnis: >I don't have any preference on whether we should add them here vs referencing the users directly there and clearly documentating that we support only partial of them.

Curious what Paul and Andrew think, but I'm thinking a full list of the `expect` methods that are supported in Synthetics where each list item is just a link to the relevant part of the Jest docs makes sense to me. That seems like a good compromise -- not too much content that might become difficult to maintain, but also clarity on what is/isn't supported.

paulb-elastic: @colleenmcginnis I 100% agree, I think this is a good compromise.
I am concerned just linking won't be clear exactly what is and isn't supported, or expected to work.  I can see this being a useful reference for us/Support as well as for users.

colleenmcginnis:
@paulb-elastic @vigneshshanmugam I updated the preview based on the conversation above. Would one (or both) of you be able to do a full review? Here's the [updated preview](https://observability-docs_2792.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-create-test.html). Thanks!

paulb-elastic (approved):
Thanks @colleenmcginnis this looks great, I'll LGTM, but defer to @vigneshshanmugam to get the final approval before merging.

docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -25,6 +25,8 @@ The synthetics agent exposes an API for creating and running tests, including:
 `step`::        Actions within a journey that should be completed in a specific order.
                 Takes two parameters: a `name` (string) and a `callback` (function).
                 Learn more in <<synthetics-create-step>>.
+`expect`::      Check that a value meet a specific condition. There are several supported checks.
+                Learn more in <<synthetics-make-assertions>>.
```
colleenmcginnis, line 29: Typo 🙈  

(Reminder to myself to fix before merging.)

```suggestion
`expect`::      Check that a value meets a specific condition. There are several supported checks.
                Learn more in <<synthetics-make-assertions>>.
```

docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -112,7 +126,70 @@ step('Load the demo page', () => {
...
+* Finding elements on a web page using https://playwright.dev/docs/api/class-locator[locators].
+* Simulating https://playwright.dev/docs/api/class-mouse[mouse],
+  https://playwright.dev/docs/api/class-touchscreen[touch], or
+  https://playwright.dev/docs/api/class-keyboard[keyboard] events.
+
+Visit the https://playwright.dev/docs[Playwright documentation] for information.
+
+[NOTE]
+====
+Playwright functionality that is not supported in Elastic Synthetics includes:
```
vigneshshanmugam, line 174: We might need to mention it as `Playwright functionality that is not supported out of the box`, as users can still access these by using different means via Synthetics ex: video and API request.

vigneshshanmugam (approved):
LGTM @colleenmcginnis 🎉 

can we update the syntax example in the Make assertions section to be
```ts
  const input = await page.locator('input.new-todo');
    expect(await input.getAttribute('placeholder')).toBe(
      'What needs to be done?'
    );
```

docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -112,7 +126,70 @@ step('Load the demo page', () => {
...
+====
+
+The Elastic Synthetics library has alternatives to some Playwright functionality.
+The alternatives are designed to work better for synthetic monitoring.
+Do not use Playwright syntax to:
+
+* *Make assertions*. Use Elastic Synthetics's `expect` methods instead.
+  Read more in <<synthetics-make-assertions>>.
+* *Request parameter for the API tests.* Use Elastic Synthetic's `request`
+  parameter instead. Read more in <<synthetics-request-param>>.
```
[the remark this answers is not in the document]
vigneshshanmugam: ++ I feel this is better than original wording and clearly mentions alternatives and what are not supported out of the box.

PR #2811, opened 2023-03-30
Need to explicityly mention why agent is running as a super user

nimarezainia:


docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -1,7 +1,8 @@
 [[elastic-agent-installation]]
 = Install {agent}s
 
-NOTE: You can install only a single {agent} per host.
+NOTE: You can install only a single {agent} per host. Due to the fact that the {agent} may read data sources that 
+are only accessible by a superuser, {agent} will therefore also need to be executed with super user permissions.
```
kilfoyle, line 5:
```suggestion
are only accessible by a superuser, {agent} will therefore also need to be executed with superuser permissions.
```

kilfoyle (approved):
LGTM! 🚀 
Just one super small suggestion.

PR #2819, opened 2023-03-31
added ~ to ingest rate  [universal-profiling.asciidoc]

iogbole:


docs/en/observability/universal-profiling.asciidoc
```diff
@@ -213,7 +213,7 @@ In this spirit, both the host-agent and storage are engineered to use as little
 === {es} storage
 
 The Universal Profiling storage budget is predictable on a per-profiled-core basis. The data we generate, at the fixed
-sampling frequency of 20 Hz, will be stored in {es} at the rate of 40 MB per core per day.
+sampling frequency of 20 Hz, will be stored in {es} at the rate of ~40 MB per core per day.
```
mdbirnstiehl, line 216:
```suggestion
sampling frequency of 20 Hz, will be stored in {es} at the rate of approximately 40 MB per core per day.

```

I would just write out approximately here to be clear. Otherwise, LGTM.

PR #2844, opened 2023-04-05
Add note on location of proxy settings for artifact download

karenzone:
Related: 
* https://github.com/elastic/ingest-docs/issues/69
* https://github.com/elastic/ingest-docs/issues/75

@cmacknz: Something like this? 

**PREVIEW:** https://observability-docs_2844.docs-preview.app.elstc.co/guide/en/fleet/master/fleet-agent-proxy-support.html

### DIFF
--- a/docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
+++ b/docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
@@ -27,7 +27,7 @@ Configure proxy settings for {agent} when it must connect through a proxy server
 to:
 
 * Download artifacts from `artifacts.elastic.co` for subprocesses or binary
-upgrades
+upgrades (use <<fleet-agent-binary-download-settings>>)
 * Send data to {es}
 * Retrieve agent policies from {fleet-server}
 * Retrieve agent policies from {es} (only needed for agents running {fleet-server})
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -132,6 +132,8 @@ HTTP_PROXY=http://my.proxy:8080
 
 After adding environment variables, restart the service.
 
+NOTE: If you use a proxy server to download artifacts from `artifacts.elastic.co`, configure <<fleet-agent-binary-download-settings>>.
```
cmacknz, line 135:
```suggestion
NOTE: If you use a proxy server to download artifacts from `artifacts.elastic.co`, configure <<fleet-agent-binary-download-settings>>. `artifacts.elastic.co` is used to download new versions of the agent when upgrading.
```

LGTM, but I think we should mention that artifacts.elastic.co is required to download upgrades so users know why they would need to configure this.

Feel free to reword this.

cmacknz: Actually this might be handled above with `Download artifacts from `artifacts.elastic.co` for subprocesses or binary upgrades` but it is a bit far away from this note

cmacknz (approved):

PR #2846, opened 2023-04-06
Add docs for Fleet UI FQDN setting

kilfoyle:
This adds docs for the new Host Format selector option added via https://github.com/elastic/kibana/pull/154563

Closes: https://github.com/elastic/ingest-dev/issues/1747

**Preview**

Screenshot 2023-04-06 at 11 22 31 AM


hop-dev (approved):
LGTM, is it worth mentioning that Hostname is the default?

kilfoyle:
Good idea @hop-dev  (and that was fast!). 
I've added `This is the default reporting format.` for the Hostname option and I've updated the screen capture in the description.

PR #2849, opened 2023-04-09
Sensitive information is already being redacted from diagnostics

nimarezainia:
the troubleshooting guide needs to be updated due to this merge: https://github.com/elastic/elastic-agent/pull/566

### DIFF
--- a/docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
+++ b/docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
@@ -309,6 +309,9 @@ This archive collects the following information:
 
 Note that the diagnostics bundle is intended for debugging purposes only, its structure may change between releases.
 
+IMPORTANT: {agent} attempts to automatically redact credentials and API keys when creating diagnostics. 
+Please review the contents of the archive before sharing to ensure there are no credentials in plain text.
+
 [discrete]
 [[not-installing-no-logs-in-terminal]]
 == Some problems occur so early that insufficient logging is available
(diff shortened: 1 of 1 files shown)


bmorelli25:
We just moved the ingest documentation to [elastic/ingest-docs](https://github.com/elastic/ingest-docs). Once we hear back from @cmacknz, I'll handle opening a new PR and backporting.

bmorelli25:
@cmacknz can you confirm if this is still an issue

docs/en/ingest-management/troubleshooting/troubleshooting.asciidoc
```diff
@@ -309,9 +309,6 @@ This archive collects the following information:
 
 Note that the diagnostics bundle is intended for debugging purposes only, its structure may change between releases.
 
-IMPORTANT: The configuration files in the archive may have credentials in *plain text*.
-These will need to be manually redacted before the archive is shared.
```
cmacknz, line 313: > IMPORTANT: {agent} attempts to automatically redact credentials and API keys when creating diagnostics. Please review the contents of the archive before sharing to ensure there are no credentials in plain text.

We have an algorithm for automatically redacting credentials, but it is best effort and may miss things. Rather than remove this I would rather document that we attempt to do the redaction but not to rely on it completely. For example, there have been bugs in this algorithm before.

nimarezainia:
@bmorelli25 I made the recommended changes. hope this doesn't conflict with what you suggested [here](https://github.com/elastic/observability-docs/pull/2849#issuecomment-1506021201)

cmacknz (approved):

bmorelli25:
Thanks, both! I've cherry-picked Nima's commits to the new `ingest-docs` repository. I'm going to close this PR in favor of https://github.com/elastic/ingest-docs/pull/190.

Issue #2818, opened 2023-03-31
[Synthetics] Best practice guide for Synthetics


paulb-elastic:
### Description

Might be good to start working on this after 8.8 has gone out (prioritising the other 8.8 docs tasks for now)

Different but linking to enhancing the troubleshooting page https://github.com/elastic/observability-docs/issues/2695

It would be useful to create a best practice guide for Synthetics.  This has come up a few times where users and the field have asked for guidance on pitfalls or other suggestions we can help them with.  We already have the [troubleshooting page](https://www.elastic.co/guide/en/observability/current/troubleshoot-uptime-mapping-issues.html) but probably want something else, more focused on best practices.

I expect this will be something we’ll want to update ad hoc as we and users learn and ask for different scenario advice.

### Resources

This will need to be a team wide collaboration as I’m sure we’ll all have ideas on what to include.

Some very initial suggestions as a starting point:

* Aware of geographical differences on websites
  * Language, content, behaviour can be different from the website when delivered to different regions (e.g. cookie banners, but much more)
  * Particularly pertinent when recording a script in one region and running it another, or expecting a single script to work in multiple regions
* The Synthetics Recorder is most useful for creating a skeleton for a script, to help build the actions and assertions and step separation
  * Although this may suffice for a simpler website or journey, additional scripting or logic it’s often needed to make the script more reliable and less brittle
  * For example, selecting a date on a ticket purchasing website should not typically use a fixed date, as over time, that date will become invalid as it’s in the past.  It may be more appropriate to dynamically generate the date to select (e.g. today + 60 days)
* Avoid repeating the same code for common actions used across multiple scripts
  * As with other development disciplines, consider the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) when writing multiple scripts
  * For example, you may have a login step that’s needed across multiple scripts.  Consider using [Project Monitors](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html) and separating out the login logic into a separate code block.
  * See an example in our [Synthetics Demo](https://github.com/elastic/synthetics-demo) repo where journeys in [advanced-example.journey.ts](https://github.com/elastic/synthetics-demo/blob/main/todos/journeys/advanced-example.journey.ts) call the `loadAppStep` method (and others) which are defined in [advanced-example-helpers.ts](https://github.com/elastic/synthetics-demo/blob/main/todos/journeys/advanced-example-helpers.ts)
[...]

PR #2866, opened 2023-04-20
Set expectations for the Synthetics Recorder

paulb-elastic:
Set some expectations for how the Recorder should be considered when using.  Whilst we have https://github.com/elastic/observability-docs/issues/2818 to add a more detailed _best practice_ guide, this is to add some expectations directly in the recorder docs.

### DIFF
--- a/docs/en/observability/synthetics-recorder.asciidoc
+++ b/docs/en/observability/synthetics-recorder.asciidoc
@@ -3,6 +3,8 @@
 
 beta[]
 
+IMPORTANT: As with any script recording technology, the Elastic Synthetics Recorder should be thought of as a tool to help create the main structure of the script. For simpler sites it may well be enough to have created the final script, but for more complex and dynamic sites, or to limit flakiness, some further scripting may be needed.
+
 You can use the Synthetics Recorder to <<synthetics-create-test, write a synthetic test>> by interacting with a web page and exporting journey code that reflects all the actions you took.
 
 image::images/synthetics-create-test-script-recorder.png[Elastic Synthetics Recorder after recording a journey and clicking Export]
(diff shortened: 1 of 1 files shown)


docs/en/observability/synthetics-recorder.asciidoc
```diff
@@ -3,6 +3,8 @@
 
 beta[]
 
+IMPORTANT: As with any script recording technology, the Elastic Synthetics Recorder should be thought of as a tool to help create the main structure of the script. For simpler sites it may well be enough to have created the final script, but for more complex and dynamic sites, or to limit flakiness, some further scripting may be needed.
```
colleenmcginnis, line 6: What do you think about something like this?

```suggestion
IMPORTANT: As with any script recording technology, the Elastic Synthetics Recorder should be used as a tool to help create the main structure of the script. For simpler sites, you may be able to use the Synthetics Recorder's output directly to create a synthetic monitor, but for more complex and dynamic sites, or to limit flakiness, you'll likely need to edit its output before using it to create a monitor.
```

colleenmcginnis (commented):
Should we backport this all the way back to when the Synthetics Recorder docs were added?

paulb-elastic:
> Should we backport this all the way back to when the Synthetics Recorder docs were added?

@colleenmcginnis I did wonder this, but then didn't know how much folks would be looking at the older version of the docs for the recorder specifically (as it doesn't align to the stack releases).

Taking a look, this page is there from 8.3.  Happy to take your lead here if you think that would be more consistent.  Does this just need the backport labels adding for each of those version?

colleenmcginnis:
>Taking a look, this page is there from 8.3. Happy to take your lead here if you think that would be more consistent. Does this just need the backport labels adding for each of those version?

Yep! Just add the labels for each version.

PR #2867, opened 2023-04-20
Clarify when monitors run after creation

paulb-elastic:
We’re clarifying expectation in the UI with [kibana#155307](https://github.com/elastic/kibana/issues/155307) but this adds some details to the docs too

### DIFF
--- a/docs/en/observability/synthetics-get-started-project.asciidoc
+++ b/docs/en/observability/synthetics-get-started-project.asciidoc
@@ -195,6 +195,12 @@ with the {kib} URL for the deployment from which to fetch available locations.
 [discrete]
 = View in {kib}
 
+
+[NOTE]
+====
+When a monitor has been created (and also when edited), it will not run immediately but at some point after. After the first run occurs, the monitor will then run regularly based on the frequency configured for the monitor (e.g. every 10 minutes). You can run a manual test if you want to see the results more quickly.
+====
+
 Then, go to the {synthetics-app} in {kib}. You should see your newly pushed monitors running.
 You can also go to the *Management* tab to see the monitors' configuration settings.
 
--- a/docs/en/observability/synthetics-get-started-project.asciidoc
+++ b/docs/en/observability/synthetics-get-started-project.asciidoc
@@ -198,7 +198,7 @@ with the {kib} URL for the deployment from which to fetch available locations.
 
 [NOTE]
 ====
-When a monitor is created or updated, it will not run immediately but at some point after. After the first run occurs, the monitor will begin running regularly based on the frequency configured for the monitor (for example, every 10 minutes). You can run a manual test if you want to see the results more quickly.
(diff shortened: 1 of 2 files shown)


docs/en/observability/synthetics-get-started-ui.asciidoc
```diff
@@ -132,6 +132,11 @@ Read more about available options in <<synthetics-command-reference>>.
 [[uptime-app-view-in-kibana]]
 = View in {kib}
 
+[NOTE]
+====
+When a monitor has been created (and also when edited), it will not run immediately but at some point after. After the first run occurs, the monitor will then run regularly based on the frequency configured for the monitor (e.g. every 10 minutes). You can run a manual test if you want to see the results more quickly.
```
vigneshshanmugam, line 137: nit: should we reword `it` as `the monitor`.

vigneshshanmugam (approved):
LGTM

docs/en/observability/synthetics-get-started-project.asciidoc
```diff
@@ -195,6 +195,12 @@ with the {kib} URL for the deployment from which to fetch available locations.
 [discrete]
 = View in {kib}
 
+
+[NOTE]
+====
+When a monitor has been created (and also when edited), it will not run immediately but at some point after. After the first run occurs, the monitor will then run regularly based on the frequency configured for the monitor (e.g. every 10 minutes). You can run a manual test if you want to see the results more quickly.
```
colleenmcginnis, line 201: What do you think of this? 

```suggestion
When a monitor is created or updated, it will not run immediately but at some point after. After the first run occurs, the monitor will begin running regularly based on the frequency configured for the monitor (for example, every 10 minutes). You can run a manual test if you want to see the results more quickly.
```

Also...

>at some point after

Can we be any more specific without giving a precise timeline? Maybe something like "typically less than a minute later" or "typically a few minutes later"?  (I don't actually know what is typical so feel free to adjust.)

colleenmcginnis (commented):
Oh! I missed this one earlier. Some minor suggestions below.

docs/en/observability/synthetics-get-started-project.asciidoc
paulb-elastic: Thanks @colleenmcginnis I've applied your tweaks.

Regarding being more precise, we have a challenge there, in that the UI doesn't know (and we can't add this in before GA).

It will be less than the frequency, but on a 1 hour monitor, that could still in theory be 59 minutes away.

I wonder if we should just include an example to clarify the point?  For example:

> When adding a 10 minute monitor, you may have to wait up to 9 minutes for the first run, after which it will run every 10 minutes. When editing a monitor, you will need to wait for its next scheduled time for those changes to be reflected

paulb-elastic: Thanks @colleenmcginnis I've gone with that copy in both locations

colleenmcginnis (approved):

bmorelli25:
It looks like this missed 8.8.

@Mergifyio backport 8.8

bmorelli25:
@Mergifyio backport 8.8

PR #2909, opened 2023-05-04
Update examples to valid schedules

paulb-elastic:
The current example shows schedules that won't be used, but will be rounded up, which is misleading


colleenmcginnis (approved):
I switched the `backport-8.4` label for `backport-8.6`. [This page didn't exist in 8.4](https://www.elastic.co/guide/en/observability/8.4/synthetics-lightweight.html).

Issue #2896, opened 2023-05-01
[Synthetics] Change how _enabling monitor management_ is documented


paulb-elastic:
### Description

With the [automatic enabling of Monitor Management](https://github.com/elastic/kibana/issues/146721), if a user has permissions, Monitor Management in Synthetics will automatically get enabled.  The current docs [here](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html#_prerequisites_2) and [here](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-ui.html#uptime-set-up-prereq) describe that this needs enabling.

From 8.8.0, when loading the Synthetics UI, if the user has permissions, Monitor Management will automatically get enabled.  If they don’t have permissions, they will [receive a message](https://github.com/elastic/kibana/issues/156166) advising them that it needs to be enabled (by someone with sufficient permissions), before they can use it.

### Resources

* Automatic enabling of Monitor Management has been done in https://github.com/elastic/kibana/issues/146721
* If insufficient permissions, the messaging is being updated in https://github.com/elastic/kibana/issues/156166
* Current docs that will need updating are [here](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html#_prerequisites_2) and [here](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-ui.html#uptime-set-up-prereq)

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** @dominiqueclarke 

**Stakeholders:**
* @paulb-elastic

PR #2925, opened 2023-05-08
[synthetics] Change how enabling monitor management is documented

colleenmcginnis:
Closes https://github.com/elastic/observability-docs/issues/2896

Removes instructions to enable Monitor Management, and replaces it with:

* **Get started guides**: A sentence about what permissions are needed to get started.
* **Setup role**: An explanation of how Monitor Management is enabled by default if a user has sufficient permissions (typically administrators).

I'm very open to feedback on the changes made here! 

cc @paulb-elastic


paulb-elastic (approved):
A couple of nit comments, but nothing blocking, LGTM

Issue #2903, opened 2023-05-02
[Synthetics] Adjust privileges for roles on grant access secured resources page


dominiqueclarke:
The `setup` role on the "grant access to secured resources page" should be updated for 8.8.0.

Replace the `cluster: read_ilm` privilege with `cluster: manage_ilm`.

Add `Index | synthetics-*: monitor` privilege.

Remove or reframe `b. Grant at least one of following privileges:`. These privileges are no longer required specifically to enable the Synthetics App. However, they would still be required for using project monitors, which is currently spelled out in the `writer` role.

PR #2930, opened 2023-05-09
[synthetics] Update setup role privileges

colleenmcginnis:
Closes https://github.com/elastic/observability-docs/issues/2903

Adjusts privileges needed for the setup role and adds a note about users with the setup role who should also be able to create, modify, and delete monitors.

Pairs with #2925 (which also makes changes to the setup role docs)


paulb-elastic:
I'll defer to @dominiqueclarke's review of this

dedemorton (commented):
A couple of minor suggestions. Otherwise the content looks good from an editorial perspective.

dominiqueclarke (approved):
LGTM! Tested the actual docs against the app.

PR #2933, opened 2023-05-10
[synthetics]: remove --no-headless flag from cli flags

vigneshshanmugam:
+ We recently removed this flag in the Synthetics agent and no longer supported.


justinkambic (commented):
@vigneshshanmugam it might be good to add somewhere, perhaps under the `--playwright-options` section, that we can include `{"headless": true|false}` to tell the lib whether it should be headful or not.

justinkambic (approved):
LGTM for synthrum

Issue #2921, opened 2023-05-08
Add Enabling Universal Profiling in Kibana section to docs


mdbirnstiehl:
We need to replace the "use the instructions in your email..." part of the profiling docs with the information [here](https://wiki.elastic.co/display/PM/How+to+enable+Universal+Profiling+in+Kibana)

PR #2937, opened 2023-05-15
Add info on enabling profiling in Kibana.

mdbirnstiehl:
### Summary
This PR closes [Issue 2921](https://github.com/elastic/observability-docs/issues/2921).

Updated the docs to no longer tell people to follow instructions in their email. 

Added the necessary steps to enable Universal Profiling in Kibana.

[Preview of changes.](https://observability-docs_2937.docs-preview.app.elstc.co/guide/en/observability/master/profiling-get-started.html#profiling-set-up-on-cloud)


docs/en/observability/profiling-get-started.asciidoc
```diff
@@ -66,16 +66,30 @@ Even if you're profiling a smaller fleet, we recommend configuring at least two
...
 
 [discrete]
 [[profiling-enable-kibana]]
 === Enable Universal Profiling in Kibana
 
-You'll receive instructions on enabling Universal Profiling at the email you used in your https://docs.google.com/forms/d/e/1FAIpQLSd-SWVgvhO7Z_jAfaV9_bFGa0dUZPuX0JORzPGS8SDP7G-dVQ/viewform[sign-up form].
+To enable Universal Profiling in {kib}, complete the following steps:
 
-Once enabled, Universal Profiling in Kibana is located under *Observability* in the navigation menu.
+. Log in to https://cloud.elastic.co[{ecloud}] and locate your deployment.
```
bmorelli25, line 78: There was a big push a while back to use attributes when linking to the cloud console so that we can track documentation referrals. Not sure if we're still concerned about it, but it can't hurt.

```suggestion
. Log in to {ess-console}[{ecloud}] and locate your deployment.

```

dedemorton, line 78: I would avoid this level of granularity when documenting steps in the cloud UI. The layout of the deployment page has changed a lot historically, and big changes can happen outside of stack releases. (This means that you could document something for 8.8 that's true now, but might not be true in a couple weeks or a month.)

I'd look at the way the cloud docs describe editing user settings and go with that approach. Maybe the topics under https://www.elastic.co/guide/en/cloud/current/ec-editing-user-settings.html.

I'm not sure if this setup is also available for Elastic Cloud Enterprise. They have a slightly different style (like in [this topic](https://www.elastic.co/guide/en/cloud-enterprise/current/ece-enable-auditing.html)).

I would ask someone on that team the best approach to follow here.

I also wonder if the cloud docs should/will cover this setup, too.

dedemorton (commented):
Just a little drive-by peer review.

docs/en/observability/profiling-get-started.asciidoc
iogbole: @mdbirnstiehl @dedemorton  - This is instruction is only specific to 8.8. We will enable profiling by default in version 8.9 - https://github.com/elastic/prodfiler/issues/3182

iogbole (approved):
LGTM

dedemorton (approved):
From an editorial perspective, LGTM™

PR #2952, opened 2023-05-18
[synthetics] Add a note about timeout types and defaults

### LINKED ISSUE #2868: [Synthetics]:  Different timeout configuration in synthetic tests
### Description

There are different types of timeouts when it comes to running Synthetic tests. Basically it can be split in to Navigation and Action timeouts. Navigation refers to the timeouts that basically happens when there is a page navigation by means of `page.goto, page.click, etc`. Action timeouts are tied to actions that Playwright takes for finding selectors or locators etc. 

### Resources

We have recently made changes in the Synthetics agent to keep the default timeout of `50 seconds` for both Navigation and Action timeouts. 

Also added the option to configure these timeouts via `playwrightOptions` which can be configured in both the UI and for project based browser monitors. Users can override these default by different which are already documented in the Playwright API's. 

Default action timeout - https://playwright.dev/docs/api/class-browsercontext#browser-context-set-default-timeout
Default navigation timeout - https://playwright.dev/docs/api/class-browsercontext#browser-context-set-default-navigation-timeout

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** @vigneshshanmugam 

**Stakeholders:**
@paulb-elastic 
@vigneshshanmugam

colleenmcginnis:
Closes #2868 

Adds a note to the [Configure Synthetics projects](https://observability-docs_2952.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-configuration.html) page about timeout types, default values, and how to change the default value in your config file. Looks like this ([preview link](https://observability-docs_2952.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-configuration.html#synthetics-configuration-playwright-options)):

Screenshot 2023-05-18 at 11 38 54 AM

## For the reviewer

@vigneshshanmugam I'm not sure if this is what you had in mind -- I'm very open to suggestions on both the language used and where this information should go in the existing docs!

### DIFF
--- a/docs/en/observability/synthetics-configuration.asciidoc
+++ b/docs/en/observability/synthetics-configuration.asciidoc
@@ -56,6 +56,16 @@ An object that defines any variables your tests require.
 
 For available options, see the https://playwright.dev/docs/test-configuration[Playwright documentation].
 
+[NOTE]
+====
+Playwright has two types of timeouts that are used in Elastic Synthetics:
+https://playwright.dev/docs/test-timeouts#action-and-navigation-timeouts[action and navigation timeouts].
+
+Elastic Synthetics uses a default action and navigation timeout of 50 seconds.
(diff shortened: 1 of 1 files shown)


paulb-elastic (approved):
LGTM

PR #2977, opened 2023-05-24
Universal Profiling: Describe range for project-id, remove Elastic Agent

christos68k:
### Summary

- Added upper bound limit to project-id
- Removed elastic agent setup instructions

@iogbole This PR removes the HA via Elastic Agent setup instructions and will also be backported to `8.8` (as part of the next release `8.8.1`). After chatting with you, I think the consensus is that we don't consider setup via Elastic Agent to be ready to be promoted yet (also, other options are missing from the Elastic Agent configuration as described in [#3383](https://github.com/elastic/prodfiler/issues/3383#issuecomment-1557239240)). Let me know if this is not OK.


mdbirnstiehl (approved):
LGTM

iogbole (approved):
LGTM

PR #2995, opened 2023-06-07
[Profiling] Add upgrade docs for 8.9

danielmitterdorfer:
With this commit we add specific instructions for the upgrade to Universal Profiling 8.9.x. For upgrades from 8.8.x to 8.9.x we offer the possibility to migrate existing data with a script if certain requirements are met. For upgrades from earlier versions we still require the prior procedure.

Note that this procedure is specific to 8.9.x; we plan to automatically upgrade data with 8.10.x. Therefore this script will only be required for upgrade to 8.9.x.


docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -9,7 +9,8 @@ beta::[]
 
 Universal Profiling is a beta program, and we are developing and iterating over it frequently upon receiving feedback.
 For this reason, you have to erase *all existing profiling data* from your deployment when upgrading between minor releases in the {stack}.
-When the product hits general availability, we will remove this requirement.
+If you are upgrading from 8.8 we also provide a script that allows you to keep your data. When the product hits general availability,
```
mdbirnstiehl, line 12:
```suggestion
When upgrading from version 8.8, you can use the migration script to keep your data. When the product hits general availability,

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -19,8 +20,8 @@ To upgrade Universal Profiling, you need to:
 
 . Perform a stack upgrade in {ecloud}
 . Stop profiling data ingestion
-. Delete existing profiling data
-. Setup Universal Profiling from scratch
+. When upgrading from version 8.8.0 or better: Migrate existing profiling data
```
mdbirnstiehl, line 23:
```suggestion
. *When upgrading from version 8.8.0 or more recent*: Migrate existing profiling data

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -19,8 +20,8 @@ To upgrade Universal Profiling, you need to:
 
 . Perform a stack upgrade in {ecloud}
 . Stop profiling data ingestion
-. Delete existing profiling data
-. Setup Universal Profiling from scratch
+. When upgrading from version 8.8.0 or better: Migrate existing profiling data
+. When upgrading from versions before 8.8.0: Delete existing profiling data and setup Universal Profiling from scratch
```
mdbirnstiehl, line 24:
```suggestion
. *When upgrading from versions prior to 8.8.0*: Delete existing profiling data and setup Universal Profiling from scratch

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +55,346 @@ Integrations Server. To do this:
 
 NOTE: When stopping incoming requests, host-agent replicas back off and retry connecting to the {ecloud} endpoint at increasing time intervals.
 
+[discrete]
+[[profiling-migrate-data]]
+=== Migrate existing profiling data
+
+For upgrades to 8.9 we offer a migration script if you meet the following requirements:
```
mdbirnstiehl, line 62:
```suggestion
When upgrading to version 8.9, you can use a migration script to migrate existing data if you meet the following requirements:

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +55,346 @@ Integrations Server. To do this:
...
+[discrete]
+[[profiling-migrate-data]]
+=== Migrate existing profiling data
+
+For upgrades to 8.9 we offer a migration script if you meet the following requirements:
+
+* The cluster has been upgraded from 8.8 to 8.9.
+* The migration is performed on a Linux workstation with `jq` and `curl` installed.
+
+If you meet these requirements you may run the provided script below. Otherwise we suggest to delete existing profiling data and set up Universal Profiling from scratch.
```
mdbirnstiehl, line 67:
```suggestion
If you meet these requirements, follow the steps below to run the migration script. If not, we suggest <<profiling-delete-data, deleting your existing profiling data>> and <<profiling-from-scratch, setting up Universal Profiling from scratch>>.

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +55,346 @@ Integrations Server. To do this:
...
+
+We also recommend to verify that you have a recent snapshot of your data so you can restore your data in case there is an error during the migration.
```
mdbirnstiehl, line 69:
```suggestion
You should also verify that you have a recent snapshot of your data so you can restore your it if there's an error during the migration.

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +55,346 @@ Integrations Server. To do this:
...
+
+main
+----
+====
+
 [discrete]
 [[profiling-delete-data]]
 === Delete existing profiling data
 
+_This step is required for upgrades from versions before 8.8.0. For upgrades from 8.8.0 or better you can <<profiling-migrate-data,migrate your data>>._
```
mdbirnstiehl, line 396:
```suggestion
NOTE: This step is required for upgrades from versions older than 8.8.0. For upgrades from 8.8.0 or newer, you can <<profiling-migrate-data,migrate your data>>.

```

mdbirnstiehl (commented):
I added a couple of languages suggestions

danielmitterdorfer:
Thank you very much for the review and the language suggestions @mdbirnstiehl! I've applied them all now; could you please have another look (no need to validate the migration script for correctness; @inge4pres will need to review that)?

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -9,7 +9,9 @@ beta::[]
 
 Universal Profiling is a beta program, and we are developing and iterating over it frequently upon receiving feedback.
 For this reason, you have to erase *all existing profiling data* from your deployment when upgrading between minor releases in the {stack}.
-When the product hits general availability, we will remove this requirement.
+When upgrading from version 8.8, you can use the migration script to keep your data. When the product hits general availability,
+
```
mdbirnstiehl, line 13:
```suggestion

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -19,8 +21,10 @@ To upgrade Universal Profiling, you need to:
 
 . Perform a stack upgrade in {ecloud}
 . Stop profiling data ingestion
-. Delete existing profiling data
-. Setup Universal Profiling from scratch
+. *When upgrading from version 8.8.0 or more recent*: Migrate existing profiling data
+
```
mdbirnstiehl, line 25:
```suggestion

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +58,350 @@ Integrations Server. To do this:
...
+* The cluster has been upgraded from 8.8 to 8.9.
+* The migration is performed on a Linux workstation with `jq` and `curl` installed.
+
+If you meet these requirements, follow the steps below to run the migration script. If not, we suggest <<profiling-delete-data, deleting your existing profiling data>> and <<profiling-from-scratch, setting up Universal Profiling from scratch>>.
+
+
+You should also verify that you have a recent snapshot of your data so you can restore your it if there's an error during the migration.
+
+
```
mdbirnstiehl, line 76:
```suggestion

```

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +58,350 @@ Integrations Server. To do this:
...
+When upgrading to version 8.9, you can use a migration script to migrate existing data if you meet the following requirements:
+
+
+* The cluster has been upgraded from 8.8 to 8.9.
+* The migration is performed on a Linux workstation with `jq` and `curl` installed.
+
+If you meet these requirements, follow the steps below to run the migration script. If not, we suggest <<profiling-delete-data, deleting your existing profiling data>> and <<profiling-from-scratch, setting up Universal Profiling from scratch>>.
+
+
+You should also verify that you have a recent snapshot of your data so you can restore your it if there's an error during the migration.
```
mdbirnstiehl, line 74:
```suggestion
You should also verify that you have a recent snapshot of your data so you can restore it if there's an error during the migration.

```

mdbirnstiehl (approved):
Just a couple of extra lines in there and an extra word, but otherwise LGTM!

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -19,8 +20,9 @@ To upgrade Universal Profiling, you need to:
 
 . Perform a stack upgrade in {ecloud}
 . Stop profiling data ingestion
-. Delete existing profiling data
-. Setup Universal Profiling from scratch
+. *When upgrading from version 8.8.0 or more recent*: Migrate existing profiling data
```
inge4pres, line 23: I have a doubt on this one: previously (up until 8.8) we never created the `profiling-metrics` data stream in customers deployment.
When running 8.9 and the new Elasticsearch-based index management, will the data stream be created?

danielmitterdorfer: Yes, the plugin will create any indices that are missing. Existing component and composable templates will be upgraded to their latest version and the ILM policy has an entirely name (we'll phase out the old one eventually).

danielmitterdorfer:
Test instructions for the migration script:

1. Create a new Elastic Cloud cluster (version `8.9.0-SNAPSHOT`) and have it restore a snapshot from the MVP cluster (Elasticcsearch cluster id `48e4c6317ff04ecdbfcc891d9f88bf35`, deployment id: `2b5c22949c2a4254a739dec16b15eeff`) (see [Elastic Cloud docs](https://www.elastic.co/guide/en/cloud/current/ec-restoring-snapshots.html#ec-restore-across-clusters) on how to restore a snapshot when creating a deployment via the UI). The cluster should have the same topology as the existing MVP cluster; see the screenshot below of an example deployment configuration.
2. Once the cluster is ready, follow the instructions in the [upgrade docs](https://github.com/elastic/observability-docs/blob/ad98c514721c9aa869afaadd6922c9dabe927287/docs/en/observability/profiling-upgrade.asciidoc#migrate-existing-profiling-data).
3. Enable Universal Profiling in Kibana `xpack.profiling.enabled: true` and verify that data show up in the UI

Here is an example deployment configuration:

deployment-settings

danielmitterdorfer:
Just as an update why this PR is not yet merged although it has been formally approved: I'll wait for an in-depth review of the migration script. As this script modifies data in the customer's cluster, a thorough check is currently pending. Once that's done I'll merge this PR.

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +56,359 @@ Integrations Server. To do this:
...
+    }
+  }'
+
+  echo "Disabling ILM temporarily..."
+  curl -s -f -o /dev/null -X POST -k -u "$ES_USER":"$ES_PASSWORD" "$ES_HOST/_ilm/stop"
+
+  echo "Deleting obsolete indices..."
+  curl -s -f -o /dev/null -X DELETE -k -u "$ES_USER":"$ES_PASSWORD" "$ES_HOST/.profiling-ilm-lock"
+  curl -s -f -o /dev/null -X DELETE -k -u "$ES_USER":"$ES_PASSWORD" "$ES_HOST/profiling-returnpads-private,profiling-sq-executables,profiling-sq-leafframes,profiling-symbols?ignore_unavailable=true"
+  # We rely only on the "next" indices
```
danielmitterdorfer, line 292: As an explicit note: In the following lines we delete the "current" indices. We can't keep both easily (a custom reindex procedure would be possible but IMHO way too complex) so we only use the "next" indices which will contain data since the last rollover but not before. So customers may see partial data for older profiles.

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -19,8 +20,9 @@ To upgrade Universal Profiling, you need to:
 
 . Perform a stack upgrade in {ecloud}
 . Stop profiling data ingestion
-. Delete existing profiling data
-. Setup Universal Profiling from scratch
+. *When upgrading from version 8.8.0 or more recent*: Migrate existing profiling data
+. *When upgrading from versions prior to 8.8.0*: Delete existing profiling data and setup Universal Profiling from scratch
+
```
inge4pres, line 25: I know it's not part of this PR, but I spotted a potential problem a little further down.
When we say `Perform a stack upgrade`, I think we should include a step that includes `Stop profiling data ingestion` between point 3 and 4.

SeanHeelan:
I am currently waiting for a migration to finish, so once it does I'll let you know if it all went fine. My only comment so far is: the docs are good, and comprehensive, but I think there's the possibility that the layout of the `profiling-upgrade.asciidoc` file could confuse some users. 

As a user I would expect to read a doc like this from top to bottom, but after the "Migration" section we go on to the deletion section (which I am presumably not meant to do), and from there to "Set up universal profiling from scratch" (which I am unsure if I am meant to do) and from there to "Start profiling data ingestion" (which I think I am supposed to do). 

My first question is: am I just being dumb? Maybe the implied actions are clear to everyone else? If so, then ignore the rest of this ;) 

If there is the potential for confusion, then the most drastic solution is to have entirely separate documents for the migration case versus the deletion case. This would result in some duplication on the stopping/starting of ingest (although we could move that to a separate, third, document and link to it), but when dealing something where users may be worried about data integrity, and craving clarity, I think this may be a worthwhile tradeoff. 

Alternatively, if we don't want separate docs, could we do something to make it clear exactly what path through the doc a user should take?

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +56,359 @@ Integrations Server. To do this:
...
+function main {
+  check_preconditions
+  ask_user_prerequisites
+  sanity_check_cluster
+  migrate
+
+  echo "Migration has finished. Please perform the following tasks:"
+  echo ""
+  echo "* Start profiling data ingestion"
+  echo "* Enable Elastic Universal Profiling in Kibana"
```
inge4pres, line 399: Do you think we can remove this line? Profiling is enabled by default in all Cloud deployments since 8.9.

inge4pres (approved):
The migration script works great 👏🏼 
I had the chance to test it using the 8.9 BC that is already published in CFT 👌🏼 
I loved the confirmation questions and the debug output, very handy - thanks 🙏🏼 

Minor correction on the script's output above in comments.

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +56,359 @@ Integrations Server. To do this:
...
+[discrete]
+[[profiling-migrate-data]]
+=== Migrate existing profiling data
+
+When upgrading to version 8.9, you can use a migration script to migrate existing data if you meet the following requirements:
+
+* The cluster has been upgraded from 8.8 to 8.9.
+* The migration is performed on a Linux workstation with `jq` and `curl` installed.
+
+If you meet these requirements, follow the steps below to run the migration script. If not, we suggest <<profiling-delete-data, deleting your existing profiling data>> and <<profiling-from-scratch, setting up Universal Profiling from scratch>>.
```
athre0z, line 68: At this point I went looking through the prior text to find a link to the script for about 30 seconds before realizing that it follows below -- might just be me though.

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +56,359 @@ Integrations Server. To do this:
...
+  *  `export ES_HOST=<elasticsearch endpoint URL>`
+. Run the migration: `./migrate-8.9.sh`
+. Unset environment variables again: `unset ES_USER ES_PASSWORD ES_HOST`
+
+[%collapsible]
+.Migration Script
+====
+[source,bash]
+----
+#!/usr/bin/env bash
```
athre0z, line 90: Marking all of this text and copying it resulted in 2 blank lines to be copied before the shebang, which then resulted in this script being executed by `sh` instead of `bash` on my box where it bailed out at `set -o pipefail`. Not sure if we can do much about this though.

docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,10 +56,359 @@ Integrations Server. To do this:
...
+  echo "Reenabling ILM..."
+  curl -s -f -o /dev/null -X POST -k -u "$ES_USER":"$ES_PASSWORD" "$ES_HOST/_ilm/start"
+
+  printf "Checking cluster health ... "
+  local status=$(curl -s -k -u "$ES_USER":"$ES_PASSWORD" "$ES_HOST/_cluster/health?wait_for_status=green&wait_for_no_initializing_shards=true&wait_for_no_relocating_shards=true" | jq -r ".status")
+  if [ "$status" == "red" ]; then
+    printf "[ERROR: status is red]\n"
+  else
+    printf "[OK: status is %s]\n" "$status"
+  fi
```
athre0z, line 381: I tend to deploy my testing clusters without ES replication to save us some money & this check prevents me from upgrading such deployments. Should we allow yellow? Not sure. Perhaps alternatively we could add an argument to allow skipping this?

athre0z (commented):
I created and populated a 8.8.2 deployment with mixed native + Java frames and upgraded it to the 8.9.0 RC (listed as 8.9.0 final in the LA Cloud region) following your instructions. It looks like most things transferred well, but I noticed two things:

- When kicking off the Profiling setup after running the migration script it would show a 404 Not Found error message toast in the corner. I unfortunately forgot to screenshot it.
  - When refreshing the page afterwards, it appeared to have completed anyway
- Traces and counts were correctly transferred, but symbol information for native frames appears to have gotten nuked
  - This includes both symbol info for public binaries and private symbols that I had uploaded
  - Once I start ingesting traces again, the symbols for public executables start showing up again, but private executables (with manually pushed symbols) remain unsymbolized. 

Is this expected?

## Before upgrade

Screenshot 2023-06-28 at 14 47 07

## After upgrade

Screenshot 2023-06-28 at 15 14 45

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: I can swap the steps, so it says:


1. Stop profiling data ingestion
2. Perform a stack upgrade in Elastic Cloud
3. When upgrading from version 8.8.0 or more recent: Migrate existing profiling data
4. When upgrading from versions prior to 8.8.0: Delete existing profiling data and setup Universal Profiling from scratch
5. Start profiling data ingestion

I'll also reorder the sections accordingly.

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: Good catch. I've reworded the entire statement now to:

> Migration has finished. You can now start profiling data ingestion.

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: Actually we do allow a `yellow` status. The script output should have been

> [OK: status is yellow]

And then the script should have continued normally. Did you see a different behavior?

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: Good point. How about we reword this sentence to:

> If you meet these requirements, follow the steps below to migrate your data.

This just mentions the goal ("migrate the data") and directs the user to the following steps. The solution (i.e. the script) is only mentioned in the steps and at that point it's hopefully easier to spot. One note: I would have wished we could just provide a download link instead of rendering the script inline but the inline approach seems to be the established practice.

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: That's not great and I could reproduce this. 

I played around a bit with CSS styles on the preview page but couldn't find the culprit. The only thing I noticed is that the behavior is identical to the ES docs. E.g. when you copy the snippet in the [index API docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#create-document-ids-automatically) you'll also notice the two leading and trailing lines.

@mdbirnstiehl is there a possibility to render a "Copy to clipboard" button at the bottom of the script block (similar to the ES docs) to simplify Copy & Paste for users?

danielmitterdorfer:
@athre0z first of all, thanks for the thorough check.

> When kicking off the Profiling setup after running the migration script it would show a 404 Not Found error message toast in the corner. I unfortunately forgot to screenshot it. 

Thanks for noticing. I'll try to reproduce this later but as it disappeared after a refresh it's probably only a minor thing.

> Traces and counts were correctly transferred, but symbol information for native frames appears to have gotten nuked 

Yes, currently `profiling-symbols` is deleted. My rationale back then was that this is transient information, especially the public symbols. So it is expected that public symbols disappear temporarily. However, the private symbols should have been preserved. Do you still have the cluster running so I can have a look what happened here?

danielmitterdorfer:
@SeanHeelan thanks for your comments.

> As a user I would expect to read a doc like this from top to bottom, but after the "Migration" section we go on to the deletion section (which I am presumably not meant to do), and from there to "Set up universal profiling from scratch" (which I am unsure if I am meant to do) and from there to "Start profiling data ingestion" (which I think I am supposed to do).

I understand your confusion. At the top of the page we explain the high-level process though. If you follow the page from top to bottom you should always see a note explaining under which circumstances you're meant to follow this section. The only section where this is actually missing is "Set up Universal Profiling from scratch".

However, I think your point of separating the two cases makes sense especially as we will need to differentiate for quite a while to ensure that any user that upgrades from 8.8.x or below needs to wipe their data (8.9.0 will be a special case). I've implemented your idea in https://github.com/elastic/observability-docs/pull/2995/commits/4ad72d0e9a1f618866b3c17ddea99db604c3fac4.

SeanHeelan:
Just completed running the script and all seems good 👍

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: Ah, I was mislead by the snippet. There are actually two checks:

1. Before the migration to ensure the cluster is in a healthy state. Indeed, there we check for cluster health `green`
2. After the migration to make sure nothing is off. We are more lenient here and allow `yellow` because it's possible that replicas are not fully up yet after the migration (they will be eventually)

I think we can be probably a bit more lenient and allow cluster health `yellow` in the first check. I've pushed the change in https://github.com/elastic/observability-docs/pull/2995/commits/14f3e39f71edb2c26b1e35c45485d35c9f58bfd5.

docs/en/observability/profiling-upgrade.asciidoc
athre0z: Sounds good!

docs/en/observability/profiling-upgrade.asciidoc
athre0z: As a pretty terrible workaround we could add an `if [ $0 == "sh" ]` and then re-run with bash if necessary. But yeah, fixing the CSS (or whatever is causing this) would probably improve things in general (for all other snippets as well). I also recently noticed empty lines to be copied when following instructions for creating an Elastic setup in K8S and it was rather annoying.

docs/en/observability/profiling-upgrade.asciidoc
athre0z: Ah, sorry for the confusion with quoting the wrong snippet.

athre0z:
> However, the private symbols should have been preserved. Do you still have the cluster running so I can have a look what happened here?

Yes, here:

https://my-deployment-26ef07.kb.us-west2.gcp.elastic-cloud.com:9243/app/profiling/flamegraphs/flamegraph?kuery=&rangeFrom=now-7d%2Fd&rangeTo=now

The two executables that I uploaded symbols for were `symbtool` and `pf-host-agent`.

> EDIT: The migration script creates a log file elastic_universal_profiling_migration_log.txt. Could you please share this with me?

I created the migrate script in `/tmp` and rebooted since -- afraid this is gone. :/

danielmitterdorfer:
> The two executables that I uploaded symbols for were symbtool and pf-host-agent. The range information appears to be present in .profiling-symbols-private-v001, but the stackframe index appears to have gotten nuked partially (all native frames that were resolved previously gone). 

The migration deletes `profiling-stackframes` and only keeps `profiling-stackframes-next`; reasoning [above](https://github.com/elastic/observability-docs/pull/2995#discussion_r1243670816), replicated here for convenience:

> As an explicit note: In the following lines we delete the "current" indices. We can't keep both easily (a custom reindex procedure would be possible but IMHO way too complex) so we only use the "next" indices which will contain data since the last rollover but not before. So customers may see partial data for older profiles.

> .profiling-returnpads-private-v001 also seems to be empty. Not sure whether the latter never had records in the first place or got nuked in the migration. I'd suspect it's the latter. 

You suspect correctly. I've pushed a fix in https://github.com/elastic/observability-docs/pull/2995/commits/c0b6c437179bdb5f305a82798cd927c11bd3a973. Would it be possible that you retry the migration with that script? Private symbols should be preserved now and public symbols should trickle in after the upgrade.

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: Thanks for the workaround. I'll keep this in mind and will add this if we don't find another way to address this (i.e. fixing the CSS).

athre0z:
> You suspect correctly. I've pushed a fix in https://github.com/elastic/observability-docs/commit/c0b6c437179bdb5f305a82798cd927c11bd3a973. Would it be possible that you retry the migration with that script? Private symbols should be preserved now and public symbols should trickle in after the upgrade.

Thanks for the fix! Yes, I'll retest the procedure later today.

docs/en/observability/profiling-upgrade.asciidoc
mdbirnstiehl: I believe the only code block source that adds the copy button is [source,console]. If that is changed, you'll see the copy button as shown below. But that will impact how the script renders in the UI. Not sure if that is a big deal or not in this instance.

image

danielmitterdorfer: Thanks Mike for the clarification. In that case we'll probably go with the workaround that Joel has proposed in https://github.com/elastic/observability-docs/pull/2995#discussion_r1246477490.

danielmitterdorfer:
Thank you for the review! I'll address the remaining small bits soon and will then merge (and backport) this.

docs/en/observability/profiling-upgrade.asciidoc
danielmitterdorfer: I played around with this a bit more and it turns out that 

* different OS configurations behave differently. For example, `/usr/bin/sh` is simply symlinked to `/usr/bin/bash`.
* it's not straightforward to determine in which shell you're actually running (see [whatshell.sh](https://www.in-ulm.de/~mascheck/various/whatshell/whatshell.sh.comments.html) for a rather complete implementation). 

I figured the safest option is to ask the user to run the script in bash by changing the instructions from:

```
./migrate-8.9.sh
```

to

```
bash migrate-8.9.sh
```

This ensures that the command is run in bash regardless whether the shebang is detected.

PR #3007, opened 2023-06-10
Update infra monitoring docs with UI changes

### LINKED ISSUE #2840: [REQUEST]: Validate and update infrastructure monitoring documentation
### Description

**Problem**

The topics under Infrastructure monitoring in the Observability Guide need to be tested and updated as required.
Although I made some tactical updates when I worked on stack monitoring updates, the docs under infrastructure monitoring have not been actively maintained by the writing team for quite awhile.

**How to fix**

Here's what we need to do:

- [x] Test the content to verify accuracy and fix any problems that are found.
- [x] Update screen captures as required.
- [ ] Reach out to the dev/prod/support teams for input on any usability issues identified by users and fix the issues as time allows.

~We can skip the new topic about the[ Hosts page](https://www.elastic.co/guide/en/observability/current/analyze-hosts.html) because it was updated recently.~ This issue is a month old, so I will test the Hosts page, too.

### Resources

n/a

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** n/a

**Stakeholders:** n/a

dedemorton:
Closes #2840.

This was meant to be a quick PR to update some screens and minor UI changes, but it seems that the Hosts view has changed quite a bit. I think what I've added just states the obvious and might need more work.

@roshan-elastic  Can you provide some guidance about what we should highlight in the docs about for the new tabs (Logs and Alerts)?

Preview link: https://observability-docs_3007.docs-preview.app.elstc.co/guide/en/observability/master/analyze-hosts.html


roshan-elastic:
Hey @dedemorton, well spotted and good call out - I can help you with this.

Here is a quick summary but I'm happy to set time aside to go through this in more detail/provide more documentation etc:

### Logs

- This will return logs for any of the hosts returned in the current query (i.e. if you see the host in the table above - including one of the hosts you can't see unless you go to page 2 or 3 or 4 etc - then logs for that host will be returned).
- It searches on `host.name` when it queries alerts to be returned

**Gotchas**
- By default, it will use the columns and indices you have configured in the Logs UI [settings](https://edge-lite-oblt.kb.us-west2.gcp.elastic-cloud.com/app/logs/settings?logView=(logViewId:default,type:log-view-reference)):
image
(this means that the `host.name` will not show by default - this is something we plan to address in a future version)

**Alerts**

- This will return alerts for any of the hosts returned in the current query (i.e. if you see the host in the table above - including one of the hosts you can't see unless you go to page 2 or 3 or 4 etc - then alerts for that host will be returned).
- It searches on `host.name` when it queries alerts to be returned

**Host Limit**

This one is quite important, we added this in to allow you to query large datasets without needing to wait for all of them to be returned (some of the existing UI features Inventory doesn't run well with large numbers of hosts). Here is how it works:

- It returns back the only the first X hosts specified in the host limit button (i.e. 10, 20, 50, 100 or 500)
- If there are more than hosts to be returned that the limit specifies (e.g. if you have 10,000 hosts and have set a limit of 10) then the UI will show that it is only returning the first 10:

image

- It will return the hosts by those which have most recently sent data (i.e. we are effectively looking through the hosts sorted by the latest @timestamp they sent).

**Gotchas**
- Because the hosts are being returned by the latest @timestamp received, the ordering is kind of sampled...it's not picking them in any particular order except for if there is a host which hasn't sent data as recently as others...it is more likely to be missed

- We intended to allow the user to sort by a field (e.g. CPU usage) and then return the top X hosts sorted by that metric (e.g. the top 100 CPU utilised hosts). Unfortunately, Elastic doesn't let us do this due to some technical limitations right now. 

- The current sorting behaviour (if you sort a table) will be that it will sort the _already returned_ dataset (i.e. if you are returned 500 of 10,000 hosts - sorting by CPU usage will just sort the 500 already returned...it won't pull out the top 500 most CPU utilised hosts from the 10,000).

- Our hope is that users will intuitively notice this as the sorting will be instant. 

- All of the above means that, if the host limit means you don't get the full data set back (i.e. you only see 10 of 10,000 hosts), you need the ensure the host count is less than the host limit. To do this, you have to either:

 (a) Increase the host limit until it is bigger than the host count (e.g. host limit of 500 if 400 are returned)...this isn't possible if you have more than 500 hosts
 (b) Filter your hosts down until there are fewer than the host limit

**Note** : We are likely going to all an 'all' option for the host limit to 'break the glass' and return all hosts - even if it's very slow (we didn't do this because the alerting & logs tabs wouldn't work with his logic. For now, this is the current behaviour though.

**Other UI changes**
- The host table now only lets you show 5, 10 or 20 hosts at a time
- The charts at the bottom will always show the same 5, 10 or 20 hosts as you see in the table

**Note** : We'll be adding the ability to filter by one or more hosts in the table directly in the future (i.e. checkboxes)

Does this help for now? Happy to provide more detail/jump on a call

docs/en/observability/monitor-infra/analyze-hosts.asciidoc
```diff
@@ -122,6 +132,41 @@ used to fetch the data and the response returned from {es}. You can click links
...
+
+On the **Logs** tab, view logs for the systems you are monitoring and search
+for specific log entries. This view shows logs for all of the hosts returned by
+the current query.
+
+[role="screenshot"]
+image::images/hosts-logs.png[Screenshot showing Logs view]
+
+TIP: This view uses the log sources and column settings specified in the
+Logs UI. To add log sources or columns (such as host name), refer to
```
dedemorton, line 147:
```suggestion
Logs app. To add log sources or columns (such as host name), refer to
```

docs/en/observability/monitor-infra/analyze-hosts.asciidoc
```diff
@@ -130,6 +175,9 @@ The **Hosts** page displays metrics from indices that match the index patterns
 specified on the **Settings page**. For example, the default patterns are
 `metrics-*` and `metricbeat-*`.
 
+Log sources are specified on the **Settings page** of the Logs app. For example,
+the default patters are `logs-*,filebeat-*,kibana_sample_data_logs*`.
```
dedemorton, line 179:
```suggestion
the default patterns are `logs-*,filebeat-*,kibana_sample_data_logs*`.
```

roshan-elastic:
> @roshan-elastic Thanks for providing more detail here. I've tried to distill your comments into some guidance in the docs. Please review my additions and offer suggestions. We probably should not invest a huge amount of effort into describing all the gotchas if they're going away, but I've tried to capture the important bits. Let me know what you think.

This is brilliant!

One thing I forgot to call out, which may/may not be useful to call out to the user, is the maroon/pink 'alert' count showing in the alert tab. This quickly shows how many active alerts there are over the date period for the hosts specified. The goal here is the quickly show a count of how many alerts are active on the hosts returned in your query (to help you get a sense of whether you should open the alerts tab or now).

The alerting docs will explain more about what an active alert is.

I'll approve this once you share your thoughts on this (I'll leave it up to you whether you think we should call out what this pink number means).

dedemorton:
>One thing I forgot to call out, which may/may not be useful to call out to the user, is the maroon/pink 'alert' count showing in the alert tab. This quickly shows how many active alerts there are over the date period for the hosts specified. The goal here is the quickly show a count of how many alerts are active on the hosts returned in your query (to help you get a sense of whether you should open the alerts tab or now).

I don't think we need to mention this in the docs because it seems fairly intuitive to me, especially after you click the Alerts tab. If you think the icon needs explanation when the tab is _not_ selected, then I suggest using something like a tooltip rather than putting the info in the docs. (IMO users are very unlike to refer to the docs to figure out what the icon means.)

roshan-elastic:
Cool sounds good. I'll approve in the morning!

roshan-elastic (approved):
Looks good to me!

PR #3039, opened 2023-06-28
Universal Profiling

florianl:
Various updates to Universal Profiling enable us to simplify the documentation.

- [ ] Universal Profiling no longer relys on APM and therefore its reference can be removed
- [ ]  Universal Profiling is no longer hidden in kibana behind `xpack.profiling.enabled`


iogbole (approved):
LGTM

mdbirnstiehl (commented):
The changes here LGTM, but I believe the profiling-enable-kibana section is linked to from the upgrade guide, which is causing this build to break. Do we need to update the upgrade guide as well?

florianl:
> [..]  profiling-enable-kibana section is linked to from the upgrade guide

good point! I have addressed this by updating the upgrade documentation as well.

mdbirnstiehl (approved):
I suggested one small update so we have a link to the Get started docs, but otherwise LGTM!

PR #3046, opened 2023-06-30
[Profiling] Update name of filter fields

danielmitterdorfer:
With this commit we update the names of two fields that users can use in Kibana filters:

* `service.name` has changed to `profiling.project.id`
* `host.ipstring` has changed to `host.ip`


mdbirnstiehl (approved):
LGTM!

PR #3061, opened 2023-07-05
[DOCS] Add category field to cases

lcawl:
Relates to https://github.com/elastic/kibana/issues/153835

This PR refreshes the automatically generated screenshot that is used in https://www.elastic.co/guide/en/observability/current/create-cases.html, such that it shows the new "Categories" column.

It also updates https://www.elastic.co/guide/en/observability/current/manage-cases.html so that the optional category field is mentioned alongside the similar tags field. Unlike tags, you can choose only a single category value for each case.


mdbirnstiehl (approved):
LGTM! Thank you!

PR #3064, opened 2023-07-06
Universal profiling - update probabilistic profiling

iogbole:


mdbirnstiehl (approved):
LGTM!

docs/en/observability/profiling-probabilistic-profiling.asciidoc
```diff
@@ -3,9 +3,9 @@
 
 beta::[]
 
-Probabilistic profiling allows you to reduce storage costs by collecting a representative sample of profiling data. 
-Over time and across the Profiling Host Agent fleet, this configuration results in approximately 50% of deployed agents collecting profiles at any given time.
-This reduces storage costs with the trade-off that profile collection is not enabled at all times on all Profiling Host Agent.
+Probabilistic profiling allows you to reduce storage costs by collecting a representative sample of profiling data. This method decreases storage costs with a visibility trade-off, as not all Profiling Host Agents will have profile collection enabled at all times. 
+
+Profiling Events and Metrics linearly correlate with the probabilistic profiling value. The lower the value, the fewer metrics and events are collected.
```
rockdaboot, line 8:
```suggestion
Profiling Events linearly correlate with the probabilistic profiling value. The lower the value, the fewer events are collected. 
```

The reason is (after looking into the 8.9 code), that the collector recently reduced the number of metrics stored in the customer's cluster. What is written here is
- host CPU usage
- host I/O throughput
- host I/O duration
- agent user CPU usage
- agent system CPU usage

a) This should reduce the customer's storage cost by a huge factor and I think the mentioning of metrics here is superfluous. But tbh, I don't have any numbers. @christos68k, do you know more about it ?

b) Probabilistic profiling should not influence the collection of the aforementioned metrics. Other metrics, like counters and errors, that occur during unwinding will indeed be reduced (roughly) linearly. But those are sent to the central Elastic metrics cluster and won't add up to the customer's storage costs. I expect the influence of metrics on the network traffic as negligible. WDYT, @christos68k ?

christos68k: Yes I wouldn't mention metrics here, we're only collecting CPU usage and IO throughput, which are not at all correlated with probabilistic profiling settings.

rockdaboot (approved):

PR #3065, opened 2023-07-06
[Profiling] Add new field to migration script

danielmitterdorfer:
With this commit we adapt the 8.9 Universal Profiling migration script to add a new field if the respective index exists.

Relates elastic/elasticsearch#97413


rockdaboot (approved):
Didn't test, but the code LGTM

Issue #3030, opened 2023-06-22
Document "project" settings in `synthetics.config.ts`


andrewvc:
Currently there's no documentation of how to update the `project` section of `synthetics.config.ts`. You can get a good config by using `npx @elastic/synthetics init` but for future changes this can be confusing to users.

Spotted by @bshetti

shahzad31:
we also need to clarify the env specified in the template file are local env , not the actual running env https://github.com/elastic/synthetics/blob/main/templates/synthetics.config.ts#L3

PR #3073, opened 2023-07-12
[synthetics] Align CLI and config docs

colleenmcginnis:
Fixes https://github.com/elastic/observability-docs/issues/3030

Updates the [Use the Synthetics CLI](https://www.elastic.co/guide/en/observability/current/synthetics-command-reference.html) and [Configure Synthetics projects](https://www.elastic.co/guide/en/observability/current/synthetics-configuration.html) docs to including:

* Inventory the options in both and make sure all options line up.
* Link between CLI and config docs.
* Update the example config to match the [latest config template](https://github.com/elastic/synthetics/blob/main/templates/synthetics.config.ts).
  * Add a note that `env` is the environment users are pushing _from_ not the environment where monitors will run.

cc @paulb-elastic


docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -101,21 +123,65 @@ However there are some limitations when using external packages:
...
+The value defined via the CLI will take precedence.
+
+https://github.com/elastic/synthetics/blob/{synthetics_version}/src/locations/public-locations.ts#L28-L37[`--locations Array<SyntheticsLocationsType>`]::
+Where to deploy the monitor. Monitors can be deployed in multiple locations so that you can detect differences in availability and response times across those locations.
++
+To list available locations, refer to <<elastic-synthetics-locations-command>>.
++
+Locations passed will be merged with locations defined in your
+<<synthetics-configuration-monitor,`synthetics.config.js` file>> and
+<<synthetics-monitor-use,when using `monitor.use`>> in individual journey files.
```
colleenmcginnis, line 174: Is this true?

paulb-elastic: I just tested this out and the CLI setting for `--locations`:
* overwrites the location set globally via `monitor.locations` in `synthetics.config.ts`
* does not override the location set directly against the individual monitor via `monitor.use.locations`
* does not get _merged_ with either (it will either override or do nothing)

vigneshshanmugam: Yes this is working as expected. Priority order
1. Individual monitor config `monitor.use(locations)`
2. If set via CLI --locations
3. Global config from `synthetics.config.ts`

Nothing gets merged, everything will be overridden.

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -21,38 +21,58 @@ You will not need to use most command line flags -- they have been implemented
 purely to interact with the {synthetics-app}.
 However, there are some you may find useful:
 
-*`--debug`*::
+`--debug`::
```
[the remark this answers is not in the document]
vigneshshanmugam: we could add something as Note
```
 Note: use DEBUG=synthetics environment variable to capture Synthetics agent logs.:
```

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -104,21 +124,65 @@ However there are some limitations when using external packages:
...
+The {kib} URL for the deployment to which you want to upload the monitors.
++
+This can also be set in the configuration file using
+<<synthetics-configuration-project,`project.url`>>.
+The value defined via the CLI will take precedence.
+
+`--space <string>`::
+The identifier of the target {kibana-ref}/xpack-spaces.html[{kib} space] for the pushed monitors.
+Spaces help you organize pushed monitors.
+Defaults to `default`.
```
vigneshshanmugam, line 152: Maybe its just me, I find it better to say like `Pushes to "default" space if unspecified`

vigneshshanmugam (commented):
LGTM, can we also add `tags/alert` as part of the `monitor` options? Release notes has some details on how its done - https://github.com/elastic/synthetics/releases

colleenmcginnis:
>LGTM, can we also add tags/alert as part of the monitor options? Release notes has some details on how its done - https://github.com/elastic/synthetics/releases

Done in https://github.com/elastic/observability-docs/pull/3087.

vigneshshanmugam (approved):
LGTM

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -13,31 +13,53 @@ in Playwright-specific methods, for example, `ignoreHTTPSErrors`, `extraHTTPHead
...
+    },
+    playwrightOptions: {
+      ignoreHTTPSErrors: false,
+    },
+    /**
+     * Configure global monitor settings
+     */
+    monitor: {
+      schedule: 10,
+      locations: [ 'us-east4-a' ],
```
paulb-elastic, line 35: nit - I know this was like this before, but I'd suggest changing it to a _real_ value such as `us_east`

paulb-elastic (approved):
These are great updates, thank you @colleenmcginnis, it really enhances the docs and shows the options available

LGTM

PR #3079, opened 2023-07-14
[synthetics] Add info on timezones

### LINKED ISSUE #3004: [Synthetics] Document how to set locale and timezone via PlaywrightOptions
### Description

At the moment, neither the timezone or locale is set in the Synthetics Service, which will work for many sites, but some sites will not work properly without one or both of these being set.

The Service is going to be [changed](https://github.com/elastic/synthetics-service/issues/1312) to set the timezone for each individual location, but the locale will not be set as this will be dependent on the use case the user is trying to achieve (sometimes the same locate will be desired from different locations, sometimes it will intentionally want to be different).

These can already be set on a per monitor or global basis using for example:
```
playwrightOptions: {
  locale: 'en-AU',
  timezoneId: 'Australia/Brisbane',
}
```
We need to make this clearer in the documentation but also clarify that the `timezoneId` will reflect the timezone the location is running in.  Likewise we should clarify that `playwrightOptions` set in the script/configuration will take priority over the default `timezoneId` set in the service.

This should wait until https://github.com/elastic/synthetics-service/issues/1312 has been implemented and rolled out.

### Resources

Need to work out the best location, maybe somewhere in https://www.elastic.co/guide/en/observability/current/synthetics-create-test.html (looking for guidance here from the Docs team)

### Collaboration

The product team will provide the initial content.

### Point of contact.

**Main contact:** @paulb-elastic

**Stakeholders:**
- @vigneshshanmugam 
- @paulb-elastic

colleenmcginnis:
Fixes #3004 

Adds info on how to set timezones using `playrightOptions`.

cc @vigneshshanmugam

### DIFF
--- a/docs/en/observability/synthetics-configuration.asciidoc
+++ b/docs/en/observability/synthetics-configuration.asciidoc
@@ -52,17 +52,52 @@ An object that defines any variables your tests require.
 [[synthetics-configuration-playwright-options]]
 = `playwrightOptions`
 
-For available options, see the https://playwright.dev/docs/test-configuration[Playwright documentation].
+For all available options, refer to the https://playwright.dev/docs/test-configuration[Playwright documentation]. 
 
-[NOTE]
-====
+Below are details on a few Playwright options that are particularly relevant to Elastic Synthetics including timeouts, timezones, and device emulation.
+
+[discrete]
+[[synthetics-configuration-playwright-options-timeouts]]
+== Timeouts
+
 Playwright has two types of timeouts that are used in Elastic Synthetics:
 https://playwright.dev/docs/test-timeouts#action-and-navigation-timeouts[action and navigation timeouts].
 
 Elastic Synthetics uses a default action and navigation timeout of 50 seconds.
 You can override this default using https://playwright.dev/docs/api/class-testoptions#test-options-action-timeout[`actionTimeout`] and https://playwright.dev/docs/api/class-testoptions#test-options-navigation-timeout[`navigationTimeout`]
 in `playwrightOptions`.
-====
+
+[discrete]
+[[synthetics-configuration-playwright-options-timezones]]
+== Timezones
+
+By default monitors will run in the ??? timezone, but you can specify a timezone in
+`playwrightOptions` on a per monitor or global basis.
+
+To set the timezone for all monitors in the project, set https://playwright.dev/docs/emulation#locale--timezone[`locale` and `timezoneId`]
+in the configuration file:
+
+[source,js]
+----
+playwrightOptions: {
+  locale: 'en-AU',
+  timezoneId: 'Australia/Brisbane',
+}
+----
+
+To set the timezone for a specific monitor, add these options to a journey using
+<<synthetics-monitor-use,`monitor.use`>>.
+
+// Uncomment note when this is added to the Synthetics Service
+
+//////
+[NOTE]
+=====
+The timezone can also be set using the {synthetics-app}.
+If you specify a timezone in the configuration file or using `monitor.use` in the journey code,
+that timezone with take precedence over the timezone set in the {synthetics-app}.
+=====
+//////
 
 [discrete]
 [[synthetics-config-device-emulation]]
--- a/docs/en/observability/synthetics-command-reference.asciidoc
+++ b/docs/en/observability/synthetics-command-reference.asciidoc
@@ -42,6 +42,9 @@ For example, `cat path/to/file.js | npx @elastic/synthetics --inline`.
 JSON object to pass in custom Playwright options for the agent.
 Options passed will be merged with Playwright options defined in your synthetics.config.js file. Options defined via `--playwright-options` 
 take precedence.
++
+For more details on relevant Playwright options, refer to the
+<<synthetics-configuration-playwright-options,the configuration docs>>.
 
 *`--screenshots <on|off|only-on-failure>`*::
 Captures screenshots for every step in the journey.


docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,52 @@ An object that defines any variables your tests require.
...
 Elastic Synthetics uses a default action and navigation timeout of 50 seconds.
 You can override this default using https://playwright.dev/docs/api/class-testoptions#test-options-action-timeout[`actionTimeout`] and https://playwright.dev/docs/api/class-testoptions#test-options-navigation-timeout[`navigationTimeout`]
 in `playwrightOptions`.
-====
+
+[discrete]
+[[synthetics-configuration-playwright-options-timezones]]
+== Timezones
+
+By default monitors will run in the ??? timezone, but you can specify a timezone in
```
colleenmcginnis, line 74: It wasn't clear to me when reading the issue what the default behavior is if this is not explicitly set.

paulb-elastic: @colleenmcginnis unfortunately that's part of the problem.  On the service we're not currently setting this, so a default of `Etc/Unknown` is being _used_ which is not valid (see https://github.com/elastic/synthetics-service/issues/1312).

Many sites won't care, but if the site needs to check the timezone, they may fail because of this.

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,52 @@ An object that defines any variables your tests require.
...
+`playwrightOptions` on a per monitor or global basis.
+
+To set the timezone for all monitors in the project, set https://playwright.dev/docs/emulation#locale--timezone[`locale` and `timezoneId`]
```
paulb-elastic, line 77: I don't think this link is correct (it's linking to `https://playwright.dev/docs/emulation#locale%E2%80%94%E2%80%8Btimezone`)

docs/en/observability/synthetics-configuration.asciidoc
colleenmcginnis: Ohh I see.

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,52 @@ An object that defines any variables your tests require.
...
 You can override this default using https://playwright.dev/docs/api/class-testoptions#test-options-action-timeout[`actionTimeout`] and https://playwright.dev/docs/api/class-testoptions#test-options-navigation-timeout[`navigationTimeout`]
 in `playwrightOptions`.
-====
+
+[discrete]
+[[synthetics-configuration-playwright-options-timezones]]
+== Timezones
+
+By default monitors will run in the ??? timezone, but you can specify a timezone in
+`playwrightOptions` on a per monitor or global basis.
```
colleenmcginnis, line 75: What about something like this?

```suggestion
By default, Elastic Synthetics does not set a timezone for synthetic monitors.
Many websites don't require a site visitor's timezone to function, but if any journey in
your project is set up to visit a website that _does_ require a timezone, you can specify
a timezone to use when the monitor runs using `playwrightOptions` on a per monitor
or global basis.
```

paulb-elastic: To be picky, the service doesn't set the timezone but a private location may well have this set.  Also, we are intending to add the timezone to the service with https://github.com/elastic/synthetics-service/issues/1312 in the short term (although I can't say when exactly)

docs/en/observability/synthetics-configuration.asciidoc
vigneshshanmugam, line 74:
```suggestion
By default monitors uses the time zone of the host machine, but you can specify a timezone
```

docs/en/observability/synthetics-configuration.asciidoc
vigneshshanmugam: Just adding a suggestion. We could start like
```
By default monitors uses the time zone of the host machine. This is not always desirable if we want to test how a given web application behaves across different timezones. 
```

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,55 @@ An object that defines any variables your tests require.
...
 Elastic Synthetics uses a default action and navigation timeout of 50 seconds.
 You can override this default using https://playwright.dev/docs/api/class-testoptions#test-options-action-timeout[`actionTimeout`] and https://playwright.dev/docs/api/class-testoptions#test-options-navigation-timeout[`navigationTimeout`]
 in `playwrightOptions`.
-====
+
+[discrete]
+[[synthetics-configuration-playwright-options-timezones]]
+== Timezones
+
+By default, monitors use the timezone of the host machine. This is not always desirable
```
vigneshshanmugam, line 74: @paulb-elastic WDYT? host machine is it better or we can say like `machine where the monitors are executed?`

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,55 @@ An object that defines any variables your tests require.
...
 https://playwright.dev/docs/test-timeouts#action-and-navigation-timeouts[action and navigation timeouts].
 
 Elastic Synthetics uses a default action and navigation timeout of 50 seconds.
 You can override this default using https://playwright.dev/docs/api/class-testoptions#test-options-action-timeout[`actionTimeout`] and https://playwright.dev/docs/api/class-testoptions#test-options-navigation-timeout[`navigationTimeout`]
 in `playwrightOptions`.
-====
+
+[discrete]
+[[synthetics-configuration-playwright-options-timezones]]
+== Timezones
```
paulb-elastic, line 72: Maybe this should read `Timezones and locale` as both are referenced?

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,55 @@ An object that defines any variables your tests require.
...
+
+By default, monitors use the timezone of the host machine. This is not always desirable
+if you want to test how a web application behaves across different timezones.
+To specify what timezone to use when the monitor runs, you can use `playwrightOptions`
+on a per monitor or global basis.
+
+To use a timezone for all monitors in the project, set
```
paulb-elastic, line 79: Maybe: _To set a timezone and/or locale for all monitors..._

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -52,17 +52,55 @@ An object that defines any variables your tests require.
...
 in `playwrightOptions`.
-====
+
+[discrete]
+[[synthetics-configuration-playwright-options-timezones]]
+== Timezones
+
+By default, monitors use the timezone of the host machine. This is not always desirable
+if you want to test how a web application behaves across different timezones.
+To specify what timezone to use when the monitor runs, you can use `playwrightOptions`
```
paulb-elastic, line 76: Maybe: _To specify what timezone and/or locale to use when..._

docs/en/observability/synthetics-configuration.asciidoc
paulb-elastic: It's a bit awkward because we're not currently setting this for the service locations.  For the sake of this 8.9 doc update, maybe we should just call that out, something like:

> The Elastic global managed testing infrastructure does not currently set the timezone. For Private Locations, the monitors will use the timezone of the host machine running the Elastic Agent. This is not always desirable if you want to test how a web application behaves across different timezones. To specify what timezone to use when the monitor runs...

colleenmcginnis: Ah this is much clearer than what I had. Thank you!

paulb-elastic (approved):
LGTM

Issue #3006, opened 2023-06-09
[REQUEST]: Conditional actions in Observability rules


lcawl:
### Description

In 8.8, security rules gained the ability to have conditional actions (e.g. only send notifications if alert occurs between these hours, or only if it matches this filter). We updated the Kibana docs and Security docs for that feature.  Now it seems like some Observability rules will be able to have this functionality in 8.9.

Expected impact:

- [ ] Review new text that appears in the affected rule creation wizards
- [ ] Update screenshots and details in the action-related sections of the affected rule pages (under https://www.elastic.co/guide/en/observability/current/create-alerts.html)
- [ ] Add to release highlights

### Resources

- https://github.com/elastic/kibana/issues/158014
- https://github.com/elastic/kibana/issues/158017
- https://github.com/elastic/security-docs/issues/3142
- #ro-feature-conditional-action

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** 

I'm not sure who's working on the implementation from the observability dev team, only who's enabling the functionality on the Kibana side per https://github.com/elastic/kibana/issues/158017

**Stakeholders:**

@lcawl - Let me know if I can help since I wrestled with this understanding this concept in 8.8.

lcawl:
It looks like this is on hold now per https://github.com/elastic/kibana/issues/158017

mdbirnstiehl:
I've moved it to Sprint 9 (8.10) sprint accordingly.

mdbirnstiehl:
After talking with the devs, this will be in 8.9. So I've brought it back to the current sprint.

PR #3101, opened 2023-07-25
Add conditional actions for observability alerts

mdbirnstiehl:
This PR closes [Issue 3006](https://github.com/elastic/observability-docs/issues/3006).

Information on conditional actions were added to log threshold, metrics threshold, and infrastructure threshold alerts and screenshots were added or updated where necessary.


bmorelli25 (approved):
LGTM. Only question is whether we want to use a tagged region since all of the text is the same. It's not a lot of text, so I'm okay with the duplication if that's what you prefer.

PR #3107, opened 2023-07-26
Reverts #2995

danielmitterdorfer:
With this commit we revert the migration instructions to version 8.9.0 for Universal Profiling. There were some edge cases that we did not anticipate and therefore we fallback to the prior upgrade procedure of requiring users to wipe data for the upgrade.


danielmitterdorfer:
superseded by https://github.com/elastic/observability-docs/pull/3108

PR #3108, opened 2023-07-26
Reverts #2995

danielmitterdorfer:
With this commit we revert the migration instructions to version 8.9.0 for Universal Profiling. There were some edge cases that we did not anticipate and therefore we fallback to the prior upgrade procedure of requiring users to wipe data for the upgrade.


docs/en/observability/profiling-tag-data-query.asciidoc
```diff
@@ -8,7 +8,7 @@ The only config setting you may want to change is `project-id` (default value is
 
 The `-project-id` flag, or the `project-id` key in the host-agent configuration file, splits profiling data into logical groups that you control.
 
-You can assign any non-zero, unsigned integer <= 4095 to a host-agent deployment you control. In Kibana, the KQL field `profiling.project.id` is mapped to `project-id` and you can use it to split or filter data.
+You can assign any non-zero, unsigned integer <= 4095 to a host-agent deployment you control. In Kibana, the KQL field `service.name` is mapped to `project-id` and you can use it to split or filter data.
```
inge4pres, line 11: We should probably leave this as-is? the new field is going to be visible as `profiling.project.id` in the UI?

danielmitterdorfer: Ah, I missed that one. Good catch!

danielmitterdorfer: Addressed in https://github.com/elastic/observability-docs/pull/3108/commits/99f8948f4465be42446869af6b43b7947e6784cc.

mdbirnstiehl (approved):

PR #3111, opened 2023-08-01
Rearrange the profiler docs for UP agent

philippkahr:
I think that we should mention Elastic Agent at the top instead fo standalone, to show that Elastic Agent is the preferred way of running it.


bmorelli25 (approved):
👍 Makes sense! Thanks

PR #3112, opened 2023-08-02
Added the --no-headless option back in

paulb-elastic:
This capability has been added back in via https://github.com/elastic/synthetics/pull/813 so adding back to the docs

### DIFF
--- a/docs/en/observability/synthetics-command-reference.asciidoc
+++ b/docs/en/observability/synthetics-command-reference.asciidoc
@@ -72,6 +72,11 @@ Runs with the browser in headful mode.
 +
 This is the same as running `--playwright-options '{"headless": false}'`.
 
+[NOTE]
+=====
+Headful mode is only to be used locally when testing, to help see the browser being shown and to interact with DOM elements. Do not attempt to run in headful mode when running through Elastic's global managed testing infrastructure, or {private-location}s.
+=====
+
 `-h, --help`::
 Shows help for the `npx @elastic/synthetics` command.
 
--- a/docs/en/observability/synthetics-command-reference.asciidoc
+++ b/docs/en/observability/synthetics-command-reference.asciidoc
@@ -67,6 +67,11 @@ Throttling can also be disabled in the configuration file using
 <<synthetics-configuration-monitor,`monitor.throttling`>>.
 The value defined via the CLI will take precedence.
 
+`--no-headless`::
+Runs with the browser in headful mode.
++
+This is the same as running `--playwright-options '{"headless": false}'`.
+
 `-h, --help`::
 Shows help for the `npx @elastic/synthetics` command.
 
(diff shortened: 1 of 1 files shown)


shahzad31 (approved):

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -67,6 +67,11 @@ Throttling can also be disabled in the configuration file using
 <<synthetics-configuration-monitor,`monitor.throttling`>>.
 The value defined via the CLI will take precedence.
 
+`--no-headless`::
+Runs with the browser in headful mode.
++
+This is the same as running `--playwright-options '{"headless": false}'`.
```
vigneshshanmugam, line 73: Its good to inform users that we don't support running headful on the Synthetics global infrastructure. 

This is one of the main reason why I removed this flag as it was not uniform.

vigneshshanmugam: My worry is that, people start using it and not seeing results in their monitor. Not worried about CLI/PW Options, but just as a Note saying we dont support headful mode in our infrastructure.

paulb-elastic: @vigneshshanmugam good point, I've added a note to that effect if you could do another review

vigneshshanmugam (approved):
LGTM

docs/en/observability/synthetics-command-reference.asciidoc
colleenmcginnis: Do you think the audience for this doc will already be familiar with the terms "headless" and "headful"? If not, a link to the Playwright docs might be helpful. I'll leave it up to you!

```suggestion
This is the same as setting https://playwright.dev/docs/api/class-testoptions#test-options-headless[Playwright's `headless` option] to `false` by running `--playwright-options '{"headless": false}'`.
```

docs/en/observability/synthetics-command-reference.asciidoc
```diff
@@ -67,6 +67,16 @@ Throttling can also be disabled in the configuration file using
...
+
+[NOTE]
+=====
+Headful mode is only to be used locally when testing, to help see the browser being shown and to interact with DOM elements. Do not attempt to run in headful mode when running through Elastic's global managed testing infrastructure, or {private-location}s.
```
colleenmcginnis, line 77:
```suggestion
Headful mode should only be used locally to see the browser and interact with DOM elements directly for testing purposes. Do not attempt to run in headful mode when running through Elastic's global managed testing infrastructure or {private-location}s.
```

colleenmcginnis (approved):
Minor suggestions below. None are blocking!

docs/en/observability/synthetics-command-reference.asciidoc
vigneshshanmugam: Can we also mention that we dont support on our infrastructure instead of saying not to use it.

paulb-elastic:
Thanks @colleenmcginnis and @vigneshshanmugam, I have applied your suggestions and will merge once the build finishes and I've checked the output.

PR #3113, opened 2023-08-03
Update logs stream

mdbirnstiehl:
This PR closes [Issue 3114](https://github.com/elastic/observability-docs/issues/3114) and updates the section on configuring your agent to no longer integrations. 

@ruflin, with the process updated to exclude Kibana and remove the integration, what additional steps will we want to add to help users organize data that the integration might have helped with? Does it make sense to add an ingest pipeline, will mapping be covered by dynamic mapping, should we show how to create a data view, etc?


docs/en/observability/logs-stream.asciidoc
```diff
@@ -31,60 +31,42 @@ include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widge
...
+
+To configure your {agent}, set the values for the fields in the example configuration:
+
+. `hosts` – Your {es} URL and the port.
+. `api-key` – An API key to grant the agent access to {es}. To create an API key for your agent, see {fleet-guide}/grant-access-to-elasticsearch.html#create-api-key-standalone-agent[Create API keys for standalone agents].
 +
-NOTE: The format of the key is <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+NOTE: The API key format should be <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+. `inputs.id` – A unique identifier for your input.
+. `type` – The type of input. For collectings logs, set this to `logfile`.
```
ruflin, line 65: @belimawr I assume we should jump here to filestream?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -31,60 +31,42 @@ include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widge
...
-    hosts:
-      - 'https://da4e3a6298c14a6683e6064ebfve9ace.us-central1.gcp.cloud.es.io:443'
-    api_key: _Nj4oH0aWZVGqM7MGop8:349p_U1ERHyIc4Nm8_AYkw
-[...]
+    hosts: '<your-elasticsearch-url>:<port>'
+    api_key: 'your-api-key'
+inputs:
+  - id: your-log-id
+    type: logfile
+    data_stream:
```
ruflin, line 48: It is possible to just leave out the namespace and use default. I wonder if we should skip it at this stage?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -31,60 +31,42 @@ include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widge
...
 +
-NOTE: The format of the key is <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+NOTE: The API key format should be <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+. `inputs.id` – A unique identifier for your input.
+. `type` – The type of input. For collectings logs, set this to `logfile`.
+. `data_stream.namespace` – The name for your namespace data stream. The namespace is useful for grouping data such as an environment (dev, prod, or qa), a team, or a strategic business unit. Using a namespace makes it easier to search for data from a given source by using a matching pattern. The default value is `default`.
+. `streams.id` – A unique identifier for your stream of log data. 
+. `data_stream.dataset` – the name for your dataset data stream. You can name this data stream anything that signifies the source of the data. The default value is `generic`.
+. `paths` – the path to your log file.
```
ruflin, line 69: Lets mention here that it can be a pattern like /var/log/foo.log*

docs/en/observability/logs-stream.asciidoc
```diff
@@ -31,60 +31,42 @@ include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widge
...
 
 [discrete]
 [[logs-stream-install-agent]]
```
ruflin, line 72: The user should first install the elastic agent and then configure it?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -117,6 +99,17 @@ The following example shows the search results for logs with an `apm.error` data
 image::images/stream-logs-example.png[example search query on the logs stream page in {kib}]
 --
 
+[discrete]
+[[logs-stream-troubleshooting]]
+= Troubleshooting 
+
+If you're not seeing your log files in {kib}, check the following in the `elastic-agent.yml` file:
+
+NOTE: After installing the {agent}, you need to update the `elastic-agent.yml` in the directory described in the {fleet-guide}/installation-layout.html[installation layout].
```
ruflin, line 108: See above, the user should directly modify the elastic-agent.yml after installation.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -117,6 +99,17 @@ The following example shows the search results for logs with an `apm.error` data
...
+
+- Verify that the path to your logs file under `paths` is correct.
+- Verify that your API key is in <id>:<key> format. If not, your API key may be in an unsupported format, and you'll need to create an API key in *Beats* format.
```
ruflin, line 111: There are commands like `elastic-agent status` that we could recommend here for debugging. @belimawr Any recommendations or links we should point to?

ruflin:
I like the direction this is taking. The user now gets into Kibana at a later stage (which I think is good). 

For Elasticsearch, should we help the user to figure out the URL when in Cloud?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -31,60 +31,42 @@ include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widge
...
+        data_stream:
+          dataset: generic
+        paths:
+          - /var/log/your-logs.log
 ----
+
+To configure your {agent}, set the values for the fields in the example configuration:
+
+. `hosts` – Your {es} URL and the port.
+. `api-key` – An API key to grant the agent access to {es}. To create an API key for your agent, see {fleet-guide}/grant-access-to-elasticsearch.html#create-api-key-standalone-agent[Create API keys for standalone agents].
```
ruflin, line 61: The link to the API key creation is ok for now. But I think this could become a stumbling point. We should improve this on the UX side ( @grabowskit ). Ideally we could give the user something they can just copy paste into Dev Tools or Click a single button to the get API key (we have this planned) instead of having to go to the right place, put the right json in and then select the encoding.

docs/en/observability/logs-stream.asciidoc
belimawr: Yes!

docs/en/observability/logs-stream.asciidoc
belimawr: The first things that come to mind are:
- [Troubleshooting guide](https://www.elastic.co/guide/en/fleet/current/fleet-troubleshooting.html)

docs/en/observability/logs-stream.asciidoc
```diff
@@ -31,60 +31,42 @@ include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widge
...
 +
-NOTE: The format of the key is <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+NOTE: The API key format should be <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
```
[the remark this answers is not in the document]
cmacknz: No, no idea why it is like this. That was before my time.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[logs-stream]]
 = Stream a log file
 
-In this guide, you'll learn how to take a log file from your host and send it to Elasticsearch using a standalone {agent}. You'll configure the {agent} and your data streams using the Custom Logs integration. From there, you'll learn how to query your logs and use the data streams you've set up to have more control and flexibility when filtering your log data. 
+In this guide, you'll learn how to take a log file and send it to Elasticsearch using a standalone {agent}. You'll configure the {agent} and your data streams using the Custom Logs integration. From there, you'll learn how to query your logs and use the data streams you've set up to have more control and flexibility when filtering your log data.
```
ruflin, line 4: You mention the "Custom logs integration" here, but that is not the case anymore as it was removed from the flow.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -28,75 +28,62 @@ On your host, download and extract the installation package that corresponds wit
 include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widget.asciidoc[]
 
 [discrete]
-[[logs-stream-agent-config]]
-== Step 2: Configure the {agent}
+[[logs-stream-install-agent]]
+== Step 2: Install and start the {agent}
+After downloading and extracting the installation package, you're ready to install the {agent}. From the agent directory, run the the command that corresponds with your system to install the {agent} as a service. Do *not* enroll the agent in Fleet.
```
ruflin, line 33: You mention here "Do not enroll ...". As Fleet and enroll is not part of the guide, is there a chance the user would do this? If not, this introduces new terms the user does not even know about so would rather leave it out.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -28,75 +28,62 @@ On your host, download and extract the installation package that corresponds wit
...
 
-NOTE: On macOS, Linux (tar package), and Windows, run the `install` command to
-install {agent} as a managed service and start the service. The DEB and RPM
-packages include a service unit for Linux systems with
-systemd, so just enable then start the service.
+To configure your {agent}, set the values for the fields in the example configuration:
 
-include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/run-standalone-widget.asciidoc[]
+. `hosts` – Your {es} URL and the port. On {ecloud}, this . 
+.. Copy the {es} endpoint and cluster ID and from your deployment's page.
```
ruflin, line 71: As discussed, only the endpoint should be needed. This will also simplify the text below.

docs/en/observability/logs-stream.asciidoc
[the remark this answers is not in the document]
ruflin: @ogupte Have you stumbled over this during creation of the onboarding script? Anything special you do to the API key?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -28,75 +28,62 @@ On your host, download and extract the installation package that corresponds wit
...
++
+.. Your host URL is formatted as `https://CLUSTER_ID.REGION.CLOUD_PLATFORM.DOMAIN:PORT`. The default port is 443. For example, `https://ec47fc4d2c53414e1307e85726d4b9bb.us-east-1.aws.found.io:443`.
+. `api-key` – An API key to grant the agent access to {es}. To create an API key for your agent, see {fleet-guide}/grant-access-to-elasticsearch.html#create-api-key-standalone-agent[Create API keys for standalone agents].
++
+NOTE: The API key format should be <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+. `inputs.id` – A unique identifier for your input.
+. `type` – The type of input. For collectings logs, set this to `filestream`.
+. `streams.id` – A unique identifier for your stream of log data. 
+. `data_stream.dataset` – The name for your dataset data stream. You can name this data stream anything that signifies the source of the data. The default value is `generic`.
+. `paths` – The path to your log files. You can use a pattern like `/var/log/foo.log*`.
```
ruflin, line 86: @mdbirnstiehl After the changes were made the config, I think the user has to restart elastic-agent or will it pick up all the changes directly?

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: You're prompted to enroll in Fleet when installing, so they will be introduced to it there.  I can clarify to say "Do not  enroll the agent in Fleet when prompted during installation" or ignore it if you think a user might be familiar enough with the concepts of Fleet and standalone agents at this point.

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: When I tested with just the endpoint, I wasn't able to ingest any logs data. I also tried with/without the port, and I was only able to ingest data with the URL in this format. Maybe I'm doing something wrong there?

mdbirnstiehl: I tested this again, and was able to get it working with the endpoint URL if I had the port added, so we can simplify at least that far.

docs/en/observability/logs-stream.asciidoc
ruflin: I doubt a user is familiar is enough and shouldn't be. ++ on having it and potentially get this step out of the way.

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: In the testing I've done, it picks it up without restarting.

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: @ruflin I used 443. I just tested again on a fresh deployment and got the same result. Is it possible I don't have something set up correctly that would require manually entering the port?

mdbirnstiehl:
@ruflin I have updated the page again. I'm still not able to get logs data with just using the ES endpoint URL as the host address. If that's a "just me" problem, I can update that field. Otherwise, I think I have addressed everything we have discussed if you could give it another look.

docs/en/observability/logs-stream.asciidoc
ruflin: My assumption is that it picks up changes around the path but not around the output. But I also have seen during my testing that sometimes I felt like it did not pick up changes / not full consistency but could not 100% reproduce it. @cmacknz Does this ring a bell?

docs/en/observability/logs-stream.asciidoc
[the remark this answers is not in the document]
bmorelli25: Unfortunately, Oliver no longer works at Elastic.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,114 +1,128 @@
...
+=== Step 1: Download and extract the {agent} installation package
 
 On your host, download and extract the installation package that corresponds with your system:
 
 include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/download-widget.asciidoc[]
 
+[discrete]
+[[logs-stream-install-agent]]
+=== Step 2: Install and start the {agent}
+After downloading and extracting the installation package, you're ready to install the {agent}. From the agent directory, run the the install command that corresponds with your system:
```
bmorelli25, line 33:
```suggestion
After downloading and extracting the installation package, you're ready to install the {agent}. From the agent directory, run the install command that corresponds with your system:
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,114 +1,128 @@
...
+
+NOTE: On macOS, Linux (tar package), and Windows, run the `install` command to
+install {agent} as a managed service and start the service. The DEB and RPM
+packages include a service unit for Linux systems with
+systemd, so just enable then start the service.
```
bmorelli25, line 38: I realize this was copied from the Fleet Guide, but this text is confusing to me.

> On macOS, Linux (tar package), and Windows, run the install command to install Elastic Agent as a managed service and start the service.

This sounds like you have to install and then start, but I think you only have to install, right?

> The DEB and RPM packages include a service unit for Linux systems with systemd, so just enable then start the service.

This sounds like you have to install and then start, which you do. Right?

What about this instead?

```suggestion
NOTE: On macOS, Linux (tar package), and Windows, run the `install` command to
install and start {agent} as a managed service. The DEB and RPM
packages include a service unit for Linux systems with
systemd. For these systems, you must enable and start the service.
```

If you do decide to update this snippet, please update the source file too: https://github.com/elastic/ingest-docs/blame/main/docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,114 +1,128 @@
...
+[[logs-stream-yml-location]]
+==== Locate your configuration file
 
-. Go to the {kib} home page and click *Add integrations*:
-+
---
-[role="screenshot"]
-image::images/add-integrations.png[{kib} home page]
---
+After installing the agent, you'll find the `elastic-agent.yml` in the one of the following locations according to your system:
```
bmorelli25, line 57:
```suggestion
After installing the agent, you'll find the `elastic-agent.yml` in one of the following locations according to your system:
```

docs/en/observability/logs-stream.asciidoc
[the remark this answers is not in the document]
bmorelli25: @dedemorton do you remember the history here?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,114 +1,128 @@
...
+. `hosts` – Copy the {es} endpoint from your deployment's page and add the port (the default port is `443`). For example, `https://my-deployment.es.us-central1.gcp.cloud.es.io:443`.
++
+--
+[role="screenshot"]
+image::images/es-endpoint-cluster-id.png[{es} endpoint and cluster id location, 50%]
+--
+. `api-key` – Use an API key to grant the agent access to {es}. To create an API key for your agent, see {fleet-guide}/grant-access-to-elasticsearch.html#create-api-key-standalone-agent[Create API keys for standalone agents].
 +
-NOTE: The format of the key is <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
+NOTE: The API key format should be <id>:<key>. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
```
bmorelli25, line 95:
```suggestion
NOTE: The API key format should be `<id>:<key>`. Make sure you selected *Beats* when you created your API key. Base64 encoded API keys are not currently supported in this configuration.
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,114 +1,128 @@
...
+. `inputs.id` – A unique identifier for your input.
+. `type` – The type of input. For collectings logs, set this to `filestream`.
```
bmorelli25, line 97:
```suggestion
. `type` – The type of input. For collecting logs, set this to `filestream`.
```

bmorelli25 (commented):
Looking good. A few recommendations.

bmorelli25 (approved):

Issue #3115, opened 2023-08-08
Create a "Get the most out of your log data" section


mdbirnstiehl:
### Summary
Create a story based off of [this document](https://docs.google.com/document/d/1QyB3P3jFKp3iFUJQ6MttC2YtUrhhZZhVn4CZRUrJAlQ/edit#heading=h.anrvn3dmntty) detailing how users can get more value out of their logs data once it's in Elasticsearch. 

The goal is to focus on what users want to accomplish, and not on the intricacies of our concepts. 

Initially, this will be documented on the [Stream a log file](https://www.elastic.co/guide/en/observability/current/logs-stream.html) page. 

### Story topics

Each topic should be a standalone section. A user can come to this document if they want to extract a specific field from their log data, or if they are starting from the very beginning.

See the following tasks for topics:

```[tasklist]
### Tasks
- [ ] https://github.com/elastic/observability-docs/issues/3119
- [ ] https://github.com/elastic/observability-docs/issues/3120
- [ ] https://github.com/elastic/observability-docs/issues/3121
- [ ] https://github.com/elastic/observability-docs/issues/3122
- [ ] https://github.com/elastic/observability-docs/issues/3124
- [ ] https://github.com/elastic/observability-docs/issues/3125
- [ ] https://github.com/elastic/observability-docs/issues/3126
```

PR #3118, opened 2023-08-09
add enhance logs and extract timestamp docs

mdbirnstiehl:
This PR closes [Issue 3120](https://github.com/elastic/observability-docs/issues/3120) and closes [Issue3119](https://github.com/elastic/observability-docs/issues/3119)

[Preview here](https://observability-docs_3118.docs-preview.app.elstc.co/guide/en/observability/master/logs-stream.html#logs-stream-enhance-logs)

### DIFF
--- a/docs/en/observability/logs-stream.asciidoc
+++ b/docs/en/observability/logs-stream.asciidoc
@@ -338,7 +338,7 @@ The results should show the `@timestamp` field extracted from the `message` fiel
 [[logs-stream-index-template]]
 ==== Configure your data stream with an index template
 
-After creating your ingest pipeline, you need to index your log data to a data stream using an index template. This command creates an index template:
+After creating your ingest pipeline, create an index template to point your log data to your pipeline using this command:
 
 [source,console]
 ----
--- a/docs/en/observability/logs-stream.asciidoc
+++ b/docs/en/observability/logs-stream.asciidoc
@@ -1,7 +1,7 @@
 [[logs-stream]]
 = Stream a log file
 
-In this guide, you'll learn how to send a log file to Elasticsearch using a standalone {agent}, configure the {agent} and your data streams using the `elastic-agent.yml` file, and query your logs using the data streams you've set up. 
+In this guide, you'll learn how to send a log file to Elasticsearch using a standalone {agent}, configure the {agent} and your data streams using the `elastic-agent.yml` file, and query your logs using the data streams you've set up. Once your log files are in {es}, see the <<logs-stream-enhance-logs>> section to learn how to parse your log data and extract fields so you can filter and sort your logs effectively.
 
 [discrete]
 [[logs-stream-prereq]]
@@ -133,7 +133,7 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
 
 [discrete]
 [[logs-stream-troubleshooting]]
-== Troubleshooting 
+== Troubleshoot your {agent} configuration
 
 If you're not seeing your log files in {kib}, check the following in the `elastic-agent.yml` file:
 
@@ -159,7 +159,7 @@ Add this log to {es} using the following command in *Dev Tools*, found in your d
 
 [source,console]
 ----
-POST logs-generic-default/_doc
+POST logs-example-default/_doc
 {
   "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%."
 }
@@ -169,30 +169,46 @@ Use the following command to look at the document's details:
 
 [source,console]
 ----
-GET /<index-name>/_doc/<document-id>
+GET /logs-example-default/_search
 ----
 
-You see something like the following:
+You see something like the following output:
 
 [source]
 ----
 {
-  "_index": ".ds-logs-generic-default-2023.08.09-000001",
-  "_id": "sERP24kB4jRKLxIff9f1",
-  "_version": 1,
-  "_seq_no": 4,
-  "_primary_term": 1,
-  "found": true,
-  "_source": {
-    "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.",
-    "@timestamp": "2023-08-09T17:19:27.73312243Z"
+  "took": 0,
+  "timed_out": false,
+  "_shards": {
+    "total": 1,
+    "successful": 1,
+    "skipped": 0,
+    "failed": 0
+  },
+  "hits": {
+    "total": {
+      "value": 1,
+      "relation": "eq"
+    },
+    "max_score": 1,
+    "hits": [
+      {
+        "_index": ".ds-logs-example-default-2023.08.09-000001",
+        "_id": "RsWy3IkB8yCtA5VGOKLf",
+        "_score": 1,
+        "_source": {
+          "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.",
+          "@timestamp": "2023-08-09T17:19:27.73312243Z"
+        }
+      }
+    ]
   }
 }
 ----
 
 Notice the text in your message isn't parsed, so you can't filter by any individual fields. Also, the `@timestamp` field shows when you added the data to {es}, not when the log occurred. In this format, neither of these fields are particularly useful for sorting or filtering. Your message, however, has all of these potential fields:
 
-- *timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *@timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
 - *log.level* – WARN – Extracting this field lets you filter logs by severity. This is helpful if you want to  focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
 - *host.ip* – 192.168.1.101 – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
 - *message* – Disk usage exceeds 90%.
@@ -222,7 +238,7 @@ To extract the timestamp you need to:
 
 [discrete]
 [[logs-stream-ingest-pipeline]]
-==== Extract the timestamp with an ingest pipeline
+==== Use an ingest pipeline to extract the `@timestamp`
 
 To extract the `@timestamp` field, use an ingest pipeline with a dissect processor and a date processor. Processors are executed in the order they are defined, so in the following example: 
 
@@ -233,19 +249,19 @@ The following example creates an ingest pipeline with a dissect and a date proce
 
 [source,console]
 ----
-PUT _ingest/pipeline/logs-generic-default
+PUT _ingest/pipeline/logs-example-default
 {
   "description": "Extracts the timestamp from log",
   "processors": [
     {
       "dissect": {
         "field": "message",
-        "pattern": "%{timestamp} %{message}"
+        "pattern": "%{@timestamp} %{message}"
       }
     },
     {
       "date": {
-        "field": "timestamp",
+        "field": "@timestamp",
         "formats": ["ISO8601"]
       }
     }
@@ -265,34 +281,28 @@ PUT _ingest/pipeline/logs-generic-default
 This extracts the timestamp, `2023-08-08T13:45:12.123Z`, to the `timestamp` field, and the rest of the message, `WARN 192.168.1.101 Disk usage exceeds 90%.`, stays in the `message` field.
 +
 . Set the following for the date processor:
-- `_ingest/pipeline/logs-generic-default` – `logs-generic-default` is the name of the pipeline. The name of your pipeline needs to match the name of your data stream. You'll set up your data stream in the next section. See {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme] for more information.
+- `_ingest/pipeline/logs-example-default` – `logs-example-default` is the name of the pipeline. The name of your pipeline needs to match the name of your data stream. You'll set up your data stream in the next section. See {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme] for more information.
 - `field` – The field you're extracting the date from, `@timestamp` in this case.
 - `formats` – The format of the current timestamp, `ISO8601` in this case. The date processor accepts {ref}/mapping-date-format.html[Java time patterns] or the following formats: `ISO8601`, `UNIX`, `UNIX_MS`, `TAI64N`.
 
-Optionally, you can set the following values for the `date` processor:
-
-- `timezone` – Sets the timezone to use when parsing the date.
-- `target_field` – Sets the output field for the timestamp.
-- `output_format` – Changes the format when writing the date to the timestamp field. Must be a valid java time pattern.
+You can set additional fields with the date processor to update timezones, set a different output field, and change the output format of your timestamp. For more on this, see the {ref}/date-processor.html[date processor] documentation.
 
 [discrete]
 [[logs-stream-index-template]]
 ==== Configure your data stream with an index template
 
-After creating your ingest pipeline, you need to index your log data to a data stream using an index template. 
-
-The following command creates an index template:
+After creating your ingest pipeline, you need to index your log data to a data stream using an index template. The following command creates an index template:
 
 [source,console]
 ----
-PUT _index_template/logs-generic-default-template
+PUT _index_template/logs-example-default-template
 {
-  "index_patterns": [ "logs-generic-default" ],
+  "index_patterns": [ "logs-example-default" ],
   "data_stream": { },
   "priority": 500,
   "template": {
     "settings": {
-      "index.default_pipeline":"logs-generic-default"
+      "index.default_pipeline":"logs-example-default"
     }
   }
 }
@@ -300,20 +310,20 @@ PUT _index_template/logs-generic-default-template
 
 Set the following values for the index template:
 
-- `index_patterns`– The index pattern needs to match your log data stream. Naming conventions for data streams are `<type>-<dataset>-<namespace>`. If you're using default values, your logs data stream is named `logs-generic-default`. Data that matches this pattern will go through your pipeline.
+- `index_patterns`– The index pattern needs to match your log data stream. Naming conventions for data streams are `<type>-<dataset>-<namespace>`. In this example, your logs data stream is named `logs-example-default`. Data that matches this pattern will go through your pipeline.
 - `data_stream` – Enables data streams.
-- `priority` – Index templates with higher priority take precedence over lower priority. If a data stream matches multiple index templates, the template with the higher priority is used. Built-in templates have a priority of `200`, so we recommend a priority higher than that.
-- `index.default_pipeline` – The name of the ingest pipeline you created.
+- `priority` – Index templates with higher priority take precedence over lower priority. If a data stream matches multiple index templates, the template with the higher priority is used. Built-in templates have a priority of `200`, so we recommend a priority higher than `200`.
+- `index.default_pipeline` – The name of your ingest pipeline. `logs-example-default` in this case.
 
 [discrete]
 [[logs-stream-create-data-stream]]
 ==== Create your data stream
 
-Create your data stream using the {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme]. The name needs to match the name of your pipeline. For this example, we'll use the default log data stream name, `logs-generic-default`, and the example log:
+Create your data stream using the {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme]. The name needs to match the name of your pipeline. For this example, we'll name the data stream `logs-example-default` and use the example log:
 
 [source,console]
 ----
-POST logs-generic-default/_doc
+POST logs-example-default/_doc
 {
   "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%."
 }
@@ -331,16 +341,32 @@ Under `_source`, you can see the `timestamp` field that was extracted by the dis
 [source]
 ----
 {
-  "_index": ".ds-logs-generic-default-2023.08.09-000001",
-  "_id": "StG93IkByBSsmSqxzBqO",
-  "_version": 1,
-  "_seq_no": 3,
-  "_primary_term": 1,
-  "found": true,
-  "_source": {
-    "message": "WARN 192.168.1.101 Disk usage exceeds 90%.",
-    "timestamp": "2023-08-08T13:45:12.123Z",
-    "@timestamp": "2023-08-08T13:45:12.123Z"
+  "took": 0,
+  "timed_out": false,
+  "_shards": {
+    "total": 1,
+    "successful": 1,
+    "skipped": 0,
+    "failed": 0
+  },
+  "hits": {
+    "total": {
+      "value": 1,
+      "relation": "eq"
+    },
+    "max_score": 1,
+    "hits": [
+      {
+        "_index": ".ds-logs-generic-default-2023.08.09-000001",
+        "_id": "RsWy3IkB8yCtA5VGOKLf",
+        "_score": 1,
+        "_source": {
+          "message": "WARN 192.168.1.101 Disk usage exceeds 90%.",
+          "@timestamp": "2023-08-08T13:45:12.123Z"
+        }
+      }
+      }
+    ]
   }
 }
 ----
--- a/docs/en/observability/logs-stream.asciidoc
+++ b/docs/en/observability/logs-stream.asciidoc
@@ -143,10 +143,104 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
 If you're still running into issues, see {fleet-guide}/fleet-troubleshooting.html[{agent} troubleshooting] and {fleet-guide}/elastic-agent-configuration.html[Configure standalone Elastic Agents].
 
 [discrete]
-[[logs-stream-whats-next]]
-== What's next?
+[[logs-stream-enhance-logs]]
+== Get the most out of your log data
 
-For more information on deploying and managing logs in Elastic Observability, see the following links:
+Now that you've sent your log data to {es}, this section helps you make sure you're getting the most out of it. You can do this by taking your unstructured log data and using ingest pipelines to extract structured fields. Extracting structured fields from your log data makes it easier to search, analyze, and filter your log data. 
 
-- The <<logs-checklist>> consolidates links to documentation on sending log data, configuring logs, and analyzing logs.
-- <<monitor-logs>> has information on visualizing and analyzing logs.
+Let's look at the following log data example:
+
+[source,log]
+----
+2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.
+----
+
+This log is made up of the following potential fields:
+
+- *timestamp*: "2023-08-08T13:45:12.123Z"
+- *log.level*: "WARN"
+- *host.ip*: "192.168.1.101"
+- *message*: "Disk usage exceeds 90%."
+
+Extracting these structured fields from your log data lets you search or filter your data in the following ways:
+
+- *timestamp* – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *log.level* – Extracting this field lets you filter logs by severity. This is helpful if you want to  focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
+- *host.ip* – Extracting this field lets you filter logs by the hosts' IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
+
+[discrete]
+[[logs-stream-extract-timestamp]]
+=== Extract the `@timestamp` field
+
+To extract the `@timestamp` field, you'll use a ingest pipeline with a `dissect` processor and a `date` processor. 
+
+The `dissect` processor takes your unstructured stream log message and puts it into structured fields. 
+
+The `date` processor then takes the`timestamp` field from the dissect processor and turns it into a date field. The date processor can also set the timezone and change the output format.
+
+If we look at the example log data from the previous section:
+
+[source,log]
+----
+2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.
+----
+
+For the dissect processor, you need to set the following values:
+//add link for dissect processor
+
+- the `field` you're extracting data from (`message` in this case).
+- the `pattern` of your log data. To extract the timestamp from the example, use the following pattern:
++
+[source,JSON]
+----
+%{@timestamp} %{message}
+----
++
+This puts the timestamp (2023-08-08T13:45:12.123Z) in the `@timestamp` field and the rest of the message (WARN 192.168.1.101 Disk usage exceeds 90%.) in the `message` field.
+
+For the `date` processor, you need to set following values:
+//add link for date processor
+
+- The `field` you're extracting the date from (`@timestamp` in this case).
+- The `formats` of the current timestamp (`ISO8601` in this case). The date processor also accepts Java time patterns or the following formats: `ISO8601`, `UNIX`, `UNIX_MS`, `TAI64N`.
+
+Optionally, you can set the following values for teh `date` processor:
+- The `timezone` to use when parsing the date.
+- The `output_format` if you want to change the format when writing the date to the `@timestamp` field. Must be a valid java time pattern.
+//need link for java time pattern
+
+Define your ingest pipeline with the dissect and date processors with the following:
+
+[source,JSON]
+----
+PUT _ingest/pipeline/my-pipeline
+{
+  "description": "Pipeline to extract timestamp from log data",
+  "processors": [
+    {
+      "dissect": {
+        "field": "message",
+        "patterns": "%{@timestamp} %{message}"
+      }
+    },
+    {
+      "date": {
+        "field": "@timestamp",
+        "formats": ["current-timestamp-format"],
+        "timezone": ["desired-timezone"],
+        "output_format": ["desired-timezone-output-format"]
+      }
+    }
+  ]
+}
+----
+
+[discrete]
+[[logs-stream-extract-loglevel]]
+=== Extract the `log.level` field
+
+
+
+[discrete]
+[[logs-stream-extract-hostip]]
+=== Extract the `host.ip` field
--- a/docs/en/observability/logs-stream.asciidoc
+++ b/docs/en/observability/logs-stream.asciidoc
@@ -96,8 +96,8 @@ NOTE: The API key format should be `<id>:<key>`. Make sure you selected *Beats* 
 - `inputs.id` – A unique identifier for your input.
 - `type` – The type of input. For collecting logs, set this to `filestream`.
 - `streams.id` – A unique identifier for your stream of log data. 
-- `data_stream.dataset` – The name for your dataset data stream. You can name this data stream anything that signifies the source of the data. The default value is `generic`.
-- `paths` – The path to your log files. You can also use a pattern like `/var/log/foo.log*`.
+- `data_stream.dataset` – The name for your dataset data stream. Name this data stream anything that signifies the source of the data. The default value is `generic`.
+- `paths` – The path to your log files. You can also use a pattern like `/var/log/your-logs.log*`.
 
 [discrete]
 [[logs-stream-restart-agent]]
@@ -172,26 +172,12 @@ Use the this command to look at the document's details:
 GET /logs-example-default/_search
 ----
 
-You see something like this output:
+You see something like this:
 
 [source,JSON]
 ----
 {
-  "took": 0,
-  "timed_out": false,
-  "_shards": {
-    "total": 1,
-    "successful": 1,
-    "skipped": 0,
-    "failed": 0
-  },
-  "hits": {
-    "total": {
-      "value": 1,
-      "relation": "eq"
-    },
-    "max_score": 1,
-    "hits": [
+...
       {
         "_index": ".ds-logs-example-default-2023.08.09-000001",
         "_id": "RsWy3IkB8yCtA5VGOKLf",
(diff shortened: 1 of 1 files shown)


docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,104 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+This log is made up of the following potential fields:
+
+- *timestamp*: "2023-08-08T13:45:12.123Z"
+- *log.level*: "WARN"
+- *host.ip*: "192.168.1.101"
+- *message*: "Disk usage exceeds 90%."
+
+Extracting these structured fields from your log data lets you search or filter your data in the following ways:
+
+- *timestamp* – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
```
ruflin, line 167: lets mention that these are also ECS fields which should be the same across all documents which allows correlation.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,104 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+Optionally, you can set the following values for teh `date` processor:
+- The `timezone` to use when parsing the date.
+- The `output_format` if you want to change the format when writing the date to the `@timestamp` field. Must be a valid java time pattern.
+//need link for java time pattern
+
+Define your ingest pipeline with the dissect and date processors with the following:
+
+[source,JSON]
+----
+PUT _ingest/pipeline/my-pipeline
```
ruflin, line 216: Two things here

* Before we have an ingest pipeline, lets make sure we have first an example on ingesting the log message we had above. The data stream to put it in must match the data stream naming scheme like `logs-foo-bar`
* The second point is that the ingest pipeline names we use also now must match that index, which is important as otherwise things will not fit together.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,104 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+        "field": "@timestamp",
+        "formats": ["current-timestamp-format"],
+        "timezone": ["desired-timezone"],
+        "output_format": ["desired-timezone-output-format"]
+      }
+    }
+  ]
+}
+----
+
```
ruflin, line 237: I would now ingest the document again and show the user the result. Also show the user how to query the data / see the result and post it here. Anyone reading this should be able to copy paste the examples and be successful in going through the full flow. 

Eventually users can put in their own message and go through the same iterations.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+{
+  "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%."
+}
+----
+
+Use the following command to look at the document's details:
+
+[source,console]
+----
+GET /<index-name>/_doc/<document-id>
```
mdbirnstiehl, line 172: @ruflin I'm not sure this is the best way to do this. Do you have a better suggestion here?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+    "@timestamp": "2023-08-08T13:45:12.123Z"
+  }
+}
+----
+
+You can now use the `@timestamp` field to sort your logs by the date and time they happened.
+
+[discrete]
+[[logs-stream-timestamp-troubleshooting]]
+==== Troubleshooting
```
mdbirnstiehl, line 352: @ruflin do you have any suggestions for issues users might run into?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.
+----
+
+Add this log to {es} using the following command in *Dev Tools*, found in your deployment in the left navigation under *Management*:
+
+[source,console]
+----
+POST logs-generic-default/_doc
+{
+  "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%."
```
ruflin, line 164: Great to see this working. Before 8.10 this would have thrown an error.

@felixbarny For debugging purpose, it would be nice if there was an indication on the doc that it did it.

docs/en/observability/logs-stream.asciidoc
ruflin: I would use the search command `GET /logs-generic-default/_search`. It will now only contain 1 doc and works even if the user does not know the doc id.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+  "_source": {
+    "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.",
+    "@timestamp": "2023-08-09T17:19:27.73312243Z"
+  }
+}
+----
+
+Notice the text in your message isn't parsed, so you can't filter by any individual fields. Also, the `@timestamp` field shows when you added the data to {es}, not when the log occurred. In this format, neither of these fields are particularly useful for sorting or filtering. Your message, however, has all of these potential fields:
+
+- *timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
```
ruflin, line 195:
```suggestion
- *@timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+To extract the `@timestamp` field, use an ingest pipeline with a dissect processor and a date processor. Processors are executed in the order they are defined, so in the following example: 
+
+. The {ref}/dissect-processor.html[dissect processor] extracts a structured field from your unstructured log message based on a pattern you set. In the following example, it extracts the timestamp to the `timestamp` field.
+. The {ref}/date-processor.html[date processor] then takes the `timestamp` field extracted by the dissect processor and extracts it to a date field. By default, this date field is called `@timestamp`. The date processor can also set the timezone, change the target field, and change the output format of the timestamp.
+
+The following example creates an ingest pipeline with a dissect and a date processor:
+
+[source,console]
+----
+PUT _ingest/pipeline/logs-generic-default
```
ruflin, line 236: @eyalkoren We need to name this based on the upcoming conventions. Can you chime in?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+----
++
+This extracts the timestamp, `2023-08-08T13:45:12.123Z`, to the `timestamp` field, and the rest of the message, `WARN 192.168.1.101 Disk usage exceeds 90%.`, stays in the `message` field.
++
+. Set the following for the date processor:
+- `_ingest/pipeline/logs-generic-default` – `logs-generic-default` is the name of the pipeline. The name of your pipeline needs to match the name of your data stream. You'll set up your data stream in the next section. See {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme] for more information.
+- `field` – The field you're extracting the date from, `@timestamp` in this case.
+- `formats` – The format of the current timestamp, `ISO8601` in this case. The date processor accepts {ref}/mapping-date-format.html[Java time patterns] or the following formats: `ISO8601`, `UNIX`, `UNIX_MS`, `TAI64N`.
+
+Optionally, you can set the following values for the `date` processor:
```
ruflin, line 272: as this is not needed for the current example, I would link to the docs here instead.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+
+- `timezone` – Sets the timezone to use when parsing the date.
+- `target_field` – Sets the output field for the timestamp.
+- `output_format` – Changes the format when writing the date to the timestamp field. Must be a valid java time pattern.
+
+[discrete]
```
ruflin, line 278: before we create an index template, we can use the simulate API so the user sees direct results if it will work as expected and then hook up everything together: https://www.elastic.co/guide/en/elasticsearch/reference/current/simulate-pipeline-api.html#:~:text=You%20can%20use%20the%20simulate,verbose%20parameter%20to%20the%20request.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+[[logs-stream-index-template]]
+==== Configure your data stream with an index template
+
+After creating your ingest pipeline, you need to index your log data to a data stream using an index template.
```
ruflin, line 282: the data stream already uses a template, just not the one we have created but a default one.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,220 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+{
+  "_index": ".ds-logs-generic-default-2023.08.09-000001",
+  "_id": "StG93IkByBSsmSqxzBqO",
+  "_version": 1,
+  "_seq_no": 3,
+  "_primary_term": 1,
+  "found": true,
+  "_source": {
+    "message": "WARN 192.168.1.101 Disk usage exceeds 90%.",
+    "timestamp": "2023-08-08T13:45:12.123Z",
```
ruflin, line 342: why 2 tiemstamp fields?

docs/en/observability/logs-stream.asciidoc
ruflin: invalid timestamps, wrong time zone, figuring out the format. You have listed some of the config options above. But debugging it is painful.

@felixbarny @graphaelli I always found it very painful debugging "non standard" timestamp formats and trying to get things work. I wonder if we could improve this in the future too.

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: I was testing on 8.9 and didn't get an error.

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: @ruflin I assumed this was a remnant of the dissect processor pulling a timestamp field and the date processor using that to pull the @timestamp field. Didn't realize that wasn't expected. I'll retest.

mdbirnstiehl: I updated the dissect processor to use the `@timestamp` field instead of `timestamp` to correct this.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+Now look at the details using the following command:
+
+[source,JSON]
+----
+GET /<index-name>/_doc/<document-id>
+----
+
+Under `_source`, you can see the `timestamp` field that was extracted by the dissect processor and the `@timestamp` field that was extracted by the date processor:
+
+[source]
```
bmorelli25, line 341:
```suggestion
[source,JSON]
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+Use the following command to look at the document's details:
+
+[source,console]
+----
+GET /logs-example-default/_search
+----
+
+You see something like the following output:
+
+[source]
```
bmorelli25, line 177:
```suggestion
[source,JSON]
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+
+- *@timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *log.level* – WARN – Extracting this field lets you filter logs by severity. This is helpful if you want to  focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
+- *host.ip* – 192.168.1.101 – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
+- *message* – Disk usage exceeds 90%.
+
+NOTE: These fields are part of the {ecs-ref}/ecs-reference.html[Elastic Common Schema (ECS)]. The ECS defines a common set of fields that you can use across Elasticsearch when storing data including log and metric data.
```
bmorelli25, line 216:
```suggestion
NOTE: These fields are part of the {ecs-ref}/ecs-reference.html[Elastic Common Schema (ECS)]. The ECS defines a common set of fields that you can use across Elasticsearch when storing data, including log and metric data.
```

I'd add a comma here but up to you

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+=== Extract the `@timestamp` field
+
+This section shows you how to extract the `@timestamp` field from the following example log:
+
+[source,log]
+----
+2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.
+----
+
+To extract the timestamp you need to:
```
bmorelli25, line 234:
```suggestion
To extract the timestamp you need to:

```
Need a new line here to render the unordered list correctly

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+
+- *@timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *log.level* – WARN – Extracting this field lets you filter logs by severity. This is helpful if you want to  focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
+- *host.ip* – 192.168.1.101 – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
+- *message* – Disk usage exceeds 90%.
+
+NOTE: These fields are part of the {ecs-ref}/ecs-reference.html[Elastic Common Schema (ECS)]. The ECS defines a common set of fields that you can use across Elasticsearch when storing data including log and metric data.
+
+See the following sections for information on extracting specific fields:
```
bmorelli25, line 218:
```suggestion
See the following sections for information on extracting specific fields:

```

Need a new line here to render the unordered list correctly

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
 If you're still running into issues, see {fleet-guide}/fleet-troubleshooting.html[{agent} troubleshooting] and {fleet-guide}/elastic-agent-configuration.html[Configure standalone Elastic Agents].
 
 [discrete]
-[[logs-stream-whats-next]]
-== What's next?
+[[logs-stream-enhance-logs]]
+== Get the most out of your log data
 
-For more information on deploying and managing logs in Elastic Observability, see the following links:
+This section shows you how to get the most out of your log data after sending it to {es} by extracting structured fields from your unstructured log data. Extracting structured fields makes it easier to search, analyze, and filter your log data.
```
bmorelli25, line 149: > This section shows you how to get the most out of your log data after sending it to {es} by extracting structured fields from your unstructured log data.

I can't explain why, but I don't like this sentence. Here's a suggestion. You don't have to take it verbatim but maybe think more about how we open this section.

```
== Get the most out of your log data

Make your logs more useful by extracting structured fields from your unstructured log data.
Extracting structured fields makes it easier to search, analyze, and filter your log data. 
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+  }
+}
+----
+
+Notice the text in your message isn't parsed, so you can't filter by any individual fields. Also, the `@timestamp` field shows when you added the data to {es}, not when the log occurred. In this format, neither of these fields are particularly useful for sorting or filtering. Your message, however, has all of these potential fields:
+
+- *@timestamp* – 2023-08-08T13:45:12.123Z – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *log.level* – WARN – Extracting this field lets you filter logs by severity. This is helpful if you want to  focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
+- *host.ip* – 192.168.1.101 – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
+- *message* – Disk usage exceeds 90%.
```
bmorelli25, line 214:
```suggestion
- *@timestamp* – `2023-08-08T13:45:12.123Z` – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
- *log.level* – `WARN` – Extracting this field lets you filter logs by severity. This is helpful if you want to focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
- *host.ip* – `192.168.1.101` – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
- *message* – `Disk usage exceeds 90%`.
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+- <<logs-stream-index-template>>
+- <<logs-stream-create-data-stream>>
+
+[discrete]
+[[logs-stream-ingest-pipeline]]
+==== Use an ingest pipeline to extract the `@timestamp`
+
+To extract the `@timestamp` field, use an ingest pipeline with a dissect processor and a date processor. Processors are executed in the order they are defined, so in the following example: 
+
+. The {ref}/dissect-processor.html[dissect processor] extracts a structured field from your unstructured log message based on a pattern you set. In the following example, it extracts the timestamp to the `timestamp` field.
```
bmorelli25, line 245: I think we're saying "In the following example" too much here

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+. The {ref}/date-processor.html[date processor] then takes the `timestamp` field extracted by the dissect processor and extracts it to a date field. By default, this date field is called `@timestamp`. The date processor can also set the timezone, change the target field, and change the output format of the timestamp.
+
+The following example creates an ingest pipeline with a dissect and a date processor:
```
bmorelli25, line 248: oop, here it is again

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
++
+[source,JSON]
+----
+%{timestamp} %{message}
+----
++
+This extracts the timestamp, `2023-08-08T13:45:12.123Z`, to the `timestamp` field, and the rest of the message, `WARN 192.168.1.101 Disk usage exceeds 90%.`, stays in the `message` field.
++
+. Set the following for the date processor:
+- `_ingest/pipeline/logs-example-default` – `logs-example-default` is the name of the pipeline. The name of your pipeline needs to match the name of your data stream. You'll set up your data stream in the next section. See {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme] for more information.
```
bmorelli25, line 284: With just a dash between two code snippets, it's a little tough to see what you're saying to do here. Not sure how to improve this though. Maybe it's fine

docs/en/observability/logs-stream.asciidoc
```diff
@@ -143,10 +143,246 @@ If you're not seeing your log files in {kib}, check the following in the `elasti
...
+- `pattern`– The pattern of the elements in your log data:
++
+[source,JSON]
+----
+%{timestamp} %{message}
+----
++
+This extracts the timestamp, `2023-08-08T13:45:12.123Z`, to the `timestamp` field, and the rest of the message, `WARN 192.168.1.101 Disk usage exceeds 90%.`, stays in the `message` field.
++
+. Set the following for the date processor:
```
bmorelli25, line 283: I'd change this to

```suggestion
. Set the following parameters for the date processor:
```

bmorelli25: It looks like you say `Set the following values...` in an example below this. That works too. I'd update this to be consistent.

bmorelli25 (commented):
First round of feedback for ya!

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,299 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+      "doc": {
+        "_index": "_index",
+        "_id": "_id",
+        "_version": "-3",
+        "_source": {
+          "message": "WARN 192.168.1.101 Disk usage exceeds 90%.",
+          "@timestamp": "2023-08-08T13:45:12.123Z"
+        },
+        "_ingest": {
+          "timestamp": "2023-08-10T20:37:29.090624429Z"
```
ruflin, line 327: Was this timestamp automatically created based on the above pipeline?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,299 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+
+[discrete]
+[[logs-stream-index-template]]
+==== Configure your data stream with an index template
+
+After creating your ingest pipeline, create an index template to point your log data to your pipeline using this command:
+
+[source,console]
+----
+PUT _index_template/logs-example-default-template
```
ruflin, line 345: @felixbarny @felixbarny If someone is building their own dataset, I assume they should not use `@custom` but the (future) `@default`?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,299 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+----
+GET /<index-name>/_doc/<document-id>
+----
+
+Under `_source`, you can see the `timestamp` field that was extracted by the dissect processor and the `@timestamp` field that was extracted by the date processor:
+
+[source,JSON]
+----
+{
+  "took": 0,
```
ruflin, line 391: For these kind of result documents, I suggest we filter out all the "non important" parts like `_shards` total hits etc. to make it focused on what matters in this case. Or do you think that is more confusing?

How can we indicate there is more content? For example using `...`

bmorelli25: ++ we do this elsewhere in the docs too

mdbirnstiehl: I was thinking about this as well. I'll see if we have any examples in our current docs to see if we have a consistent approach for this situation.

mdbirnstiehl: I guess I was commenting on an outdated version of the PR.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,270 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+        "_id": "RsWy3IkB8yCtA5VGOKLf",
+        "_score": 1,
+        "_source": {
+          "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.",
+          "@timestamp": "2023-08-09T17:19:27.73312243Z"
+        }
+      }
+    ]
+  }
+}
```
bmorelli25, line 192: I'm nitpicking here, but when shortening these, I think it's important to maintain the hierarchy as much as possible. The content we're interested in is `hits.hits[0].*`, so I'd shorten like this instead: 

```suggestion
{
  ...
  "hits": {
    ...
    "hits": [
      {
        "_index": ".ds-logs-example-default-2023.08.09-000001",
        "_id": "RsWy3IkB8yCtA5VGOKLf",
        "_score": 1,
        "_source": {
          "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.",
          "@timestamp": "2023-08-09T17:19:27.73312243Z"
        }
      }
    ]
  }
}
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,270 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+        "_id": "RsWy3IkB8yCtA5VGOKLf",
+        "_score": 1,
+        "_source": {
+          "message": "WARN 192.168.1.101 Disk usage exceeds 90%.",
+          "@timestamp": "2023-08-08T13:45:12.123Z"
+        }
+      }
+    ]
+  }
+}
```
bmorelli25, line 389: See above comment

```suggestion
{
  ...
  "hits": {
    ...
    "hits": [
      {
        "_index": ".ds-logs-generic-default-2023.08.09-000001",
        "_index": ".ds-logs-example-default-2023.08.09-000001",
        "_id": "RsWy3IkB8yCtA5VGOKLf",
        "_score": 1,
        "_source": {
          "message": "WARN 192.168.1.101 Disk usage exceeds 90%.",
          "@timestamp": "2023-08-08T13:45:12.123Z"
        }
      }
      }
    ]
  }
}
```

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: Yeah, this was the result I got when running the simulate pipeline api.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+
+You can now use the `@timestamp` field to sort your logs by the date and time they happened.
+
+[discrete]
+[[logs-stream-timestamp-troubleshooting]]
+=== Troubleshoot your `@timestamp` field
+
+Check the following common issues for possible solutions:
+
+- *Timestamp failure* – If your data has inconsistent date formats, you can set `ignore_failure` to `true` for your date processor. This process logs with correctly formatted dates and ignore those with issues.
```
bmorelli25, line 409:
```suggestion
- *Timestamp failure* – If your data has inconsistent date formats, you can set `ignore_failure` to `true` for your date processor. This processes logs with correctly formatted dates and ignores those with issues.
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+- *Incorrect timezone* – Set your timezone using the `timezone` option on the {ref}/date-processor.html[date processor].
+- *Incorrect timestamp format* – Your timestamp can be a java time pattern or one of the following formats: ISO8601, UNIX, UNIX_MS, or TAI64N. See the {ref}/mapping-date-format.html[mapping date format] for more information on timestamp formats.
```
bmorelli25, line 411:
```suggestion
- *Incorrect timestamp format* – Your timestamp can be a Java time pattern or one of the following formats: ISO8601, UNIX, UNIX_MS, or TAI64N. See the {ref}/mapping-date-format.html[mapping date format] for more information on timestamp formats.
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+- `index_patterns`– The index pattern needs to match your log data stream. Naming conventions for data streams are `<type>-<dataset>-<namespace>`. In this example, your logs data stream is named `logs-example-default`. Data that matches this pattern will go through your pipeline.
+- `data_stream` – Enables data streams.
+- `priority` – Index templates with higher priority take precedence over lower priority. If a data stream matches multiple index templates, the template with the higher priority is used. Built-in templates have a priority of `200`, so we recommend a priority higher than `200`.
+- `index.default_pipeline` – The name of your ingest pipeline. `logs-example-default` in this case.
+
+[discrete]
+[[logs-stream-create-data-stream]]
+==== Create your data stream
+
+Create your data stream using the {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme]. The name needs to match the name of your pipeline. For this example, we'll name the data stream `logs-example-default` and use the example log:
```
bmorelli25, line 358: This link doesn't render

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+
+[source,console]
+----
+POST logs-example-default/_doc
+{
+  "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%."
+}
+----
+
+Now look at the details using this command:
```
bmorelli25, line 368: Look at the details of what?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+. Set these values for the dissect processor:
+- `field` – The field you're extracting data from, `message` in this case.
+- `pattern`– The pattern of the elements in your log data:
++
+[source,JSON]
+----
+%{timestamp} %{message}
+----
++
+This extracts the timestamp, `2023-08-08T13:45:12.123Z`, to the `timestamp` field, and the rest of the message, `WARN 192.168.1.101 Disk usage exceeds 90%.`, stays in the `message` field.
```
bmorelli25, line 271: and to while? idk

```suggestion
This extracts the timestamp, `2023-08-08T13:45:12.123Z`, to the `timestamp` field, while the rest of the message, `WARN 192.168.1.101 Disk usage exceeds 90%.`, stays in the `message` field.
```

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
++
+. Set these values for the date processor:
+- `_ingest/pipeline/logs-example-default` – The name of the pipeline,`logs-example-default`, needs to match the name of your data stream. You'll set up your data stream in the next section. See {fleet-ref}/data-streams.html#data-streams-naming-scheme[data stream naming scheme] for more information.
```
bmorelli25, line 274: this link doesn't render

bmorelli25 (commented):
More suggestions for ya

docs/en/observability/logs-stream.asciidoc
ruflin: Interesting. Lets leave it as this. Will dig into this on my end.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+----
+{
+...
+{
+  ...
+  "hits": {
+    ...
+    "hits": [
+      {
+        "_index": ".ds-logs-generic-default-2023.08.09-000001",
```
ruflin, line 387: This seems off?

ruflin (commented):
I suggest we get this in rather soonish and then start to iterate on it. 

What I like is that this is all on a single page in the docs. What is unfortunate, that it seems this means users don't get a side navigation for it to directly jump to a section. Or is there a way to haven ankers on the side without it being split into multiple pages?

docs/en/observability/logs-stream.asciidoc
eyalkoren: IIUC, this example is about registering a custom index templates with higher priority to override the default one. In such cases, neither `@custom` nor `@default` should be used and the only recommendation is to avoid using `@`. If that's the case, this looks right as it is

ruflin: What the user is building here is their custom dataset without having an instance of an integration. My assumption, a user could convert it to it eventually. 

As it is the user owned dataset, I'm tempted to recommend to use `@default`. It means, if someone else would want to extend the dataset could use `@custom`. 

> In such cases, neither @custom nor @default should be used and the only recommendation is to avoid using @. If that's the case, this looks right as it is

@eyalkoren In the context of the data stream naming scheme, should "random" templates ever be used? @joshdover How would Fleet keep track of these?

@weltenwort Also bringing you in here as I think this discussion could become interesting in the context of creating integrations.

@mdbirnstiehl Please don't block on this conversation, we can always update it.

eyalkoren: > @eyalkoren In the context of the data stream naming scheme, should "random" templates ever be used?

Again, my understanding is that this depends on who owns/registers the templates. The way I see it, the naming scheme is only relevant for templates managed by Elastic (never mind which component). This frees us from keeping track of anything else. We make best efforts to prevent clashes **between custom components and such managed by Elastic** by documenting that `@` should be avoided, but we are not eligible for preventing clashes between custom components. I hope this makes sense.

felixbarny: In the future, I think that ideally users shouldn't need to create an index template but rather hook into extension points for the managed index templates. See also https://github.com/elastic/elasticsearch/issues/97664.

But for the time being, creating a custom `logs-example-default-template` is fine. The only thing I'd change here is to use the built-in index template for `logs-*-*` as a foundation for the custom index template. The built-in template can be retrieved via `GET _index_template/logs`.

docs/en/observability/logs-stream.asciidoc
felixbarny: > @felixbarny For debugging purpose, it would be nice if there was an indication on the doc that it did it.

I don't disagree but I'm not sure whether adding additional storage for that additional piece of debugging metadata is worth it.

ruflin (approved):

docs/en/observability/logs-stream.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[logs-stream]]
 = Stream a log file
 
-In this guide, you'll learn how to send a log file to Elasticsearch using a standalone {agent}, configure the {agent} and your data streams using the `elastic-agent.yml` file, and query your logs using the data streams you've set up. 
+In this guide, you'll learn how to send a log file to Elasticsearch using a standalone {agent}, configure the {agent} and your data streams using the `elastic-agent.yml` file, and query your logs using the data streams you've set up. Once your log files are in {es}, see the <<logs-stream-enhance-logs>> section to learn how to parse your log data and extract fields so you can filter and sort your logs effectively.
 
 [discrete]
 [[logs-stream-prereq]]
```
bmorelli25, line 7: (continued from Slack)

Float (or discrete) these at L1 headings so that you can bump the heading level up for every topic below this

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+- <<logs-stream-create-data-stream>>
+
+[discrete]
+[[logs-stream-ingest-pipeline]]
+==== Use an ingest pipeline to extract the `@timestamp`
+
+To extract the `@timestamp` field, use an ingest pipeline with a dissect processor and a date processor. Processors are executed in the order they are defined. The example pipeline runs these processors: 
+
+. The {ref}/dissect-processor.html[dissect processor] extracts a structured field from your unstructured log message based on a pattern you set. In the following example, the dissect processor extracts the timestamp to the `timestamp` field.
+. The {ref}/date-processor.html[date processor] then takes the `timestamp` field extracted by the dissect processor and extracts it to a date field. By default, this date field is called `@timestamp`. The date processor can also set the timezone, change the target field, and change the output format of the timestamp.
```
felixbarny, line 236: I don't think the date processor is required in this case. ES will be able to parse string timestamps. According to https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html, the default format for `date` fields is `strict_date_optional_time||epoch_millis`. According to https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html, `strict_date_optional_time` supports the formats `yyyy-MM-dd'T'HH:mm:ss.SSSZ` or ` yyyy-MM-dd`. As the timestamp in the log file already uses that format, the date processor is not necessary.

However, if the date looks a bit more complex and is non-standard, the date processor is certainly necessary.

I think in this case, I'd remove the date processor but add a link to the date processor docs and explain that it can be useful for parsing dates that are not in ISO or format epoch_millis.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -133,20 +133,279 @@ image::images/stream-logs-example.png[example search query on the logs stream pa
...
+- *@timestamp* – `2023-08-08T13:45:12.123Z` – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *log.level* – `WARN` – Extracting this field lets you filter logs by severity. This is helpful if you want to  focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
+- *host.ip* – `192.168.1.101` – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
+- *message* – `Disk usage exceeds 90%.`
+
+NOTE: These fields are part of the {ecs-ref}/ecs-reference.html[Elastic Common Schema (ECS)]. The ECS defines a common set of fields that you can use across Elasticsearch when storing data, including log and metric data.
+
+See these sections information on extracting specific fields:
+
+- <<logs-stream-extract-timestamp>>
```
felixbarny, line 209: Are we planning to add more sections to extract the other fields, such as `log.level` and `host.ip`?

mdbirnstiehl: Yep, goal is to merge this shortly and continue to "Extract the `log.level`". This [meta issue](https://github.com/elastic/observability-docs/issues/3115) outlines the plan.

bmorelli25 (approved):
🚢

docs/en/observability/logs-stream.asciidoc
joshdover: > How would Fleet keep track of these?

We haven't designed how someone would do this conversion from a existing index template and ingest pipeline to a proper package, so I can only speculate at this point.

The way I'd imagine it working though is that we could extract any settings defined at the top level and move them into the `@package` settings later (strictly speaking, we'd put them the package-spec definition, in the data stream's `manifest.yml` `elasticsearch.settings` block)

ruflin: I think there should be an option (eventually) to add a dataset to an integration. Not sure we can make this work if the dataset itself doesn't already follow the convention around naming.

@weltenwort Have you thought more about this?

weltenwort: I tend to agree with what @joshdover hinted at:

- Guidance for the "manual quick-start" approach avoids any names that might collide with integration-managed assets (i.e. avoid `@`s).
- The "uplift into an integration" process would pick settings from the user-created assets and use them to create proper integration-managed assets. If those are ambiguous or conflicting we could let the user choose during the workflow.

The advantage of this approach would be that we don't front-load all the naming knowledge, which could increase the hurdle to get data in. It would also mean the instructions wouldn't have to change frequently as our conventions evolve. We "just" need to adapt our conversion routine.

PR #3135, opened 2023-08-14
[DOCS] Remove content reuse for cases

lcawl:
This PR removes a tagged region to err on the side of simplicity as we migrate our content.


bmorelli25 (commented):
Oh wow. Is this really our only Kibana include? That's awesome. I'm going to push a quick update to this PR

lcawl:
> Oh wow. Is this really our only Kibana include? That's awesome. I'm going to push a quick update to this PR

Yeah, I originally intended to re-use more but that seems like the wrong strategy now so we might as well yank it!

Issue #3116, opened 2023-08-08
Create a logs troubleshooting page


mdbirnstiehl:
Create a logs troubleshooting page that includes the errors that users might encounter during onboarding. Use this [document](https://docs.google.com/document/d/193N-IH8wG8Ipra0gNpa1-sceB5xefDMZ6qReZIhs7hI/edit) as a resource.

PR #3157, opened 2023-08-21
Add onboarding troubleshooting

mdbirnstiehl:
This PR closes [Issue 3116](https://github.com/elastic/observability-docs/issues/3116)

I've updated the **Troubleshoot your logs** page to include the information from @yngrdyn 's [word doc](https://docs.google.com/document/d/193N-IH8wG8Ipra0gNpa1-sceB5xefDMZ6qReZIhs7hI/edit). I'm adding this as a draft PR because I think there is still room for more discussion and information on what needs to be added. But I thought this would give us a starting point for discussion.

### DIFF
--- a/docs/en/observability/logs-troubleshooting.asciidoc
+++ b/docs/en/observability/logs-troubleshooting.asciidoc
@@ -1,4 +1,111 @@
 [[logs-troubleshooting]]
 = Troubleshoot your logs
 
-_coming soon_
+This section provides solutions to errors you might encounter while onboarding your logs in {kib}.
+
+[discrete]
+[[logs-troubleshooting-insufficient-priv]]
+== User does not have permissions to create API key
+
+If you don't have the required privileges to create an API key, you'll see the following error message:
+
+----
+User does not have permissions to create API key.
+
+Required cluster privileges are [`monitor`, `manage_own_api_key`] and 
+required index privileges are [`auto_configure`, `create_doc`] for 
+indices [`logs-*-*`, `metrics-*-*`], please add all required privileges 
+to the role of the authenticated user.
+----
+
+*Solution:*
+
+You need an administrator to give you the necessary permissions listed in the error messages and then to restart the onboarding flow.
+
+[discrete]
+[[logs-troubleshooting-API-key-failed]]
+== Failed to create API key
+
+If you don't have the privileges to create `savedObjects` in {kib}, you'll see the following error message:
+
+----
+Failed to create API key
+
+Something went wrong: Unable to create observability-onboarding-state
+----
+
+*Solution:*
+
+You need an administrator to give you privileges to create `savedObjects` in {kib} to generate the required `observabilityOnboarding` flow state.
+Once you have necessary privileges, restart the onboarding flow.
+
+[discrete]
+[[logs-troubleshooting-kib-not-accessible]]
+== {kib} not accessible from host
+
+If {kib} is not accessible from the host, you'll see the following error message after pasting the *Install the {agent}* instructions your host:
+
+----
+New integrated terminal
+
+To set your external terminal as the default, change your preferences.
+----
+
+*Solution:*
+The host needs access to {Kib}. If that's not possible, you can manually download and configure the {agent}. See the <<logs-stream, Stream a log file>> documentation for more information.
+
+[discrete]
+[[logs-troubleshooting-download-agent]]
+== Download {agent} failed
+
+If the host was able to download the installation script but cannot connect to the public artifactory repository, you'll see the following error message:
+
+----
+Download Elastic Agent
+
+Failed to download Elastic Agent, see script for error.
+----
+
+*Solutions:*
+
+* **Agent version and OS architecture is not available*
++
+The combination of {agent} version and OS architecture is not available. If this is the case you'll see the following error message on your host:
++
+----
+New integrated terminal
+
+To set your external terminal as the default, change your preferences.
+----
++
+To fix this, modify the {agent} version in the installation instructions to a known version of the {agent}.
+
+* You're an Elastic Cloud Enterprise user without access to the Elastic downloads page.
+* The {agent} was fully downloaded previously. In this case, delete previous downloads and restart the onboarding.
+
+[discrete]
+[[logs-troubleshooting-install-agent]]
+== Install {agent} failed
+
+If an {agent} already exists on your host, you'll see the following error message:
+
+----
+Install Elastic Agent
+
+Failed to install Elastic Agent, see script for error.
+----
+
+*Solution:*
+You can uninstall the current {agent} using the `elastic-agent uninstall` command, and run the script again.
+
+WARNING: Uninstalling the current {agent} removes the entire current setup including the existing configuration. 
+
+[discrete]
+[[logs-troubleshooting-wait-for-logs]]
+== Waiting for Logs to be shipped... step never completes
+
+If the *Waiting for Logs to be shipped...* step never completes, logs are not being shipped to {es}, and there is most likely an issue with your {agent} configuration. 
+
+*Solution:*
+
+See the {fleet-guide}/debug-standalone-agents.html#inspect-standalone-agent-logs[Debug standalone {agent}s] documentation for information on checking {agent} logs to look for errors.
--- a/docs/en/observability/logs-troubleshooting.asciidoc
+++ b/docs/en/observability/logs-troubleshooting.asciidoc
@@ -1,7 +1,7 @@
 [[logs-troubleshooting]]
 = Troubleshoot your logs
 
-This section provides solutions to errors you might encounter while onboarding your logs.
+This section provides possible solutions for errors you might encounter while onboarding your logs.
 
 [discrete]
 [[logs-troubleshooting-insufficient-priv]]
@@ -25,8 +25,8 @@ to the role of the authenticated user.
 
 You need to either:
 
-* Get the necessary privileges listed in the error messages from an administrator, and restart the onboarding flow.
-* Get an API key from an administrator and manually add the API to the {agent} configuration. See <<logs-stream-agent-config, Configure the {agent}>> for more on manually updating the configuration and adding the API.
+* Have an administrator give you the `monitor` and `manage_own_api_key` cluster privileges and the `auto_configure` and `create_doc` indices privileges. Once you have these privileges, restart the onboarding flow.
+* Get an API key from an administrator and manually add the API to the {agent} configuration. See <<logs-stream-agent-config, Configure the {agent}>> for more on manually updating the configuration and adding the API key.
 
 [discrete]
 [[logs-troubleshooting-API-key-failed]]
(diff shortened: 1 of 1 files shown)


docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
 [[logs-troubleshooting]]
 = Troubleshoot your logs
 
-_coming soon_
\ No newline at end of file
+This section provides solutions to errors you might encounter while onboarding your logs in {kib}.
```
ruflin, line 4: I stumbled over `in {kib}`. Do you onboard your logs in Kibana? To Elasticsearch? To the Elastic Stack? How does a new user think here?

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+
+Required cluster privileges are [`monitor`, `manage_own_api_key`] and 
+required index privileges are [`auto_configure`, `create_doc`] for 
+indices [`logs-*-*`, `metrics-*-*`], please add all required privileges 
+to the role of the authenticated user.
+----
+
+*Solution:*
+
+You need an administrator to give you the necessary permissions listed in the error messages and then to restart the onboarding flow.
```
ruflin, line 23: An additional solution is that the user gets an API key from an admin and then does the step manually?

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+
+----
+Failed to create API key
+
+Something went wrong: Unable to create observability-onboarding-state
+----
+
+*Solution:*
+
+You need an administrator to give you privileges to create `savedObjects` in {kib} to generate the required `observabilityOnboarding` flow state.
```
ruflin, line 39: To make this more actionable, could we share here the API call that the admin would have to run?

docs/en/observability/logs-troubleshooting.asciidoc
ruflin, line 23: To make this more actionable, could we share here the API call that the admin would have to make? Otherwise the admin comes back to the user and asks what exactly they need to do.

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+Once you have necessary privileges, restart the onboarding flow.
+
+[discrete]
+[[logs-troubleshooting-kib-not-accessible]]
+== {kib} not accessible from host
+
+If {kib} is not accessible from the host, you'll see the following error message after pasting the *Install the {agent}* instructions your host:
+
+----
+New integrated terminal
```
ruflin, line 49: Is that the error you get? If this is expected, seems very abstract.

docs/en/observability/logs-troubleshooting.asciidoc
mdbirnstiehl: My thought was that these current troubleshooting steps are mostly for errors users encounter while using the Kibana onboarding, so I was trying to be more specific there. Maybe even the title of this doc can be Logs onboarding troubleshooting until it is filled out with more generic troubleshooting information.

docs/en/observability/logs-troubleshooting.asciidoc
mdbirnstiehl: @ruflin Could someone provide this API call? Not sure exactly how to format that.

docs/en/observability/logs-troubleshooting.asciidoc
mdbirnstiehl: @ruflin Same as above. Could someone provide this API call.

docs/en/observability/logs-troubleshooting.asciidoc
mdbirnstiehl: According to the word doc provided. @yngrdyn is this still the cases?

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
\ No newline at end of file
+This section provides solutions to errors you might encounter while onboarding your logs in {kib}.
+
+[discrete]
+[[logs-troubleshooting-insufficient-priv]]
+== User does not have permissions to create API key
+
+If you don't have the required privileges to create an API key, you'll see the following error message:
+
+----
```
bmorelli25, line 12: All of these code blocks need `[source,*]` to render correctly. Maybe `[source,plaintext]`? A full list of options is available [here](https://github.com/rouge-ruby/rouge/blob/master/docs/Languages.md) if you want to look through them.

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+*Solution:*
+
+You need an administrator to give you privileges to create `savedObjects` in {kib} to generate the required `observabilityOnboarding` flow state.
+Once you have necessary privileges, restart the onboarding flow.
+
+[discrete]
+[[logs-troubleshooting-kib-not-accessible]]
+== {kib} not accessible from host
+
+If {kib} is not accessible from the host, you'll see the following error message after pasting the *Install the {agent}* instructions your host:
```
bmorelli25, line 46:
```suggestion
If {kib} is not accessible from the host, you'll see the following error message after pasting the *Install the {agent}* instructions into your host:
```

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+----
+
+*Solution:*
+The host needs access to {Kib}. If that's not possible, you can manually download and configure the {agent}. See the <<logs-stream, Stream a log file>> documentation for more information.
+
+[discrete]
+[[logs-troubleshooting-download-agent]]
+== Download {agent} failed
+
+If the host was able to download the installation script but cannot connect to the public artifactory repository, you'll see the following error message:
```
bmorelli25, line 61: Is artifactory a word?

docs/en/observability/logs-troubleshooting.asciidoc
ruflin: @weltenwort Could someone in your team help out here?

docs/en/observability/logs-troubleshooting.asciidoc
ruflin: @weltenwort ^

docs/en/observability/logs-troubleshooting.asciidoc
weltenwort: Are we looking for an API call to grant the privileges? Or instructions to assign them via Kibana's management UI?

weltenwort: Hm, that's a bit tricky as much of the call depends on the specific user and its roles. They'd need to add these privileges to one of the roles of the current user, but we don't know how they've structured their roles. And if the user doesn't have these privileges, they probably don't have the privilege to grant new privileges either.

Would it maybe make sense to ship with a default observability role that grants the privileges and instruct them to ask the admin to grant that role?

docs/en/observability/logs-troubleshooting.asciidoc
yngrdyn: I think this is not the proper error should be more align to:

```
Failed to connect to {host} port {port} after 0 ms: Connection refused
```

docs/en/observability/logs-troubleshooting.asciidoc
yngrdyn: Maybe not 🤔 
What about `public artifacts repository`

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+----
+New integrated terminal
+
+To set your external terminal as the default, change your preferences.
+----
++
+To fix this, modify the {agent} version in the installation instructions to a known version of the {agent}.
+
+* You're an Elastic Cloud Enterprise user without access to the Elastic downloads page.
+* The {agent} was fully downloaded previously. In this case, delete previous downloads and restart the onboarding.
```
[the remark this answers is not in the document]
yngrdyn: Here the error in the console should be more align to
```
The requested URL returned error: 404
```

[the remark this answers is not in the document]
yngrdyn: Concretely this is what we are showing
image

What do you suggest @ruflin ?

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,111 @@
...
+----
+
+*Solutions:*
+
+* *Agent version and OS architecture is not available*
++
+The combination of {agent} version and OS architecture is not available. If this is the case you'll see the following error message on your host:
++
+----
+New integrated terminal
```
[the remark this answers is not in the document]
mdbirnstiehl: @yngrdyn do you have a different error for this one as well?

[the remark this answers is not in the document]
mdbirnstiehl: Never mind, I see it below.

docs/en/observability/logs-troubleshooting.asciidoc
mdbirnstiehl: @ruflin any suggestion here? How do we want to move forward with the docs?

ruflin: Lets take this on in steps. Lets first document the permissions that are needed. From there, we can discuss if we should have a role for it and API calls.

docs/en/observability/logs-troubleshooting.asciidoc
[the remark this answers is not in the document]
ruflin: If the agent was already downloaded also the above error shows? I was referring to:

```
The {agent} was fully downloaded previously. In this case, delete previous downloads and restart the onboarding.
```

I'm surprised we see a 404 in this case.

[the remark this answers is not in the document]
yngrdyn: ah no, the 404 it's in case the url for the agent returns the 404 or we cannot connect to kibana to download the script.

In the case of an agent already existing in the host, we would probably would see something like
```
Error: already installed at: /opt/Elastic/Agent
```

Let me double check this scenario

mdbirnstiehl:
@ruflin @yngrdyn I think I have made the recommended changes. I've updated the privileges for the for now in the places where we may add API calls. If you could have another look through, we can move forward with this for now, and update as we go on.

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,133 @@
...
+To fix this, update the {agent} version in the installation instructions to a known version of the {agent}. 
+* If the {agent} was fully downloaded previously, you'll see the following error message:
++
+[source, plaintext]
+----
+Error: cannot perform installation as Elastic Agent is already running from this directory
+----
++
+To fix this, delete previous downloads and restart the onboarding.
+* You're an Elastic Cloud Enterprise user without access to the Elastic downloads page.
```
mdbirnstiehl, line 101: Not sure if there is a solution to this one that we can add as it's more of a cause at this point.

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,133 @@
...
+[[logs-troubleshooting-API-key-failed]]
+== Failed to create API key
+
+If you don't have the privileges to create `savedObjects` in {kib}, you'll see the following error message:
+
+[source, plaintext]
+----
+Failed to create API key
+
+Something went wrong: Unable to create observability-onboarding-state
```
ruflin, line 41: This is a comment for the devs, not docs: We should improve this error so users get already details in the error on what to do.

docs/en/observability/logs-troubleshooting.asciidoc
```diff
@@ -1,4 +1,133 @@
...
+
+[discrete]
+[[logs-troubleshooting-kib-not-accessible-solution]]
+=== Solution
+
+The host needs access to {Kib}. If that's not possible, you can manually download and configure the {agent}. See the <<logs-stream, Stream a log file>> documentation for more information.
+
+[discrete]
+[[logs-troubleshooting-download-agent]]
+== Download {agent} failed
```
ruflin, line 70: Happy to get this section in but I think we should revisit it. It seems odd that someone could have a link for a not supported platform in the first place?

ruflin (approved):
Lets get it in and as usual, iterate on it.

Issue #3125, opened 2023-08-10
Create a "Routing" section


mdbirnstiehl:
This [blog ](https://www.elastic.co/blog/simplifying-log-data-management-flexible-routing-elastic)is a resource for this section.

PR #3165, opened 2023-08-24
Add reroute processor to stream a log page

mdbirnstiehl:
This PR closes [Issue 3125](https://github.com/elastic/observability-docs/issues/3125) and adds a section on using the reroute processor to simplify log data management.

### DIFF
--- a/docs/en/observability/logs-stream.asciidoc
+++ b/docs/en/observability/logs-stream.asciidoc
@@ -880,7 +880,7 @@ preview::[]
 
 By default, an ingest pipeline sends your log data to a single data stream. To simplify log data management, you can use a {ref}/reroute-processor.html[reroute processor] to route data from the generic data stream to a target data stream. For example, you might want to send high-severity logs to a specific data stream that's different from low-severity logs to help with categorization. 
 
-This section shows you how to use a reroute processor to send the high-severity logs (`WARN` or `ERROR`) from the following examples to a specific data stream and keep regular logs (`DEBUG` and `INFO`) in the default data stream:
+This section shows you how to use a reroute processor to send the high-severity logs (`WARN` or `ERROR`) from the following log examples to a specific data stream and keep regular logs (`DEBUG` and `INFO`) in the default data stream:
 
 [source,log]
 ----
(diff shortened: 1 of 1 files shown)


docs/en/observability/logs-stream.asciidoc
```diff
@@ -855,4 +870,141 @@ PUT /logs-example-default/_mapping
...
+  "description": "Extracts fields and reroutes WARN",
+  "processors": [
+    {
+      "dissect": {
+        "field": "message",
+        "pattern": "%{@timestamp} %{log.level} %{host.ip} %{message}"
+      },
+      "reroute": {
+        "tag": "high_severity_logs",
+        "if" : "ctx.log?.level == 'WARN' || ctx.log?.level == 'ERROR'",
```
ruflin, line 918: @felixbarny Even though the example is correct not sure  if we should encourage to route on severity levels for logs? I rather use an example with service.name?

ruflin:
@felixbarny Would be great to get your review on this one.

docs/en/observability/logs-stream.asciidoc
felixbarny: I don't think that's a bad example to be honest. One thing I would add is to add a recommendation to keep the number of data streams relatively low. We can link to this page for more details: https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html.

Also, we should probably advocate for using the field API here

```suggestion
        "if" : "$('log.level', '') == 'WARN' || $('log.level', '') == 'ERROR'",
```

felixbarny (approved):

docs/en/observability/logs-stream.asciidoc
felixbarny: @mdbirnstiehl have you tested that this syntax works? I've just added that suggestion from how I remember the syntax.

I'd also like to see a follow up PR that links to https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html.

mdbirnstiehl: I'll verify the syntax and update accordingly. I added a note on line 893 linking to the size your shards doc

PR #3175, opened 2023-08-28
Update reroute processor config

mdbirnstiehl:


bmorelli25 (approved):
After reading the [Field API docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.9/script-fields-api.html), I'm surprised this doesn't work. Can you open an issue to update our code samples to use the field API. Probably not a high priority right now, but worth doing eventually.

PR #3179, opened 2023-08-29
[Profiling] Clarify index cleanup instructions

danielmitterdorfer:
With this commit we make sure users first disable index management so it cannot interfere with the deletion process. Then we instruct them to delete data streams first in order to avoid that the user attempts to delete underlying indices before the data streams are gone.

### DIFF
--- a/docs/en/observability/profiling-upgrade.asciidoc
+++ b/docs/en/observability/profiling-upgrade.asciidoc
@@ -54,13 +54,24 @@ NOTE: When stopping incoming requests, host-agent replicas back off and retry co
 
 You can delete existing profiling data in Kibana:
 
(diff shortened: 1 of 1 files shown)


docs/en/observability/profiling-upgrade.asciidoc
```diff
@@ -54,13 +54,24 @@ NOTE: When stopping incoming requests, host-agent replicas back off and retry co
 
 You can delete existing profiling data in Kibana:
 
+. If you are upgrading from 8.9.0 or higher, from the navigation menu, go to *Dev Tools*, paste the following snippet and execute it. If you are upgrading from an earlier version you can skip this step.
```
mdbirnstiehl, line 57:
```suggestion
. If you're upgrading from 8.9.0 or later, go to *Dev Tools* from the navigation menu, and execute the following snippet. If you're upgrading from an earlier version, skip this step.

```

mdbirnstiehl (approved):
Just one suggestion, but otherwise LGTM

PR #3183, opened 2023-08-29
Update logs docs for consistency and formatting

mdbirnstiehl:
I did a run through of the current logs docs and updated to make sure things were consistent and logical.

### DIFF
--- a/docs/en/observability/logs-stream.asciidoc
+++ b/docs/en/observability/logs-stream.asciidoc
@@ -165,7 +165,7 @@ POST logs-example-default/_doc
 }
 ----
 
-The previous command stores the document in `logs-example-default`. You can retrieve it with the following search:
+The previous command stores the document in `logs-example-default`. Retrieve it with the following search:
 
 [source,console]
 ----
@@ -210,7 +210,7 @@ GET logs-example-default/_search
 }
 ----
 
-While you can search for phrases in the `message` field, you can't use it to filter log data. Your message, however, contains all of the following potential fields you can extract and use to filter and aggregate your log data:
+While you can search for phrases in the `message` field, you can't use this field to filter log data. Your message, however, contains all of the following potential fields you can extract and use to filter and aggregate your log data:
 
 - *@timestamp* – `2023-08-08T13:45:12.123Z` – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
 - *log.level* – `WARN` – Extracting this field lets you filter logs by severity. This is helpful if you want to focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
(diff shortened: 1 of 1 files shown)


docs/en/observability/logs-stream.asciidoc
```diff
@@ -210,7 +210,7 @@ GET logs-example-default/_search
 }
 ----
 
-Your message, however, contains all of the following potential fields. Extracting these will allow you to filter and aggregate based on these fields:
+While you can search for phrases in the `message` field, you can't use it to filter log data. Your message, however, contains all of the following potential fields you can extract and use to filter and aggregate your log data:
```
bmorelli25, line 213: I'm not sure what "it" is referring to in this sentence. I can't use the search to filter log data?

docs/en/observability/logs-stream.asciidoc
```diff
@@ -235,29 +235,24 @@ When you ingested the document in the previous section, you'll notice the `@time
...
+. <<logs-stream-simulate-api>>
+. <<logs-stream-index-template>>
+. <<logs-stream-create-data-stream>>
 
 [discrete]
 [[logs-stream-ingest-pipeline]]
 === Use an ingest pipeline to extract the `@timestamp`
 
-To extract the `@timestamp` field from the example log, use an ingest pipeline with a dissect processor. Ingest pipelines in {es} are used to process incoming documents. The {ref}/dissect-processor.html[dissect processor] is one of the available processors that extracts structured fields from your unstructured log message based the pattern you set. In the following example command, the dissect processor extracts the timestamp to the `@timestamp` field.
+Ingest pipelines consist of a series of processors that perform common transformations on incoming documents before they are indexed. To extract the `@timestamp` field from the example log, you can use an ingest pipeline with a dissect processor. The {ref}/dissect-processor.html[dissect processor] extracts structured fields from unstructured log messages based on a pattern you set.
```
bmorelli25, line 251: I could be wrong about this. @dedemorton will know.

I wouldn't say:

> To extract the `@timestamp` field from the example log, **you can** use an ingest pipeline with a dissect processor.

I would say:

> To extract the `@timestamp` field from the example log, use an ingest pipeline with a dissect processor.

docs/en/observability/logs-stream.asciidoc
```diff
@@ -388,14 +382,14 @@ POST logs-example-default/_doc
...
-Now look at your document's details using this command:
+You can see your document's using this command:
 
 [source,console]
 ----
 GET /logs-example-default/_search
 ----
 
-You can see the pipeline extracted the `@timestamp` field:
+The results should look like the following showing the pipeline has extracted the `@timestamp` field:
```
bmorelli25, line 392: This sentence is tough for me

bmorelli25 (commented):
Generally looks good. I have a few suggestions. I'd love for @dedemorton to take a look if she has time.

docs/en/observability/logs-stream.asciidoc
mdbirnstiehl: I think you're right here, and this may have been a blind spot in my writing. 

At least for Microsoft:

> When you see can in your writing, consider deleting it. Can implies ability but not action. Rewrite to describe the action if possible. When ability is what you need to express, it's OK to use can to describe actions or tasks that the reader or program is able to do.

I'm putting this can in the trash can.

bmorelli25: Microsoft said it much better than I ever could have.

bmorelli25 (approved):

PR #3190, opened 2023-08-30
[DOCS] Add conditional actions highlight

lcawl:
Relates to https://github.com/elastic/kibana/pull/165137

This PR moves a "What's New" item for 8.10 into https://www.elastic.co/guide/en/observability/8.10/whats-new.html


bmorelli25 (approved):
We usually leave these out of `main` and just put them in `8.x`. But changing the base branch didn't work (it never does—I don't know why GitHub even has that option). So we can merge this into `main` and backport.

bmorelli25:
Also, thank you!

Issue #3127, opened 2023-08-10
[REQUEST]: Infrastructure Observability : Explaining 'dotted lines' in metrics charts


roshan-elastic:
### Background

The Hosts List by default uses a 15 minute duration which can cause charts to be unusable for end users:

_Example - spikes_
<img width="1108" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/e2d43ad5-21ce-47c2-aa43-5b50a1069c9c">

We are [implementing a change](https://github.com/elastic/obs-infraobs-team/issues/1134) to eradicate this in 8.10 but we need to explain to the end user how it works. To do this, we are implementing a popover directly in the UI with a link through to documentation explaining this behaviour:

_Popover in UI - 'why am I seeing dotted lines?'_
<img width="825" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/3259fe51-e1ee-4173-a71e-cd2b7170f22d">

### What do we need?

We need this documentation to exist so we can point users at it so they can learn how the new behaviour works.

> **How are you pointing a a document which doesn't exist yet?** : We are pointing to a short-link which we can change at any time. The UI currently links to https://ela.st/docs-infra-why-dotted - we can change this to direct to the new docs once it is ready

### When do we need it?

We need this for the 8.10 release.

# Detail of Issue

## What causes this behaviour?

- The 'spikey' chart is caused by the 'interval' in the lens chart being more frequent than that of the data being collected
- This causes lens to try and plot not existing values in the intervals where there is not data - resulting in an oscillation between the real values and what Lens perceives to be '0'

_Example spikey chart_
<img width="1108" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/e2d43ad5-21ce-47c2-aa43-5b50a1069c9c">

## When does it happen?

The 'interval' is automatically set by lens by default based on the duration chosen. If the user extends the duration then it will get rid of the spikey behaviour (alternatively, they can open the chart in lens and adjust the 'interval' temporarily but this isn't saved to the actual hosts view).

## How did we resolve it?

We are using the (relatively undocumented) 'missing data' feature in Lens. It joins in the lines between the missing data to make it smooth.

Additionally, we use the **use dotted line** setting to ensure that when data is filled, it is differentiated to data which has not been filled.

_Lens setting applied to all charts from 8.10 in Host List and fly-out_
[...]

PR #3213, opened 2023-09-06
Explaining 'dotted lines' in metrics charts

bmorelli25:
### Summary

The Hosts List by default uses a 15-minute duration which can cause charts to be unusable for end users. This documentation attempts to explain why users might see dashed lines in their hosts charts, and how to determine if the dashed lines are merely an interval problem, or a missing data problem.

Closes https://github.com/elastic/observability-docs/issues/3127.

@roshan-elastic, thanks for the very in-depth explanation in #3127.


bmorelli25:
I'm going to merge this to ensure we're ready for the 8.10 release. All feedback is still welcome—I'll open a new PR with suggested changes.

docs/en/observability/monitor-infra/analyze-hosts.asciidoc
```diff
@@ -145,3 +145,55 @@ From the **Actions** menu, you can choose to:
...
+You may want to investigate this time period to determine if there is an outage or issue.
+
+[role="screenshot"]
+image::images/hosts-missing-data.png[Screenshot showing missing data]
+
+[discrete]
+[[dashed-interval-and-missing]]
+==== The chart interval is too short and data is missing
+
+In the screenshot below, the data emission rate is lower than the Lens chart interval **and** there is missing data.
```
crespocarlos, line 191: > the data emission rate is lower

Would it make sense to explain that this happens because of the [`period` interval](https://www.elastic.co/guide/en/beats/metricbeat/current/configuration-metricbeat.html#metricset-period) configured in metricbeat or integration package?

PR #3231, opened 2023-09-13
Update `/welcome-to-elastic` link in AWS firehose docs

jrodewig:
**Problem:** In https://github.com/elastic/docs/pull/2752, we updated the URL prefix (`welcome-to-elastic`) and name for the "Welcome to Elastic Docs" docs. However, we still have some stray links that use the old `/welcome-to-elastic` URL prefix

**Solution:** Replace an outdated link with one that uses an attribute.


jrodewig:
No backport needed here — The AWS firehose docs only build on `main`.

bmorelli25 (approved):
Looks good! Thank you.

Issue #3224, opened 2023-09-11
Create a Filter and Aggregate log data page


mdbirnstiehl:
We have documentation on getting the most out of your log data by using ingest pipelines to extract fields and allow for filtering and aggregation. We discuss some ways to filter data in these sections, but we should have more guidance on how to filter and aggregate after the user has extracted these fields.

[Outline here](https://docs.google.com/document/d/12n9MUa2gMVBcwvszm-xH0hehtm7lL_KdkfYLkRuN4W0/edit?usp=sharing).

Issue #3222, opened 2023-09-11
Reorganize the Stream a Log documentation


mdbirnstiehl:
Currently the Stream a log page is made up of two parts, Configure a standalone agent to stream your logs to ES and Get the most out of your log data after your stream it to ES. 

The problem is that these two sections do not flow together. You start by streaming your actual log data, but the second part on "getting the most out of your log data" has to use example log data to make sure it has all of the correct fields, log.levels, etc. Because of this, the document is serving two different audiences, one that is trying to use their actual data, and one that is interested in conceptual improvements. 

There are two potential changes:

1. Separating the two documents
2. Somehow dictating the log data that the user streams in the first part of the document to match what we use in the second part.

I want to explore these two options and move forward with whichever makes the most sense.

PR #3246, opened 2023-09-26
Create Filter and Aggregate log data section

mdbirnstiehl:
This PR closes [Issue 3224](https://github.com/elastic/observability-docs/issues/3224) and [Issue 3222](https://github.com/elastic/observability-docs/issues/3222). 

### Summary
I've separated the **Get the most out of your log data** section from the Stream a log file section and renamed it **Parse and explore logs**.

I've added sections on filtering and aggregating log data to the **Parse and explore logs** documentation. This will serve as a starting point, and we can iterate and dive deeper into the topics.

Because I created a new section, the diffs are a little crazy. The filtering and aggregating data additions start on line 910 in the **Parse and explore logs** page.

This [preview](https://observability-docs_3246.docs-preview.app.elstc.co/guide/en/observability/master/logs-parse-filter.html#logs-filter-and-aggregate) goes directly to the new filter and aggregate section.


docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+          "message": "2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.",
+          "@timestamp": "2023-08-09T17:19:27.73312243Z"
+        }
+      }
+    ]
+  }
+}
+----
+
+{es} indexes the `message` field by default meaning you can search for phrases like `WARN` or `Disk usage exceeds`. For example, the following command searches for the phrase `WARN` in the log `message` field:
```
ruflin, line 65: Lets mention that the timestamp was automatically added as it was not set and is set to `now`.

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+  }
+}
+----
+
+While you can search for phrases in the `message` field, you can't use this field to filter log data. Your message, however, contains all of the following potential fields you can extract and use to filter and aggregate your log data:
+
+- *@timestamp* – `2023-08-08T13:45:12.123Z` – Extracting this field lets you sort logs by date and time. This is helpful when you want to view your logs in the order that they occurred or identify when issues happened.
+- *log.level* – `WARN` – Extracting this field lets you filter logs by severity. This is helpful if you want to focus on high-severity WARN or ERROR-level logs, and reduce noise by filtering out low-severity INFO-level logs.
+- *host.ip* – `192.168.1.101` – Extracting this field lets you filter logs by the host IP addresses. This is helpful if you want to focus on specific hosts that you’re having issues with or if you want to find disparities between hosts.
+- *message* – `Disk usage exceeds 90%.` – You can search for keywords in the message field.
```
ruflin, line 86: Nit: I stumbled over `keywords` in message.  It is kind of accurate it just confused me at first as keyword is also a type.

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+When you're ingesting a large batch of log data, a single malformed IP address can cause the entire batch to fail. Prevent this by setting `ignore_malformed` to `true` for the `host.ip` field. Update the `host.ip` field to ignore malformed IPs using the {ref}/indices-put-mapping.html[update mapping API]:
+
+[source,console]
+----
+PUT /logs-example-default/_mapping
+{
+  "properties": {
+    "host.ip": {
+      "type": "ip",
+      "ignore_malformed": true
```
ruflin, line 775: As the user is using the ECS templates, I would assume `ignore_malformed` is already set. ++ on introducing `ignore_malformed` but maybe we can just explain it with an example. @felixbarny Can you confirm it is set by default?

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+    }
+  }
+}
+----
+
+[discrete]
+[[logs-stream-reroute]]
+= Reroute log data to specific data stream
+
+preview::[]
```
ruflin, line 785: We made it GA in 8.10 so this can be removed.

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+
+[source,log]
+----
+2023-08-08T13:45:12.123Z WARN 192.168.1.101 Disk usage exceeds 90%.
+2023-08-08T13:45:14.003Z ERROR 192.168.1.103 Database connection failed.
+2023-08-08T13:45:15.004Z DEBUG 192.168.1.104 Debugging connection issue.
+2023-08-08T13:45:16.005Z INFO 192.168.1.102 User changed profile picture.
+----
+
+NOTE: When routing data to different data streams, we recommend keeping the number of data streams relatively low to avoid oversharding. See {ref}/size-your-shards.html[Size your shards] for more information.
```
ruflin, line 799: This warning might be overkill at this stage. The user just got started with routing and already has to think about shard sizes etc. Lets have this rather on the doc side for the reroute processor.

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+* <<logs-aggregate>> — Analyze and summarize data to find patterns and gain insight.
+
+[discrete]
+[[logs-filter]]
+== Filter your logs
+
+Filter your data to focus on logs that meet specific criteria, such as log levels, timestamp ranges, or host IPs. You can filter your log data in different ways:
+
+* <<logs-filter-qdsl, Query DSL>> – Filter log data from Dev Tools using Query DSL.
+* <<logs-filter-discover, Discover>> – Filter and visualize log data in {kib} using Discover.
```
ruflin, line 935: Lets point people to the new Logs Explorer here.

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+After sending log data to {es} and <<logs-stream-parse, extracting structured fields>>, filtering and aggregating your data helps you find specific information, gain insight, and monitor your systems more efficiently. 
+
+* <<logs-filter>> — Narrow down your log data by applying specific criteria.
+* <<logs-aggregate>> — Analyze and summarize data to find patterns and gain insight.
+
+[discrete]
+[[logs-filter]]
+== Filter your logs
+
+Filter your data to focus on logs that meet specific criteria, such as log levels, timestamp ranges, or host IPs. You can filter your log data in different ways:
```
ruflin, line 932: In most cases, I hope users go to Log Explorer to do the work. Can we phrase this in a way that we say: We explain you here what is happening under the hood but Logs Explorer is what we recommend?

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1227 @@
...
+
+[discrete]
+[[logs-filter-discover]]
+=== Filter logs with Discover
+
+Discover is a {kib} tool that focuses on rapid consumption of {es} data, including logs, using the {kibana-ref}/kuery-query.html[{kib} Query Language (KQL)]. Navigate to Discover from the navigation menu by clicking *Discover* under *Analytics*.
+
+[discrete]
+[[logs-filter-discover-logs-data-view]]
+==== `logs-*` data view
```
ruflin, line 1036: If we go to Logs Explorer, data views are not needed anymore as we pregenerate these and users can directly start to explore. Lets change this all to Logs Explorer.

docs/en/observability/logs-parse-filter.asciidoc
mdbirnstiehl: I believe I added this at @felixbarny request, so I'm not sure if he has a strong opinion one way or another, but I see your point and am open to removing it.

docs/en/observability/logs-parse-filter.asciidoc
mdbirnstiehl: If we want to focus on discover, maybe we don't need to show the inner workings here. We already covered some of querying and filtering with QDSL through the examples in the extracting fields sections, maybe at this point we talk specifically about logs Explorer. WDYT?

docs/en/observability/logs-parse-filter.asciidoc
mdbirnstiehl: I updated the component templates definition in the index template section to better explain this flag and removed the standalone section.

docs/en/observability/logs-parse-filter.asciidoc
ruflin: @felixbarny WDYT? I'm hoping at this stage of the users journey they don't have to think about this yet and only worry later about it.

docs/en/observability/logs-parse-filter.asciidoc
ruflin: I like that we show both in parallel. My only point is that we use Logs Explorer instead of Discover. But if you feel some examples are overkill and it can be achived much easier in Logs Explorer (what I hope is the case), happy to remove the samples. It will be interesting to see what the workflows are that you have which are not available in Logs Explorer yet

docs/en/observability/logs-parse-filter.asciidoc
```diff
@@ -0,0 +1,1072 @@
...
+{ "message": "2023-09-20T09:40:32.345Z INFO 192.168.1.106 User logout initiated." }
+{ "create": {} }
+{ "message": "2023-09-21T15:20:55.678Z DEBUG 192.168.1.102 Database connection established." }
+----
+
+Next, run this command to aggregate your log data using the `log.level` field:
+
+[source,console]
+----
+POST logs-example-default/_search?size=0&filter_path=aggregations
```
ruflin, line 967: This is similar to set a filter for the `example` dataset in the UI and use `log.level` as the field for the histogram:

Screenshot 2023-09-28 at 11 06 17

Hope this becomes our default eventually. Just wanted to share as info.

docs/en/observability/logs-parse-filter.asciidoc
felixbarny: I think there should be some mentioning of that early on. If we don't, there's a risk that users may route on a high cardinality field which then leads to frustration later on.

I don't think that users need to read the full shard sizing guide at this point. But they should know that they shouldn't overdo it.

ruflin: The part that sticks out to me is `oversharding`. We expose the user to a term (shard) which I would hope a user should never have to think about especially serverless. Maybe we can rephrase it to leave out the shard part but still link to it. Something like:

> NOTE: When routing data to different data streams, we recommend picking a field with low cardinality to prevent explosion of the number of data streams. For more details see {ref}/size-your-shards.html[Size your shards].

I don't like where the link points, much rather also in the target users would think about number of data streams not shards.

docs/en/observability/logs-parse.asciidoc
```diff
@@ -0,0 +1,881 @@
...
+
+The index template sets the following component templates:
+
+- `logs-mappings` – general mappings for log data streams that include disabling automatic date detection from `string` fields and specifying mappings for {ecs-ref}/ecs-data_stream.html[`data_stream` ECS fields].
+- `logs-settings` – general settings for log data streams including the following: 
+** The default lifecycle policy that rolls over when the primary shard reaches 50 GB or after 30 days.
+** The default pipeline that sets a `@timestamp` if there isn't one using the ingest timestamp and places a hook for the `logs@custom` pipeline. If a `logs@custom` pipeline is installed, it's applied to logs ingested into this data stream.
+** Sets the {ref}/ignore-malformed.html[`ignore_malformed`] flag to `true`. When ingesting a large batch of log data, a single malformed field like an IP address can cause the entire batch to fail. When set to true, malformed fields with a mapping type that supports this flag are still processed.
```
colleenmcginnis, line 228: This sentence was a bit confusing... 

```suggestion
- `logs-settings` – general settings for log data streams including the following: 
** The default lifecycle policy that rolls over when the primary shard reaches 50 GB or after 30 days.
** The default pipeline that, will use the ingest timestamp if there is no specified `@timestamp` and place a hook for the `logs@custom` pipeline. If a `logs@custom` pipeline is installed, it's applied to logs ingested into this data stream.
** Sets the {ref}/ignore-malformed.html[`ignore_malformed`] flag to `true`. When ingesting a large batch of log data, a single malformed field like an IP address can cause the entire batch to fail. When set to true, malformed fields with a mapping type that supports this flag are still processed.
```

docs/en/observability/logs-parse.asciidoc
```diff
@@ -0,0 +1,881 @@
...
+}
+----
+
+[discrete]
+[[logs-stream-extract-host-ip]]
+== Extract the `host.ip` field
+
+Extracting the `host.ip` field lets you filter logs by host IP addresses allowing you to focus on specific hosts that you're having issues with or find disparities between hosts. 
+
+The `host.ip` field is part of the {ecs-ref}/ecs-reference.html[Elastic Common Schema (ECS)]. Through the ECS, the `host.ip` field is mapped as an {ref}/ip.html[`ip` field type]. `ip` field types allow range queries so you can find logs with IP addresses in a specific range. You can also query `ip` field types using CIDR notation to find logs from a particular network or subnet.
```
colleenmcginnis, line 473: Do we want to assume the reader knows what "CIDR notation" is? (I didn't, but I'm not the target audience.)

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
+[[logs-filter-and-aggregate]]
+= Filter and aggregate logs
+
+Filter and aggregate your log data to find specific information, gain insight, and monitor your systems more efficiently. You can filter and aggregate based on structured fields like timestamps, log levels, and IP addresses that you've extracted from your log data. If still need to parse your log data and extract structured fields, see the <<logs-stream-parse>> documentation.
+
+See the following sections for more on filtering and aggregating your log data:
+
+* <<logs-filter>> — Narrow down your log data by applying specific criteria.
+* <<logs-aggregate>> — Analyze and summarize data to find patterns and gain insight.
```
colleenmcginnis, line 9: Maybe restructure?

```suggestion
Filter and aggregate your log data to find specific information, gain insight, and monitor your systems more efficiently. You can filter and aggregate based on structured fields like timestamps, log levels, and IP addresses that you've extracted from your log data. This guide shows you how to:

* <<logs-filter>> — Narrow down your log data by applying specific criteria.
* <<logs-aggregate>> — Analyze and summarize data to find patterns and gain insight.

If you haven't parsed your log data and extract structured fields yet, start with the <<logs-stream-parse>> documentation.
```

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
...
+
+[discrete]
+[[logs-filter-and-aggregate-prereq]]
+== Before you get started
+
+The examples on this page use the following ingest pipeline and index template, which you can set in *Dev Tools*. For more on using ingest pipelines and index templates with your log data, see the <<logs-stream-parse>> documentation.
```
colleenmcginnis, line 15:
```suggestion
The examples on this page use the following ingest pipeline and index template, which you can set in *Dev Tools*. For more on using ingest pipelines and index templates with your log data, refer to the <<logs-stream-parse>> documentation.
```

docs/en/observability/logs-filter.asciidoc
colleenmcginnis: Not sure if line 9 in the suggestion above is necessary. Is it just repeating what's in the next section?

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
...
+
+[discrete]
+[[logs-filter-logs-explorer]]
+=== Filter logs in Logs Explorer
+
+Logs Explorer is a {kib} tool that automatically provides views of your log data based on integrations and data streams. You can find Logs Explorer in the Observability menu under *Logs*. Use the {kibana-ref}/kuery-query.html[{kib} Query Language (KQL)] in the search bar to narrow down the log data displayed in Logs Explorer.
+
+For example, you might want to look into an event that occurred within a specific time range (between September 14th and 15th).
+
+First, from *Dev Tools* add some logs with varying timestamps and log levels to your data stream with the following command:
```
colleenmcginnis, line 76: It might be good to reorient the user since we're on a new page:

```suggestion
Add some logs with varying timestamps and log levels to your data stream:

. In {kib}, go to *Management* -> *Dev Tools*.
. In the *Console* tab, run the following command:
```

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
...
+----
+
+Under the *Documents* tab, you'll see the filtered log data matching your query. 
+
+[role="screenshot"]
+image::images/logs-kql-filter.png[Filter data by log level using KQL]
+
+Make sure the logs you're looking for fall within the time range in Logs Explorer. If you don't see your logs, update the time range by clicking the image:images/time-filter-icon.png[calendar icon, width=36px].
+
+For more on using Logs Explorer, see the {kibana-ref}/discover.html[Discover] documentation.
```
colleenmcginnis, line 105:
```suggestion
For more on using Logs Explorer, refer to the {kibana-ref}/discover.html[Discover] documentation.
```

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
...
+
+[discrete]
+[[logs-filter-qdsl]]
+=== Filter logs with Query DSL
+
+{ref}/query-dsl.html[Query DSL] is a JSON-based language used to send requests to {es} and retrieve data from indices and data streams. Filter your log data using Query DSL from *Dev Tools*.
+
+For example, you might want to troubleshoot an issue that happened at a specific date or time. To do this, use a boolean query with a {ref}/query-dsl-range-query.html[range query] to filter for the specific timestamp range and a {ref}/query-dsl-term-query.html[term query] to filter for `WARN` and `ERROR` log levels.
```
colleenmcginnis, line 113: Should it be "at a specific date or time", "on a specific date or time", or "on a specific date or at a specific time"? I'm stumped.

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
...
+{ "message": "2023-09-15T08:15:20.234Z WARN 192.168.1.101 Disk usage exceeds 90%." }
+{ "create": {} }
+{ "message": "2023-09-14T10:30:45.789Z ERROR 192.168.1.102 Critical system failure detected." }
+{ "create": {} }
+{ "message": "2023-09-10T14:20:45.789Z ERROR 192.168.1.105 Database connection lost." }
+{ "create": {} }
+{ "message": "2023-09-20T09:40:32.345Z INFO 192.168.1.106 User logout initiated." }
+----
+
+Let's say you want to look into an event that occurred between September 14th and 15th. The following boolean query  filters for logs with timestamps during those days with a log level of `ERROR` or `WARN`.
```
colleenmcginnis, line 130:
```suggestion
Let's say you want to look into an event that occurred between September 14th and 15th. The following boolean query filters for logs with timestamps during those days that also have a log level of `ERROR` or `WARN`.
```

docs/en/observability/logs-filter.asciidoc
```diff
@@ -0,0 +1,334 @@
...
+    "log_level_distribution": {
+      "terms": {
+        "field": "log.level"
+      }
+    }
+  }
+}
+----
+
+NOTE: Searches with an aggregation return both the query results and the aggregation, so you would see the logs matching the data and the aggregation. Setting `size` to `0` in the previous example limits the results to aggregations.
```
colleenmcginnis, line 241: Should this be a code callout for line 230 instead?

colleenmcginnis (commented):
Phew. So much great content! I left (too many?) comments and suggestions, but they are all pretty minor and in some cases subjective so definitely feel free to use or lose 'em!

From your review request:

>I attempted using callouts to explain some of the fields and would love an opinion on that

I think the code callouts work well!

colleenmcginnis (approved):
✨

PR #3253, opened 2023-10-04
Update profiling-get-started.asciidoc

yomduf:
Change the upgrade version recommendation version to 8.10.3 because of https://github.com/elastic/kibana/pull/167068


inge4pres:
Thanks for opening this @yomduf 🙏🏼 
I will close it in favor of #3254 where I also added more context for the readers

PR #3286, opened 2023-10-13
[Logs] Update filter docs

mdbirnstiehl:
Remove comma to fix code block and update with some changes from serverless update.


dedemorton (approved):
LGTM

PR #3289, opened 2023-10-13
[Logs] update reroute processor query

mdbirnstiehl:
Reroute processor query should be `GET logs-critical-default/_search`


colleenmcginnis (approved):
✨

PR #3299, opened 2023-10-19
[synthetics] Add `retestOnFailure` option

### LINKED ISSUE #3256: [Synthetics] Document retest feature
### Description

From 8.11 onwards, the default behaviour for Synthetics will be to automatically retest if the monitor (lightweight and/or browser) goes from an UP condition to DOWN.  If this retest is also DOWN, then the error will be created, and if configured, the alert sent.

This is to prevent noisy errors/alerts for transient problems.  See https://github.com/elastic/synthetics/issues/792 for more details.

- This feature required Kibana to be running 8.11.0+
- Heartbeat/Synthetics will be upgraded to support retest, on the service, before 8.11 is released
- Anyone running < 8.11.0 Kibana will still have the existing behaviour
- For private locations, this will also become the default behaviour if running Elastic Agent 8.11.0+
- Users can disable the default behaviour in the UI, or via project monitor configuration
- The UI also includes some additional info and filters to identify which tests are retests
- With this behaviour, errors are only created after the first retest (and if we get DOWN, DOWN), which would trigger the alert

It's also worth documenting somewhere, that monitors can be a maximum of 15 minutes per run, at which point they will be terminated (for any browser test, and a retest will allow 2x 15 minute runs, when the second is a retest).

### Resources

The main feature is described in https://github.com/elastic/synthetics/issues/792, with the Kibana changes done in https://github.com/elastic/kibana/issues/162399.

A new version of the Synthetics Agent is needed (planned in 1.5.0) that will include https://github.com/elastic/synthetics/pull/845, which allows this default behaviour to be toggled off for lightweight and/or browser based project monitors.

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** @shahzad31 

**Stakeholders:**
- @drewpost 
- @paulb-elastic

colleenmcginnis:
Closes #3256 

Documents the new default behavior for retesting monitors the first time the status goes from "up" to "down" and how to disable this default behavior.

- [x] Update screenshots to include the new "Enable retest on failure" option
- [x] Update the [Configure Synthetics projects](https://observability-docs_3299.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-configuration.html) page to include the `retestOnFailure` option
- [x] Update the [Analyze data from synthetic monitors](https://observability-docs_3299.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-analyze.html) page to show users how to identify which tests are retests
    * @shahzad31 I haven't been able to get a monitor to retest in my test deployment so I'm not sure what the behavior is supposed to be. I looked through the Kibana PR linked in #3256, but it's not clear to me there what the final behavior looks like.
- [x] ~~Update the [Use the CLI](https://observability-docs_3299.docs-preview.app.elstc.co/guide/en/observability/master/synthetics-command-reference.html) page to include a `retestOnFailure` flag~~ _Not available_


colleenmcginnis:
>It's also worth documenting somewhere, that monitors can be a maximum of 15 minutes per run, at which point they will be terminated (for any browser test, and a retest will allow 2x 15 minute runs, when the second is a retest).

@paulb-elastic I'm not sure exactly what this means so I'm having trouble figuring out where the information belongs. Is there a specific error that users see if a run exceeds 15 minutes? Should we put this on the [troubleshooting page](https://www.elastic.co/guide/en/observability/current/synthetics-troubleshooting.html#synthetics-troubleshooting-missing-browser-schedules)?

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -217,6 +217,13 @@ Enable monitor status alerts.
 +
 `tls.enabled` (`boolean`):::
 Enable TLS certificate alerts.
+
+`retestOnFailure` (`boolean`)::
+Enable or disable retesting when a monitor fails. Default is `true`.
++
+By default, monitors are automatically retested if the monitor goes from an "up" to "down" condition.
```
paulb-elastic, line 224: I think this is clear, but do you think it's worth reinforcing this, so it's clear retests won't happen if the monitor was already down?

e.g.
- scheduled_test (UP)
- scheduled_test (DOWN)
- retest (DOWN)
- scheduled_test (DOWN)
- scheduled_test (DOWN)
i.e. it won't continually retest for every DOWN

paulb-elastic (approved):
Looks good, thanks Colleen.  Added an inline comment for you to consider.

Not blocking, and can be a follow up but we should probably add the configuration option for lightweight monitors, which has a slightly different syntax in the YML.  Probably needs adding to synthetics-lightweight.html#synthetics-lightweight-common-options

For lightweight (YML), the syntax is `retest_on_failure: [true|false]`

Also (and unrelated to this), I notice we don't document how to set locations or private locations for lightweight project monitors (again, the syntax is slightly different as it uses snake_case, so would be `private_locations: ['Private Location 1', 'Private Location 2']` for example)

colleenmcginnis:
>Also (and unrelated to this), I notice we don't document how to set locations or private locations for lightweight project monitors (again, the syntax is slightly different as it uses snake_case, so would be private_locations: ['Private Location 1', 'Private Location 2'] for example)

I'll open a separate PR for this so I can backport to previous versions.

docs/en/observability/synthetics-configuration.asciidoc
colleenmcginnis: @elastic/obs-docs I'm curious what you think. To me, _goes from "up" to "down"_ makes it clear that it's only when "down" comes immediately after an "up", but if there is a way to make it clearer I'm open to suggestions!

dedemorton: @colleenmcginnis The ambiguous part for me is: what happens after the monitor goes "down" and the error is generated? It sounds like the monitor is tested again at the scheduled time (rather than being retesting over and over again until the monitor is back "up"). If this is the case, it wouldn't hurt to add another sentence that clarifies.

dedemorton: You could add something like: " The monitor will be tested again at the scheduled time."

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -217,6 +217,14 @@ Enable monitor status alerts.
 +
 `tls.enabled` (`boolean`):::
 Enable TLS certificate alerts.
+
+`retestOnFailure` (`boolean`)::
+Enable or disable retesting when a monitor fails. Default is `true`.
++
+By default, monitors are automatically retested if the monitor goes from "up" to "down".
+If the result of the retest is also "down", an error will be created, and if configured, an alert sent.
+Then the monitor will resume running according to the defined schedule.
```
dedemorton, line 226: LGTM!

mdbirnstiehl (approved):

PR #3300, opened 2023-10-19
Fix browser project monitor alert config

paulb-elastic:
The `alert` config was showing the YML syntax (that's already documented in https://www.elastic.co/guide/en/observability/current/synthetics-lightweight.html#synthetics-lightweight-common-options), which is not valid for browser monitors.  Have removed TLS as that's not available today for browser monitors.


docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -212,11 +212,7 @@ Control whether or not to capture screenshots. Options include `'on'`, `'off'`,
 `alert` (https://github.com/elastic/synthetics/blob/{synthetics_version}/src/dsl/monitor.ts#L47-L54[`AlertConfig`])::
 Enable or disable alerts. Read more about alerts in <<synthetics-settings-alerting>>.
 +
-`status.enabled` (`boolean`):::
-Enable monitor status alerts.
-+
-`tls.enabled` (`boolean`):::
-Enable TLS certificate alerts.
+`alert: {status: {enabled: [true|false]}}`
```
vigneshshanmugam, line 215: nit: can we also also add tls to the fields?

vigneshshanmugam (approved):
LGTM

docs/en/observability/synthetics-configuration.asciidoc
paulb-elastic: @vigneshshanmugam there isn't a TLS option for browser monitor though is there?

paulb-elastic: It is possible in lightweight monitors, yes, but we don't do TLS validation for browser monitors

colleenmcginnis, line 215: I wonder if we reformat lines 212-215 so:

1. We can remove [the link to `AlertConfig`](https://github.com/elastic/synthetics/blob/v1.5.1/src/dsl/monitor.ts#L47-L54) since it sounds like that's not the correct type (since the `tls` portion is not supported).
2. We can make it clear that this is for _monitor status_ alerts only.

```
`alert.status.enabled` (`boolean`)::
Enable or disable monitor status alerts. Read more about alerts in <<synthetics-settings-alerting>>.
```

paulb-elastic: Good idea, thanks Collen - have used that

colleenmcginnis (approved):
✨

colleenmcginnis:
👋  @paulb-elastic is this PR ready to merge or are you still waiting on feedback from someone on your team?

paulb-elastic:
Thanks @colleenmcginnis sorry for the delay

Issue #3134, opened 2023-08-14
[REQUEST]: Adding documentation for different equations that the new custom threshold rule supports


maryam-saeidi:
### Description

We have a new custom threshold rule in observability that supports a custom equation, we would like to have a page for the possible equations that can be supported here with some examples to link to the new rule as shown below:

<img src="https://github.com/elastic/observability-docs/assets/12370520/e11764cb-1a93-461a-acad-17b8267d4cd5" width=500>

This rule will not be released in v8.10, but I didn't see a template for the next version so I used `8.10 Documentation request` template.

### Resources

@simianhacker Any suggestion about what content/examples to include on this page? Is there another similar documentation page that we can get inspiration from?

This rule can be accessed via Observability > Alerts > Rules > Create a new rule: Threshold (Technical Preview)
Make sure `xpack.observability.unsafe.thresholdRule.enabled: true` is added to the Kibana yaml config file.

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** @maryam-saeidi 

**Stakeholders:**
@elastic/actionable-observability

katrin-freihofner:
I have an additional request for documenting the aggregation types (avg, ratio,...) we provide in our alerting rules. I could not find any existing docs about this.
I'm not sure if we want to handle it separately or together with this request here.

mdbirnstiehl:
@maryam-saeidi @katrin-freihofner I think it makes sense to include both of these things in the documentation for the custom threshold rule, for which I've opened a [draft PR](https://github.com/elastic/observability-docs/pull/3314). I could still use some examples of equations and use cases for custom equations.

maryam-saeidi:
@mdbirnstiehl For the equation, it would be nice to have 2 types of equation:

1. Mathematical equation
[...]

Issue #3031, opened 2023-06-22
[REQUEST]: New Observability threshold rule


lcawl:
### Description

Per https://github.com/elastic/actionable-observability/issues/36 there's a new rule that will combine three previous (logs threshold, infrastructure threshold, and metrics threshold) Observability rules, though I think all four will continue to exist in parallel for a little while.

Per @katrin-freihofner she has discussed with @mdbirnstiehl and they have added a link from the new rule to the docs (currently just a stub file), but will ultimately be similar to the existing rule-specific pages in this section: https://www.elastic.co/guide/en/observability/master/create-alerts.html

### Resources

- https://github.com/elastic/actionable-observability/issues/36
- Design [figma file](https://www.figma.com/file/WtGj2hHUPVGTpQjSs1TtTe/Rule-creation-experience?type=design&mode=design&t=PlaI1ITWRFar8FfZ-0)
- https://github.com/elastic/kibana/issues/160045

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:**

**Stakeholders:**
@katrin-freihofner

katrin-freihofner:
This rule is behind a feature flag - we can postpone the documentation for it to 8.10

mdbirnstiehl:
Due to a privilege issue, this has been pushed to 8.11.

PR #3314, opened 2023-10-27
Add custom threshold rule

mdbirnstiehl:
This PR closes Issues [3031 ](https://github.com/elastic/observability-docs/issues/3031)and [3134](https://github.com/elastic/observability-docs/issues/3134).

Preview [HERE](https://observability-docs_3314.docs-preview.app.elstc.co/guide/en/observability/master/custom-threshold-alert.html)


lcawl:
FYI In my local tests, this new rule is also affected by https://github.com/elastic/kibana/pull/162605. That is to say, it's got a "role visibility" field that you must set when you create the rule, and you can pick either "metrics" or "logs". In case it's helpful, I updated the ES query rule in the Kibana docs recently for the same new feature: https://www.elastic.co/guide/en/kibana/master/rule-type-es-query.html#_define_the_conditions_2

> You must select a scope value, which affects the [Kibana feature privileges](https://www.elastic.co/guide/en/kibana/master/kibana-privileges.html#kibana-feature-privileges) that are required to access the rule. For example when it’s set to Stack Rules, you must have the appropriate Management > Stack Rules feature privileges to view or edit the rule.

I think the same is likely true for this new rule (except "Stack rules" isn't one of the options here).

mdbirnstiehl:
> FYI In my local tests, this new rule is also affected by [elastic/kibana#162605](https://github.com/elastic/kibana/pull/162605). That is to say, it's got a "role visibility" field that you must set when you create the rule, and you can pick either "metrics" or "logs". In case it's helpful, I updated the ES query rule in the Kibana docs recently for the same new feature: https://www.elastic.co/guide/en/kibana/master/rule-type-es-query.html#_define_the_conditions_2
> 
> > You must select a scope value, which affects the [Kibana feature privileges](https://www.elastic.co/guide/en/kibana/master/kibana-privileges.html#kibana-feature-privileges) that are required to access the rule. For example when it’s set to Stack Rules, you must have the appropriate Management > Stack Rules feature privileges to view or edit the rule.
> 
> I think the same is likely true for this new rule (except "Stack rules" isn't one of the options here).

Thanks, Lisa! I'll update accordingly.

docs/en/observability/threshold-alert.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+image::images/custom-threshold-rule.png[Custom threshold alert configuration,75%]
+
+[discrete]
+[[custom-threshold-scope]]
+== Define rule data
+
+Specify the following settings to define the data the rule applies to:
+
+* *Select a data view:* Select a data view that points to the indices or data streams that you're creating a rule for. Create a new data view by clicking *Create a data view*. Refer to {kibana-ref}/data-views.html[Create a data view] for more on creating data views.
+* *Define query filter (optional):* Use a query filter to narrow down the data that the rule applies to. For example, set a query filter to a specific host name--`host.name:host-1`--to only apply the rule to that host.
```
maryam-saeidi, line 22: Is this intentional?
image

docs/en/observability/threshold-alert.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+. Set the following aggregations:
+** *Aggregation A:* Average `system.cpu.user.pct`
+** *Aggregation B:* Average `system.cpu.system.pct`
+** *Aggregation C:* Max `system.cpu.cores`.
+. Set the equation to `(A + B) / C * 100`
+. Set the threshold to `IS ABOVE 95` to alert when CPU usage is above 95%.
+*Boolean logic*:: Use conditional operators and comparison operators with you aggregations to define conditions for alerting.
++
+*Example* +
+Set an equation and threshold to trigger an alert when two fields that should have same value differ. For this example, we'll use `kubernetes.statefulset.replicas.ready` and `kubernetes.statefulset.replicas.desired`, and trigger an alert when their values differ. To do this, set the following aggregations, equation, and threshold:
```
maryam-saeidi, line 72: @framsouza is this example a good one to be used in the documentation of custom threshold rule, or do you have another example in mind? ([link](https://observability-docs_3314.docs-preview.app.elstc.co/guide/en/observability/master/custom-threshold-alert.html#custom-threshold-equation))

What is a good description of what can be achieved in this use case?

docs/en/observability/threshold-alert.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+Set one or more *group alerts by* fields for custom threshold rules to perform a composite aggregation against the selected fields.
+When any of these groups match the selected rule conditions, an alert is triggered _per group_.
+
+When you select multiple groupings, the group name is separated by commas.
+
+For example, if you group alerts by the `host.name` and `host.architecture` fields, and there are two hosts (`Host A` and `Host B`) and two architectures (`Architecture A` and `Architecture B`), the composite aggregation forms multiple groups.
+
+If the `Host A, Architecture A` group matches the rule conditions, but the `Host B, Architecture B` group doesn't, one alert is triggered for `Host A, Architecture A`.
+
+If you select one field--for example, `host.name`--and `Host A` matches the conditions but `Host B` doesn't, one alert is triggered for `Host A`.
```
maryam-saeidi, line 103: Same here
image

docs/en/observability/threshold-alert.asciidoc
```diff
@@ -0,0 +1,152 @@
...
+If both groups matches the conditions, alerts are triggered for both groups.
+
+[IMPORTANT]
+=====
+When *group alerts by* fields are selected, but no documents contain the selected field(s) within the given time range of when the alert is triggered, you can't determine the group(s). This is relevant when the rule condition is sensitive to a certain number of documents, and that number might be `0`.
```
maryam-saeidi, line 108: This part was a bit confusing, is it about the fact that missing group setting only is applicable when a group was seen previously? So when the rule is running, Host-1 should be seen at least once in order to report it as a missing group?

maryam-saeidi (commented):
Looks good already, added some comments for clarification.

docs/en/observability/threshold-alert.asciidoc
mdbirnstiehl: I actually pulled the **group by** info from another rule doc, so I have made the assumption that this is still relevant. If not, please let me know.

My understanding is that if a field that you're grouping by doesn't have data during the time range of your query, your rule won't trigger alerts based on that field. So if you have a field with no data during the time range, and you have a rule configured to alert when that field is or includes `0`, you might expect the rule to trigger an alert assuming that no data and `0` are equivalent. But that wouldn't happen in this case. 

I've attempted to rewrite it, let me know if that makes more sense or I can give it another shot.

maryam-saeidi: It is still not clear to me what we are trying to mention here, I think it is relevant as this rule is similar to the Metric threshold, so the same conditions are applied in both rules.

@simianhacker Is [this warning](https://observability-docs_3314.docs-preview.app.elstc.co/guide/en/observability/master/custom-threshold-alert.html#custom-threshold-group-by) related to what you shared that when we have a group by field, `no data` alert will not be triggered anymore? (This [comment](https://github.com/elastic/sdh-kibana/issues/4159#issuecomment-1750940737)) Or is this warning about something else?

simianhacker: Hum... I think we are trying to indicate that when there are ZERO documents we can't see ANY information about the groups. It's important to understand, `doc_count < 0` doesn't make sense when are alerting on `NO_DATA` because `NO_DATA` takes precedence. We can only alert on `NO_DATA` for a group by rule BUT ONLY after the rule execution has "seen" the instance of the group at least once.

mdbirnstiehl: Ok, I _think_ I understand, do we want to say that if you have **group alerts by** fields selected and **Alert me if groups stop reporting data** enabled, you won't be alerted if the fields in the time range have no data. You would only be alerted if there was data at one point and it stops?

docs/en/observability/threshold-alert.asciidoc
framsouza: @maryam-saeidi one example that i can think of is related to kuberetes statefulset, something like:

This will trigger if a stetefulset pods running is different from the stateful pods desired.

**Aggregation A**: SUM, `kubernetes.statefulset.ready `
**Aggregation B**:  SUM,` kubernetes.statefulset.desired`
**Equation**:  `A == B ? 1 : 0 `
**Threshold**: `IS BELOW 0 `

Here you can group by the name of cluster and statefulset deployment if you wish.

docs/en/observability/threshold-alert.asciidoc
maryam-saeidi: I just noticed this group by information was copied from [log threshold](https://www.elastic.co/guide/en/observability/7.17/logs-threshold-alert.html#group-by) rule, not metric threshold, so it is not as relevant as I initially thought.

I did some tests, and I am leaning toward removing this warning. This is more applicable when there is a filter involved, and it points to the difference between 0 and `No data.`

Here are some tests that I did:
1. When we filter data for a field that does not match any document, we get alerts for all the groups: (`container-name1` does not match any document)

|Alerts|Rule definition|
|---|---|
|image|image|

2. When I stopped sending data, I got no data for the same alert group:



3. When I selected a group by a field that does not exist in the documents (`host.type`), no alerts were generated, but since we have the preview, it was visible there that no data is available.



In conclusion, let's remove this warning and add it later if we see there is confusion about it.

mdbirnstiehl: Thanks for doing that testing. Makes sense to me to remove it.

docs/en/observability/threshold-alert.asciidoc
maryam-saeidi: Last point, how about changing
> Set an equation and threshold to trigger an alert when two fields that should have same value differ.

to

> Set an equation and threshold to trigger an alert when the number of stateful pods differs from the number of desired ones.

maryam-saeidi (approved):
Great document and thanks for taking care of comments :)

docs/en/observability/threshold-alert.asciidoc
mdbirnstiehl: Done, thanks for all of your help on this one!

colleenmcginnis (commented):
Just a few little suggestions below!

docs/en/observability/threshold-alert.asciidoc
```diff
@@ -0,0 +1,146 @@
...
+[[custom-threshold-equation]]
+=== Set the equation and threshold
+
+Set an equation using your aggregations. Based on the results of your equation, set a threshold to define when to trigger an alert. The equations use basic math or boolean logic. Refer to the following examples for possible use cases.
+
+[horizontal]
+*Math equation*:: Add, subtract, multiply, or divide your aggregations to define conditions for alerting.
++
+*Example* +
+Set an equation and threshold to to trigger an alert when a metric is above a threshold. For this example, we'll use average CPU usage—the percentage of CPU time spent in states other than `idle` or `IOWait` normalized by the number of CPU cores—and trigger an alert when CPU usage is above a specific percentage. To do this, set the following aggregations, equation, and threshold:
```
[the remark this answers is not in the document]
mdbirnstiehl: 🩰

colleenmcginnis (approved):
✨

PR #3321, opened 2023-10-31
New issue templates and automation

bmorelli25:
### Summary

This PR replaces our issue templates with just two options: community or internal. The community issue template is largely unchanged. The internal template is similar but different. Notable changes:

* Updated descriptions
* Adds a requirement to specify the impacted docset
* Adds a requirement to specify a release that the change is tied to
* Changes the collaboration model to a dropdown

In addition to these changes, I've added automation that looks at the selected docset, release, and collaboration model and applies labels to the issue based on the user's selections. Labels are mapped to selections in the new `advanced-issue-labeler.yml` file. Here are the potential labels by dropdown:

* **`version`**: `Request`, `Request:8.11`, `Request:8.12`, etc.
* **`doc-set`**: `Docset:All`, `Docset:Serverless`, `Docset:Stateful`, `Docset:Integrations`, `Docset:Unknown`
* **`collaboration`**: `Lead:Writer`, `Lead:Product`, `Lead:Engineering`, `Lead:Unknown`

Here's what the new Internal template looks like:

Screenshot 2023-10-30 at 8 24 08 PM


mdbirnstiehl (approved):
LGTM! I think requesting more specific detail with specific fields will be helpful, not everyone will answer, but it should at least remind some to add detail that they hadn't thought of.

dedemorton:
Looks 👍 overall, but I have a couple of concerns with the section about resources:

- Have we verified that it's really OK to point to internal resources from within an issue in a public repo? I do that sometimes (even though some folks in the past have said that we shouldn't), but I always qualify it with a statement to indicate that the resources are only available to internal folks. What I'm not sure about is whether providing a link to an internal reseource might reveal something about how we do things internally that could result in a security exposure.

- I worry that "please link to as many resources as possible" might open us up to having to dig through a bunch of documents that don't really have the info we need to know. We might also get links to resources we can't access or resources that are stale.  Instead could we ask them to describe the details users need to know about the new or changed feature? And then maybe ask them to provide links to resources that provide additional details?

colleenmcginnis (dismissed):
✨

bmorelli25:
Good feedback. Let me push a change.

bmorelli25:
In general, we try to avoid linking to internal resources whenever possible, as we want all users (internal and external) to be able to contribute to a discussion. But yes, it is okay to link to internal resources when necessary.

bmorelli25:
Updated image:

Screenshot 2023-10-31 at 11 30 19 AM

dedemorton (approved):
LGTM! 🚀 ✨

bmorelli25:
Thank you!

Issue #3250, opened 2023-09-28
[REQUEST]: Infrastructure UI : Inventory : Updated Fly-out


roshan-elastic:
### Description

We will be deprecating the existing Host fly-out within Inventory and replacing it with a new consistent version (as used in Hosts):

_Current behaviour_
![inventory - existing fly-out](https://github.com/elastic/observability-docs/assets/117740680/2c04fbc2-a8f2-41a2-b41d-4535c2a6ced2)

_New behaviour_

TBC - will provide but it'll use this new fly-out instead as used within 'Hosts'

<img width="744" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/23d46aba-81be-402b-993d-bcb8b5ef4896">

### Docs which would need changing

It will be this page:

https://www.elastic.co/guide/en/observability/current/view-infrastructure-metrics.html

_Specifically, this section:_
<img width="899" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/8a9d2b65-6c17-4a5e-bb4b-4354254202b6">

### Resources

**New behaviour**
[See latest edge build](https://edge-oblt.kb.us-west2.gcp.elastic-cloud.com/app/metrics/inventory?inventoryViewId=%270%27&waffleTime=(currentTime:1695896903735,isAutoReloading:!f)&waffleFilter=(expression:%27%27,kind:kuery)&waffleOptions=(accountId:%27%27,autoBounds:!t,boundsOverride:(max:1,min:0),customMetrics:!(),customOptions:!(),groupBy:!(),legend:(palette:cool,reverseColors:!f,steps:10),metric:(type:cpu),nodeType:host,region:%27%27,sort:(by:name,direction:desc),timelineOpen:!f,view:map)) (at the time of writing, this is not showing the new version yet)

**Old behaviour**
[See demo environment](https://demo.elastic.co/app/metrics/inventory?inventoryViewId=%270%27&waffleFilter=(expression:%27%27,kind:kuery)&waffleOptions=(accountId:%27%27,autoBounds:!t,boundsOverride:(max:1,min:0),customMetrics:!(),customOptions:!(),groupBy:!(),legend:(palette:cool,reverseColors:!f,steps:10),metric:(type:cpu),nodeType:host,region:%27%27,sort:(by:name,direction:desc),timelineOpen:!f,view:map))

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** @roshan-elastic 

[...]

Issue #3210, opened 2023-09-06
[REQUEST]: Host Metrics : Rename metric name


roshan-elastic:
### Description

Hi there,

Can we rename the following metric in the docs?

https://www.elastic.co/guide/en/observability/current/host-metrics.html#key-metrics-network

### Current

<img width="885" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/83ee8749-b145-4f70-8107-969172882ccf">

### Changes

1. 
**Current** : `Disk Space Availability (%)`
**New** : `Disk Usage - Available (%)`

2. 
**Current** : `Disk Space Usage (%)`
**New** : `Disk Usage - Used (%)`

3. 
**Current** : `Disk Space Available`
**New** : `DELETE THIS METRIC` (we don't use it right now so will add it in when we do)

### Resources

Feel free to check edge for where these are used:

[Host List](https://edge-oblt.kb.us-west2.gcp.elastic-cloud.com/app/metrics/hosts?waffleFilter=(expression:%27%27,kind:kuery)&waffleTime=(currentTime:1693989886732,isAutoReloading:!f)&waffleOptions=(accountId:%27%27,autoBounds:!t,boundsOverride:(max:1,min:0),customMetrics:!(),customOptions:!(),groupBy:!(),legend:(palette:cool,reverseColors:!f,steps:10),metric:(type:cpu),nodeType:host,region:%27%27,sort:(by:name,direction:desc),source:url,timelineOpen:!f,view:map)&_a=(dateRange:(from:now-15m,to:now),filters:!(),limit:100,panelFilters:!(),query:(language:kuery,query:%27%27))&controlPanels=(cloud.provider:(explicitInput:(fieldName:cloud.provider,id:cloud.provider,title:%27Cloud%20Provider%27),grow:!f,order:1,type:optionsListControl,width:medium),host.os.name:(explicitInput:(fieldName:host.os.name,id:host.os.name,title:%27Operating%20System%27),grow:!f,order:0,type:optionsListControl,width:medium))&tableProperties=(detailsItemId:gke-edge-oblt-edge-oblt-pool-3109d673-dbmq-Ubuntu,pagination:(pageIndex:0,pageSize:10),sorting:(direction:desc,field:diskSpaceUsage)))

[...]

PR #3328, opened 2023-11-01
Update docs about inventory page to show new host details flyout

### LINKED ISSUE #3249: [REQUEST]: Infrastructure UI : Update docs for new host detail view
### Background

In 8.11, we will be replacing the existing host detail view with a faster loading and more consistent view than before - we need to update the documentation which refers to these old views:

**Old Version**

_Previous Host Detail View (via Inventory) - 8.10_
8 10 - inventory to legacy host view

_Previous Host Detail View (via Host) - 8.10_
8 9 - hosts to old detail view

**8.11 - Now**

_8.11 : Inventory to new host detail view_
8 11 - inventory to new host view

_8.11 : Hosts to new host detail view_
8 11 - host to detail view

### Docs to be updated
There appears to be one main place where the docs would be updated:

https://www.elastic.co/guide/en/observability/current/view-infrastructure-metrics.html#analyze-resource-metrics

image

### Where can I see the new functionality?

[Latest 8.11 build - Inventory - edge](https://edge-oblt.kb.us-west2.gcp.elastic-cloud.com/app/metrics/inventory)
[Latest 8.11 build - Hosts - edge](https://edge-oblt.kb.us-west2.gcp.elastic-cloud.com/app/metrics/hosts)

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** @roshan-elastic 

**Stakeholders:** @smith @neptunian

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/3249, https://github.com/elastic/observability-docs/issues/3250, and https://github.com/elastic/observability-docs/issues/3210

Note that I've rearranged some of the content to improve the flow a bit and avoid confusing users:
* Rearranged the sections into a more logical order. For example, moved the filtering section to the top because it seemed like the first thing users will want to do. Also rearranged the expandable sections for each tab so that they appear in the same order as the UI.
* Added a section title for the content about viewing host metrics to make it easier for users to scan the page.
* Changed some section titles because they weren't specific enough or were misleading.

This restructuring might make the diff kind of confusing, so you'll probably want to read through the previews here:

* https://observability-docs_3328.docs-preview.app.elstc.co/guide/en/observability/master/view-infrastructure-metrics.html 
* https://observability-docs_3328.docs-preview.app.elstc.co/guide/en/observability/master/analyze-hosts.html (updates added Nov 2)

Changes on this page are minor and are best reviewed in the diff:

https://observability-docs_3328.docs-preview.app.elstc.co/guide/en/observability/master/host-metrics.html

TODO:
- [x] Update topic about Inventory page to reflect UI changes (new host details and a few other bits)
- [x] Make some structural tweaks to improve flow.
- [x] Update host metrics page.
- [x] Resolve outstanding questions:
  - [x]  Why does clicking a host in the waffle map vs the table behave differently? In the table view, a popup with links appears (instead of the flyout opening). Is this intentional or a bug?
  - [x] Looks like new pinning capability was added to the Metadata tab. How does this work? Should we mention it in the docs?
  - [x] The section about adding custom metrics isn't working as documented. I don't see the custom metric shown in a chart or in the "group by" menu. Does this feature still work as documented? Am I doing something wrong?
- [x] Update remaining screenshots.
- [x] Update the Hosts page as required.


roshan-elastic (changes requested):
Hey @dedemorton, thanks for this - looking great!

I can see that the OSQuery and Anomaly tabs use the old screenshots but I imagine you did this because you weren't able to get screenshots in the environment you have?

Could I supply new screenshots for you to use?

> **Note** : I'm supplying with/without headers in case including the headers makes it hard for you to maintain

**Overview**

_with header_
image

_without header_
image

**Metadata**

_with header_
image

_without header_
image

**Logs**

_with header_
image

_without header_
image


**Anomalies**

_with header_
image

_without header_
image

**OSQuery**

_with header_
image

_without header_
image

dedemorton:
@roshan-elastic I've added the images (thanks again for creating those!) and updated the docs about the hosts page. I decided to include the tab descriptions by importing a shared file.

As we move towards GAing the Hosts page, I should take some time to do more cleanup on this page because I feel we might be describing some details that are obvious to users (like the filtering section). But for now, I wanted to focus on getting the existing content updated for 8.11. Let me know what you think.

roshan-elastic:
@dedemorton - as promised:

> Why does clicking a host in the waffle map vs the table behave differently? In the table view, a popup with links appears (instead of the flyout opening). Is this intentional or a bug?

This could be construed as either but either way, it's not a great experience. At the moment, very few people use the table view (and we're actually aiming to completely rebuild inventory) so we're try to avoid investing in the current version of inventory where we can avoid it.

>  Looks like new pinning capability was added to the Metadata tab. How does this work? Should we mention it in the docs?

Yeah, this is pretty cool. In short, when you 'pin' something it will stay at the top in the host view across any hosts (it is saved at the browser-level though so won't persist if you log in from another browser/laptop etc).

Feel free to mention but I feel like it's probably intuitive enough for users to see/understand so might not be worth the effort of maintaining that content.

>  The section about adding custom metrics isn't working as documented. I don't see the custom metric shown in a chart or in the "group by" menu. Does this feature still work as documented? Am I doing something wrong?

For the 'group by', I think this is might be documented incorrectly. When you add a custom metric, the 'group by' will respect the metric you have chosen (inc if it is custom) but the actual group by workflow doesn't involve selecting metrics:

image

On the seeing the metric in the chart, did you mean this functionality?

image

If so, we completely missed this functionality. This isn't available in the newer fly-out. We'll prioritise it though if we find users really need it...for now it might be best to remove the docs where it says:

"A chart is also displayed for each [custom metric](https://www.elastic.co/guide/en/observability/current/view-infrastructure-metrics.html#custom-metrics) that you have added and defined on the Inventory page."

Hopefully that's everything?

roshan-elastic (approved):
Looks good to me!

dedemorton:
@roshan-elastic Thanks! I think you are right about the pinning feature, especially since it's a convenience thing...not really necessary for everyone to use.

If you decide to add the custom metric chart to the new flyout, just open a docs issue so we know to update the docs.

docs/en/observability/monitor-infra/host-details-partial.asciidoc
```diff
@@ -0,0 +1,149 @@
...
+
+[%collapsible]
+.*Anomalies*
+====
+
+[role="screenshot"]
+image::images/anomalies-overlay.png[Anomalies]
+
+The *Anomalies* table displays a list of each single metric {anomaly-detect} job for the specific host. By default, anomaly
+jobs are sorted by time to show the most recent job.
```
mdbirnstiehl, line 102:
```suggestion
jobs are sorted by time, showing the most recent jobs first. 
```
Maybe not necessary, but I got hung up on this sentence, thinking of "time to show the most recent job" was a way to sort.

docs/en/observability/monitor-infra/host-details-partial.asciidoc
```diff
@@ -0,0 +1,149 @@
...
+The *Osquery* tab allows you to build SQL statements to query your host data.
+You can create and run live or saved queries against
+the {agent}. Osquery results are stored in {es}
+so that you can use the {stack} to search, analyze, and
+visualize your host metrics. To create saved queries and add scheduled query groups,
+refer to {kibana-ref}/osquery.html[Osquery].
+
+In the example above, we query for the top 5 memory hogs running on the host.
+Under the *Results* tab, the total virtual memory size (`total_size` renamed to
+`memory_used` to be a little more user friendly) is returned in descending order,
```
mdbirnstiehl, line 136: It looks like the new screenshot doesn't show the **Results** or **Status** tab, so the user won't see it in the example. Not sure if this would maybe be more meaningful showing those tabs, if the tabs still exist?

mdbirnstiehl (dismissed):
Just a couple of small suggestions, otherwise LGTM. I'm going to approve even though it will be dismissed if you decide to make any changes 😁

docs/en/observability/monitor-infra/host-details-partial.asciidoc
dedemorton: @mdbirnstiehl Good catch! I should have checked the new screen to make sure it had the query details. Unfortunately this screen requires a setup I don't currently have. When I enter the query shown in the screen capture and click Submit, it doesn't do anything (probably because I haven't installed the agent).

I also think it's not great that we refer to an example query that _only_ appears in the screen capture (not great for accessibility).

To get this merged today, I am going to remove the content that I can't validate and follow up with Roshan tomorrow about getting an updated screen shot. Thanks!

mdbirnstiehl (approved):

docs/en/observability/monitor-infra/host-details-partial.asciidoc
dedemorton: I'm going to leave the bit about the tabs because I'm pretty sure they still exist, but I will confirm when I see the updated screenshot.

Opened an issue to make sure we get that screenshot updated: https://github.com/elastic/observability-docs/issues/3350

PR #3333, opened 2023-11-03
fix: update synthetics agent project id CLI option

vigneshshanmugam:
+ fix https://github.com/elastic/synthetics/issues/866
+ The project ID option is used as `--id` in the synthetics agent CLI option when pushing the project based monitors.


colleenmcginnis (approved):
✨

PR #3364, opened 2023-11-13
Clarify path to anomaly detection page

dedemorton:
The existing docs are confusing because they lead users to believe there should be an Anomaly detection nav item.


mdbirnstiehl (approved):
LGTM!

PR #3382, opened 2023-11-23
Add documentation clarification

paulb-elastic:
Following on from https://github.com/elastic/observability-docs/pull/3263, there are more places where this limitation was not reflected in the documentation.  This is also true for not supporting `ports` in Project Monitors.

The docs have also been updated to clarify that monitors cannot run in headful mode, but always headless, and cannot be overridden with Playwright Options.

### DIFF
--- a/docs/en/observability/synthetics-configuration.asciidoc
+++ b/docs/en/observability/synthetics-configuration.asciidoc
@@ -79,7 +79,7 @@ For all available options, refer to the https://playwright.dev/docs/test-configu
 
 [NOTE]
 ====
-Synthetics will always run browser monitors in headless mode (it is not possible to configure them to run in headful mode, even if setting Playwright options).
+Do not attempt to run in headful mode (using `headless:false`) when running through Elastic's global managed testing infrastructure or Private Locations as this is not supported.
 ====
 
 Below are details on a few Playwright options that are particularly relevant to Elastic Synthetics including timeouts, timezones, and device emulation.
--- a/docs/en/observability/synthetics-configuration.asciidoc
+++ b/docs/en/observability/synthetics-configuration.asciidoc
@@ -77,6 +77,11 @@ Read more in <<synthetics-params-secrets>>.
 
 For all available options, refer to the https://playwright.dev/docs/test-configuration[Playwright documentation].
 
+[NOTE]
+====
+Synthetics will always run browser monitors in headless mode (it is not possible to configure them to run in headful mode, even if setting Playwright options).
+====
+
 Below are details on a few Playwright options that are particularly relevant to Elastic Synthetics including timeouts, timezones, and device emulation.
 
 [discrete]
--- a/docs/en/observability/synthetics-create-test.asciidoc
+++ b/docs/en/observability/synthetics-create-test.asciidoc
@@ -167,6 +167,11 @@ Use Playwright to simulate and validate user workflows including:
 
 Visit the https://playwright.dev/docs[Playwright documentation] for information.
 
+[NOTE]
+====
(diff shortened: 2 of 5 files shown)


paulb-elastic:

paulb-elastic:

docs/en/observability/synthetics-configuration.asciidoc
```diff
@@ -77,6 +77,11 @@ Read more in <<synthetics-params-secrets>>.
 
 For all available options, refer to the https://playwright.dev/docs/test-configuration[Playwright documentation].
 
+[NOTE]
+====
+Synthetics will always run browser monitors in headless mode (it is not possible to configure them to run in headful mode, even if setting Playwright options).
```
lucabelluccini, line 82: I would have said "even forcing `headless:false` via Playwright options, we do not support running browser monitors in headful mode".

+ possibly we might give an hint of passing a user agent like here if really needed? https://github.com/elastic/synthetics/blob/d33d899bafa686da0b547afd2145db51b411ed1b/examples/todos/synthetics.config.ts#L12

docs/en/observability/synthetics-create-test.asciidoc
```diff
@@ -167,6 +167,11 @@ Use Playwright to simulate and validate user workflows including:
 
 Visit the https://playwright.dev/docs[Playwright documentation] for information.
 
+[NOTE]
```
emilioalvap, line 170: This is a bit confusing for me, I've used headful mode for debugging on synthetics cli, shouldn't this specifically refer to Elastic hosted synthetics service/heartbeat monitors instead?

Running headful locally used to be supported. IMO, it should still be

paulb-elastic: @emilioalvap that's correct, it is still possible via the CLI, which is why I didn't add the note [there](https://www.elastic.co/guide/en/observability/current/synthetics-command-reference.html#elastic-synthetics-command)

I also specifically mentioned `...run browser monitors...` to try and distinguish this as the monitor (not running the journey locally for debugging purposes), but it sounds like this was still confusing.  I didn't want to add a link here to the CLI saying it was still possible, as I thought it would get too confusing, but do you think that would help then?

docs/en/observability/synthetics-configuration.asciidoc
paulb-elastic: @lucabelluccini I thought the confusion was originally because someone was specifically trying to run in headful mode - were they doing this because they wanted a non headless UA string then?

docs/en/observability/synthetics-create-test.asciidoc
emilioalvap: I think the wording on the CLI specific page is quite detailed, I'd replicate that here:
```
Do not attempt to run in headful mode when running through Elastic’s global managed testing infrastructure or Private Locations as this is not supported.
```

IMO, this page is more generic than the CLI one to simply state that headful mode is not supported, it might be confusing for users as well.

docs/en/observability/synthetics-configuration.asciidoc
lucabelluccini: I do not have the whole detail about it, we can check together in Slack

emilioalvap (dismissed):
LGTM

lucabelluccini (approved):

colleenmcginnis:
@elasticmachine, run elasticsearch-ci/docs

colleenmcginnis (commented):
Some questions and suggestions below.

paulb-elastic:
Thanks for the feedback @colleenmcginnis, updated as per the comments

colleenmcginnis (approved):
✨

PR #3400, opened 2023-12-04
Remove beta label for SLOs

paulb-elastic:
Relates to https://github.com/elastic/kibana/issues/170456 as part of [Moving SLOs to GA](https://github.com/elastic/actionable-observability/issues/165)

### DIFF
--- a/docs/en/observability/slo-burn-rate-alert.asciidoc
+++ b/docs/en/observability/slo-burn-rate-alert.asciidoc
@@ -5,8 +5,6 @@
 <titleabbrev>Create an SLO burn rate rule</titleabbrev>
 ++++
 
-beta::[]
-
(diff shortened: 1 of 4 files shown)


dedemorton (approved):
LGTM.

Do we also want to remove the beta labels from the serverless docs? If so, when should we apply the change? (I can take care of it when we're ready.)

mgiota (approved):
LGTM! Agree with @paulb-elastic. If SLO docs for serverless show Beta badge, then it should be removed from there as well!

paulb-elastic:
With the approvals, I am going to merge this, so it will be included in the 8.12 docs (when the feature branch is created shortly).

@dedemorton I'll keep a note to update you when we can give the go ahead for the serverless docs to be updated too, thanks.

Issue #3480, opened 2023-12-21
[Request]: Filter your Host by Services 


roshan-elastic:
### Description

We have added a improvement to allow users to quickly filter their hosts by the service running on them.

This helps users who are investigating a service (or a host that a service is running on) compare different hosts to see if the problem lies with the service or the host.

_Screenshot - see **Service Name_** drop-down
![image](https://github.com/elastic/observability-docs/assets/117740680/6b329619-d990-404e-982d-89b9abae687e)

### Notes

- Right now, we will only show the services which are instrumented with APM (and this is mentioned in the popver):

<img width="273" alt="image" src="https://github.com/elastic/observability-docs/assets/117740680/b5ecb385-d7f2-4ac2-857c-c6ed3f184fc9">

- As with the other drop-downs, the values are sorted by 'document count'. This is unlikely to mean much to the end user but it is effectively the count of events received by Elastic for the hosts that are running that service  - we may want to mention what the 'document count' means. 

> **Note** : We acknowledge this is probably confusing and [we're seeing if we can change](https://github.com/elastic/kibana/issues/173821) this to show the host count instead (in the future)

### Tag Line
Here is how we're describing it:

"Easily filter down to the hosts that your service is running on so you can more easily identify resource constraints, compare behavior across different host and identify the root causes for incidents quicker"

### Link in UI
See it running on edge:

[...]

PR #3517, opened 2024-01-05
Hosts: Add service name filter to filter list

bmorelli25:
### Summary

Closes https://github.com/elastic/observability-docs/issues/3480.

### To do

- [x] Open a PR for Serverless https://github.com/elastic/staging-serverless-observability-docs/pull/212


docs/en/observability/monitor-infra/analyze-hosts.asciidoc
```diff
@@ -62,6 +62,15 @@ hosts specified by the host limit).
 and include (or exclude) metrics for those hosts.
 ** In the **Cloud Provider** list, select one or more cloud providers to
 include (or exclude) metrics for the selected cloud providers.
```
bmorelli25, line 64: This update adds a nice explanation to to the Service name filter that explains what it does (thanks, @roshan-elastic)

> This filter is useful for comparing different hosts to determine whether a problem lies
with a service or the host that it is running on.

What can we say about Operating System and Cloud Provider? Is the use case similar, i.e. to determine if the problem lies within the service or the System/Provider?

roshan-elastic: Hey Brandon, that's a good point. I think they're fairly self-explanatory but prob worth adding (maybe even worth adding in the UI as tooltips too actually):

**Operating System**
Filter your hosts by which operating system running on them.

**Cloud Provider**
Filter by the cloud provider your host is running on.

dedemorton (approved):
LGTM

PR #3545, opened 2024-01-15
docs(apm-server): update golang version in version.asciidoc

v1v:
### What

Update the Golang version when the version has been updated in https://github.com/elastic/apm-server/blob/main/.go-version.

Notify the APM-server team when changes need to be reviewed... However, we could automate the merge since those changes already exist in the main branch.

### Why

https://github.com/elastic/apm-server/pull/12412 caused it since https://github.com/elastic/apm-server/pull/12341 moved the version.asciidoc in this repository.

### Question

This is only supported for the `main` branch, so we might need to support the same for the existing active release branches. I'd like to discuss this in a follow-up.

### DIFF
--- a/.github/workflows/bump-golang.yml
+++ b/.github/workflows/bump-golang.yml
@@ -4,7 +4,7 @@ name: bump-golang
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 20 * * 6'
+    - cron: '0 3 * * 1-6'
 
 permissions:
     contents: read
--- a/.github/workflows/bump-golang.yml
+++ b/.github/workflows/bump-golang.yml
@@ -1,0 +1,22 @@
+---
+name: bump-golang
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 20 * * 6'
+
+permissions:
+    contents: read
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/bump-golang.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: bump-golang
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 20 * * 6'
```
v1v, line 7:
```suggestion
    - cron: '0 3 * * 1-6'
```

reakaleek (approved):

bmorelli25:
Thanks, @v1v. We didn't consider this when moving the docs from `elastic/apm-server` to `elastic/observability-docs`. I appreciate you looking out!

Question: Why do we need to maintain this variable in the documentation? AFAICT, we set the `:go-version:` in the `version.asciidoc` file but never actually use it anywhere in the docs.

v1v:
Hey @bmorelli25 ,

> Question: Why do we need to maintain this variable in the documentation? 

No clue :/

> EDIT: It looks like we don't use any of these version.asciidoc variables in the docs:

If so, then we don't need this PR, let's things to maintain that's even better 🙏 

I'll close this PR now and then let you decide what to do with those variables. Thanks Brandon for your review

PR #3559, opened 2024-01-22
[apm consolidation] Add temporary redirects back to APM guide (8.11)

colleenmcginnis:
Related to https://github.com/elastic/obs-docs-projects/issues/216

From the APM team:

>With the switch to https://www.elastic.co/guide/en/observability/current/apm.html , when trying to chose a previous version, e.g. 8.11 there are no redirects and the chosen version just fails with an error for every apm subpage; only the apm overview page works and links to the previous docs. Can this be improved?

This is one approach: adding redirects from the Observability guide to the APM guide for v8.11. We'll see how it works in the preview and determine how far back we should do this. 

cc @bmorelli25


bmorelli25 (approved):
https://elastic.slack.com/archives/D03149MR14J/p1705962917054879

Issue #3578, opened 2024-01-26
[Logs] Update ES host location


mdbirnstiehl:
Users can now get the ES host from the help menu under connection details. The Stream any log and Application logs docs need to be updated to show this new, easier way to find your ES host.

PR #3579, opened 2024-01-26
[Logs] Update the location of the Elasticsearch host

mdbirnstiehl:
This PR closes Issue #3578. 
Users can now get the ES host from the help menu under connection details.


bmorelli25 (commented):
Can we remove the `es-endpoint-cluster-id.png` image? Or is that used elsewhere?

mdbirnstiehl:
> Can we remove the `es-endpoint-cluster-id.png` image? Or is that used elsewhere?

It's out of here!

bmorelli25 (approved):

PR #3356, opened 2023-11-08
add application logs pages

mdbirnstiehl:

PR #3588, opened 2024-01-31
Remove outdated log correlation docs

bmorelli25:
### Summary

In https://github.com/elastic/observability-docs/pull/3356, @mdbirnstiehl rewrote our application logs documentation. The old [page](https://www.elastic.co/guide/en/observability/current/log-correlation.html), however, still exists.

This PR removes the [old page](https://www.elastic.co/guide/en/observability/current/log-correlation.html) and points to the new [Stream Application Logs page](https://www.elastic.co/guide/en/observability/current/application-logs.html).

### Related issue

@lucabelluccini requested two enhancements to our log correlation docs in https://github.com/elastic/apm-server/issues/11243:
1. it should be more prominent that using Filebeat is a possible example
2. we should document not only the field names but also their types (keyword?) and the mandatory need for a `@timestamp` and `event.dataset` (preferred).

I believe Mike's PR addressed point one. I've updated our new content to include `@timestamp` as required (cc @mdbirnstiehl). I also believe our new docs do a much better job of linking out to the ECS docs for more ino.

Closes https://github.com/elastic/apm-server/issues/11243.

### Remaining questions

> `event.dataset` (preferred)

@lucabelluccini can you explain this a little more? @mdbirnstiehl, do you know what this refers to?

### To do

- [ ] Port to Serverless


mdbirnstiehl:
I don't really know about `event.dataset`. I'm guessing the user would need to add the field through the config somehow, but I'm not sure how exactly to do that.

lucabelluccini (approved):
LGTM!

The new pages are very rich in content and also include both Filebeat & Elastic Agent ingestion. Thanks!

mdbirnstiehl (approved):
Not sure if you're waiting on anything additional besides this approval, but LGTM!

PR #3599, opened 2024-02-08
profiling: use Universal Profiling Agent consistently

florianl:
Do not mix various terms for the agent and use Universal Profiling Agent consistently.


mdbirnstiehl (approved):
Looks GREAT to me! Sorry for the late review.

Issue #3580, opened 2024-01-26
[Request]: Document new Timeslice Metric indicator for SLOs


dedemorton:
### Description

Reported by @grabowskit The new timeslice metric indicator added in 8.12 is not documented. We need to add it to the SLO documentation here:

* https://www.elastic.co/guide/en/observability/master/slo-create.html
* https://docs.elastic.co/serverless/observability/create-an-slo

### Resources

This feature was implemented in https://github.com/elastic/kibana/pull/168539.

### Which documentation set does this change impact?

Stateful and Serverless

### Feature differences

Need to confirm this, but I believe the feature is identical in stateful and serverless.

### What release is this request related to?

8.12

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** TBD

**Stakeholders:**

dedemorton:
Reopening because this needs to be copied to serverless, too.

PR #3604, opened 2024-02-09
Add docs about timeslice SLI

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/3580 for stateful docs.

Also updates "Partition by" to "Group by" to match the 8.12 UI.

A couple of thoughts:
* I think we can improve the formatting on this page, but I didn't have time to wrestle with that during this sprint.
*  The examples would probably be clearer with screen captures, but the topic is already pretty long. If we plan to add a bunch more indicators, we might want to reconsider how this topic is structured (everything on one page) and consider breaking it down into multiple topics.

### DIFF
--- a/docs/en/observability/slo-create.asciidoc
+++ b/docs/en/observability/slo-create.asciidoc
@@ -26,6 +26,7 @@ The type of SLI to use depends on the location of your data:
 
 * <<custom-kql-sli, Custom KQL>> — create an SLI based on raw logs coming from your services.
 * <<custom-metric-sli, Custom metric>> — create an SLI to define custom equations from metric fields in your indices.
+* <<timeslice-metric-sli, Timeslice metric>> — create an SLI based on a custom equation that uses multiple aggregations.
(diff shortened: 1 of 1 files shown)


dedemorton:
TODO: Use example suggested by  Francismara Souza
  
"Here's one timeslice example i could think of. In this scenario, i'm creating an SLO for a Kubernetes Statefulset, i'm retrieving the last values of both kubernetes.statefulset.replicas.ready and kubernetes.statefulset.replicas.observed. Then, I'm using the equation A == B ? 1 : 0, where A represents the number of ready replicas and B represents the number of observed replicas. This equation returns 1 if the condition A == B is true (indicating the same number of replicas) and 0 otherwise. The condition is when this equation evaluates to a value less than 1."

image
 
Screenshot 2024-02-14 at 13.37.31.png

framsouza (approved):

bmorelli25 (approved):

PR #3640, opened 2024-02-23
Make alerting navigation more consistent with serverless docs

### LINKED ISSUE #3639: Make alerting navigation consistent with serverless
### Description

The content in the stateful docs should be similar (almost identical) in content and structure to serverless, but unfortunately they have diverged. To better align the docs and address some of the concerns expressed in https://github.com/elastic/observability-docs/issues/2605, we need to change the navigation hierarchy to look something like:

- Alerting
  - Create and manage rules*
    - Custom threshold
    - Inventory threshold
    - Log threshold
    - Metric threshold
    - Monitor status
    - SLO burn rate
    - TLS certificate
    - Uptime duration anomaly
    - View  alerts
  
*check stateful to make sure these names match the UI)

### Resources

n/a

Stateful and Serverless

### Feature differences

Some rules that are in stateful are not in serverless, so the navigation will be slightly different across the two docsets.

N/A

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @dede

**Stakeholders:**

dedemorton:
Closes #3639 

Changes are mostly structural/organizational to make the navigation easier to scan and more consistent with the serverless docs. Note that the full topic titles still appear on the page to add context for SEO, but the short title is used in the navigation for easier scanning.

More changes are coming to the content later....

@mdbirnstiehl I've changed "Log**s** threshold rule" > "Log threshold rule" to match the UI. When we refer to the rule, I think we should use text that matches the UI. If the text in the UI isn't right, we should change it in the UI then update the docs. WDYT?

## Before

image

## After

image

### DIFF
--- a/docs/en/observability/create-alerts.asciidoc
+++ b/docs/en/observability/create-alerts.asciidoc
@@ -15,8 +15,13 @@ Alerts and rules related to service level objectives (SLOs), and {observability}
 You can also manage {observability} app rules alongside rules for other apps from the {kibana-ref}/create-and-manage-rules.html[{kib} Management UI].
 
 [discrete]
+== Next steps
+
+* <<create-alerts-rules>>
+* <<view-observability-alerts>>
+
 [[create-alerts-rules]]
-== Create rules
+== Create and manage rules
 
 The first step when setting up alerts is to create a rule.
 To create and manage rules related to {observability} apps,
@@ -56,14 +61,14 @@ tie into other third-party systems. Connectors allow actions to talk to these se
 
 Learn how to create specific types of rules:
 
+* {kibana-ref}/apm-alerts.html[APM rules]
 * <<custom-threshold-alert,Custom threshold rule>>
-* <<logs-threshold-alert,Logs threshold rule>>
+* <<logs-threshold-alert,Log threshold rule>>
 * <<infrastructure-threshold-alert,Infrastructure threshold rule>>
 * <<metrics-threshold-alert,Metrics threshold rule>>
 * <<monitor-status-alert,Monitor status rule>>
 * <<tls-certificate-alert,TLS certificate rule>>
 * <<duration-anomaly-alert,Uptime duration anomaly rule>>
-* {kibana-ref}/apm-alerts.html[APM rules]
 * <<slo-burn-rate-alert,SLO burn rate rule>>
 
 [discrete]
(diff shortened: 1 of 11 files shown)


docs/en/observability/metrics-threshold-alert.asciidoc
```diff
@@ -1,5 +1,8 @@
 [[metrics-threshold-alert]]
 = Create a metrics threshold rule
+++++
+<titleabbrev>Metrics threshold</titleabbrev>
```
mdbirnstiehl, line 4: Should these be Metric instead of Metrics? Looks like that's how it is in the UI. 
image

mdbirnstiehl (approved):
One question otherwise LGTM!

docs/en/observability/metrics-threshold-alert.asciidoc
```diff
@@ -1,5 +1,8 @@
 [[metrics-threshold-alert]]
 = Create a metrics threshold rule
```
dedemorton, line 2:
```suggestion
= Create a metric threshold rule
```

docs/en/observability/metrics-threshold-alert.asciidoc
dedemorton, line 4:
```suggestion
<titleabbrev>Metric threshold</titleabbrev>
```

dedemorton: Good catch!

docs/en/observability/create-alerts.asciidoc
```diff
@@ -56,14 +61,14 @@ tie into other third-party systems. Connectors allow actions to talk to these se
 
 Learn how to create specific types of rules:
 
+* {kibana-ref}/apm-alerts.html[APM rules]
 * <<custom-threshold-alert,Custom threshold rule>>
-* <<logs-threshold-alert,Logs threshold rule>>
+* <<logs-threshold-alert,Log threshold rule>>
 * <<infrastructure-threshold-alert,Infrastructure threshold rule>>
 * <<metrics-threshold-alert,Metrics threshold rule>>
```
dedemorton, line 68:
```suggestion
* <<metrics-threshold-alert,Metric threshold rule>>
```

mdbirnstiehl (approved):

PR #3646, opened 2024-02-29
[Profiling] Update minimum X86 kernel version

athre0z:
We [bumped the minimum kernel version](https://github.com/elastic/prodfiler/pull/4764) for profiling and need to update the doc accordingly.


mdbirnstiehl (approved):
LGTM!

Issue #3530, opened 2024-01-10
Documenting rate aggregation for the new custom threshold rule


maryam-saeidi:
### Description

We are introducing **rate** aggregation to the `custom threshold` alerting rule in this [PR](https://github.com/elastic/kibana/pull/173728). (To check this aggregation, you need to create a custom threshold rule from Observability > Alerts > Create rule > Custom threshold (Beta) and select `rate` aggregation)

This logic is similar to rate aggregation in the `metric threshold` alerting rule, and based on some questions (like this [one](https://elastic.slack.com/archives/C04S4RWV2M6/p1703660821678909)) it is not clear how this aggregation works behind the scene. So, we would like to share more information through documentation and possibly include a link in the rule creation flyout when rate aggregation is selected, something like:

<img src="https://github.com/elastic/observability-docs/assets/12370520/fe64b40d-a95e-42d5-8b55-4c63088cfe8e" width=300 />

### Information that can be useful to our users
We want to make sure we explain (1) how the rate is calculated and also (2) what the number shown on the preview chart means. ([example](https://elastic.slack.com/archives/C04S4RWV2M6/p1703678319854449?thread_ts=1703660821.678909&cid=C04S4RWV2M6))

The rate is calculated as follows: ([code](https://github.com/elastic/kibana/pull/173728/files#diff-b3675c41cd8592cab279989084bf6f78b0325532f42d196921aae4841356b185R25))
```
params.second > 0.0 && params.first > 0.0 && params.second > params.first ? (params.second - params.first) / ${intervalInSeconds}
```
Where first is the max value of the selected field in the **previous** bucket, and second is the maximum value for the **current** bucket.
The number shown on the preview chart is, therefore, the difference between the second and first bucket divided by the number of seconds in the selected interval.

#### Tasks
- [x] Create a dedicated section/page for rate aggregation + example
- [x] Add rate and percentile aggregations to the list of custom threshold aggregations (related [comment](https://github.com/elastic/staging-serverless-observability-docs/pull/222#pullrequestreview-1924978998))
        <img src="https://github.com/elastic/observability-docs/assets/12370520/4d9b6043-8b25-4472-af79-f62cd590f4f7" width=500 />
- [x] Custom threshold rule has a different header than other rules

| Custom threshold | Inventory |
|---|---|
[...]

PR #3676, opened 2024-03-14
Add rate and percentile aggregations to docs for custom threshold rule

dedemorton:
Adds changes requested in https://github.com/elastic/observability-docs/issues/3530.

TODO:
- [x] Add rate and percentile aggregations.
- [x] Create a dedicated section/page for rate aggregation that:
  - Explains why you would use rate instead of average for something like a counter.
  - Shows how the rate is calculated.
  - Provides a realistic example.
Then point to the page from both rules.
- [x] Refine the new content about rate aggregation and add example (with input from development team)
- [ ] Port this change to the serverless docs, too (making sure to add the missing header).


dedemorton:
@maryam-saeidi I think I've taken this as far as I can without help from the development team. I looked through all the resources you referenced in the issue, but it's really a lot to slog through. I did add a new page called "Understanding rate aggregations" but I suspect that this should probably be refined (written?) by someone who didn't just learn what a rate aggregation was yesterday. :-) The quality of the content perhaps reflects my ignorance. At any rate, it's a starting point. I welcome your comments and guidance to make it better. Plus I would appreciate your input on a good example to provide on that page.

maryam-saeidi:
Hi @dedemorton,

Thanks for adding this page! 🙏🏻

How would you like to collaborate, shall I just write information for the sections that you provided in a Google doc and share it here or do you prefer it differently? I will also get feedback from one of our SREs who worked with this aggregation to see what he thinks would be a nice addition to this document. :)

dedemorton:
@maryam-saeidi I'm totally OK with you pushing your changes directly to my fork to make them part of the PR review. I will add you as a contributor. If you prefer to work in google docs, that's OK too.

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+|===
+
+[discrete]
+[[network-rate-example]]
+== Network rate example
+
+//I think the example needs some introduction. How about:
+
+The rule definition in this example creates an alert if the number of bytes received by a host on all network interfaces exceeds 10 bit/s:
```
dedemorton, line 75: I took this description from the docs, but it seems a bit weird to say, "number of bytes received by a host...exceeds 10 bit/s."

Would "volume of data received" be better?

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+[role="screenshot"]
```
dedemorton, line 77: I think I am not understanding this as well as I should because I don't understand why you would not need to divide the threshold by 60 like we did in the previous example.

maryam-saeidi: I think it is already divided, and because of that we have `bit/s` not `bit/min`

maryam-saeidi:
@dedemorton I had a chat with @jasonrhodes and we think it would be better to put this document under a separate category like aggregations. The current place of this document does not match the rest of the items in the same group, which are different rule type:



So, either we can add another item in the following list or a separate group like aggregations:

|Current|Suggestion|
|---|---|
||<ul><li>Create and manage rules</li><li>Aggregations <br> > rate</li><li>View alerts</li></ul>|

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+// At the very least, I think we need some kind of transition that introduces this table.
+// Maybe something like this (please feel free to suggest something better):
+
+The calculation you need to use depends on the interval that's selected:
+
+|===
+| Interval | Calculation
+
+| 1m
+| restarts from [now-1m → now] - restarts from [now-2m → now-1m] / 60
```
maryam-saeidi, line 62: I see above I used `(now-1m → now)` and here `[now-1m → now]`, maybe we can use one of the styles. Do we have such representation in other documents for an interval?

dedemorton:
>The current place of this document does not match the rest of the items in the same group 

@maryam-saeidi Oh I totally agree. I've shoehorned it into the wrong place. The problem with this content is that it's an outlier compared to how we cover other types of aggregations. We are very uneven in how we cover aggregations and don't provide comprehensive reference documentation about all the aggregations we support in alerting rules. I don't want to have Aggregations at the top level in the nav right now because that will raise expectations for users who want to learn about other aggregations and will be disappointed if we only cover rate aggregations. (Plus it's not a best practice to have a nav container that contains a single page.)

For now, I can move the new page up a level in the TOC and rename it. So we'll have something like:

- Alerting
  - Create and manage rules
  - Learn about rate aggregations*
  - View Alerts

*I want to avoid calling the new page "Rate aggregations" to keep the wording parallel in the nav and prevent folks from accidentally reading "rate" as a verb initially.

In the future, we probably need a reference section somewhere in the alerting docs that covers aggregations in more depth, but that's out-of-scope for this PR. If you think it would be valuable, please open an issue, and we'll add it to our sprint planning. Thanks!

docs/en/observability/rate-aggregations.asciidoc
dedemorton: Hmm...good question. If this convention exists somewhere else in the docs, I'm not aware of it.

jasonrhodes:
@dedemorton could we make that something like this:

* Alerting
  - Create and manage rules
  - Aggregation options
    - Rate
    - [later, we can add 99th percentile, etc]
  - View Alerts

dedemorton:
@jasonrhodes I can do that if it would help your team build out more content later, but I think users are going to be misled and potentially annoyed by the missing info (plus the extra clicks). But maybe they feel that way already?

There's been some scope creep with this work (originally sized at <1 day based on the original description in the issue, which was edited after we did our sizings). That's OK, but it means that the container topic for the new section is going to be very basic. Does that sound OK?

jasonrhodes:
Of course, whatever makes sense on the scope side of things. I also defer to your expertise on the navigation choices. My perspective is that the "Information Architecture" being communicated by having "Learn about rate aggregations" by itself in the nav would, for me, be more confusing and frustrating than having some of the aggregations missing, but I understand what the page is in part because the IA makes sense to my brain (compared to that drop down that I use when I create/edit rules, in other words). 

Either way, I'll create a new issue in this repo to represent making a few more pages for the aggregations that seem like they could benefit the most from a similar explanation, @maryam-saeidi I'll probably need help choosing which ones make the most sense.

Thanks for your help, @dedemorton!

docs/en/observability/rate-aggregations.asciidoc
jasonrhodes: What about "if the rate of bytes received"? (ties "rate" into it right from the start)

dedemorton:
I'll add a commit so you can see how it looks, and we can go from there.

docs/en/observability/aggregation-options.asciidoc
```diff
@@ -0,0 +1,42 @@
+[[aggregation-options]]
+= Aggregation options
+
+Aggregations summarize your data to make it easier to analyze.
+In some alerting rules, you can specify aggregations to gather data for the rule.
+
+The following aggregations are available in some rules:
+
+|===
+|Aggregation | Description
```
dedemorton, line 10: I think the current structure works OK as long as we provide a summary table like this. I don't know if this table is complete, though. @jasonrhodes @maryam-saeidi Can you help me flesh this table out if anything is missing?

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+// I might be a little timestamp naive, but I don't understand why we start with 0000 in this example (or whether it matters either way).
+
+Let’s assume the timestamp is a UNIX timestamp in milliseconds,
+and we started counting on Thursday, January 1, 1970 12:00:00 AM.
+In that case, the rate will be calculated as follows:
+
+`(max_value_in_current_bucket - max_value_in_previous_bucket)/interval_in_seconds`, where:
+
+* `max_value_in_current_bucket` (now-1m → now): 1
+* `max_value_in_previous_bucket` (now-2m → now-1m): 0
```
dedemorton, line 44:
```suggestion
* `max_value_in_current_bucket` [now-1m → now]: 1
* `max_value_in_previous_bucket` [now-2m → now-1m]: 0
```

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+* `interval_in_seconds`: 60
+
+The rate calculation would be: `(1 - 0) / 60 = 0.0166666666667`
+
+If you want to alert when the rate of restarts is above 1 within a 1-minute window, you would set the threshold above `0.0166666666667`.
+
+// I think it might be confusing that we are representing the calculation in two different ways here,
+// especially since the first row in the table shows the calculation from the example.
+// At the very least, I think we need some kind of transition that introduces this table.
+// Maybe something like this (please feel free to suggest something better):
```
dedemorton, line 54:
```suggestion
```

docs/en/observability/rate-aggregations.asciidoc
dedemorton: I'll use square brackets. We will be doing a lot of cleanup during the asciidoc > docsmobile migration, so I'm not too concerned about this small potential inconsistency right now.

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+|===
+
+[discrete]
+[[network-rate-example]]
+== Network rate example
+
+//I think the example needs some introduction. How about:
+
+The rule definition in this example creates an alert if the number of bytes received by a host on all network interfaces exceeds 10 bit/s:
```
dedemorton, line 75:
```suggestion
The rule definition in this example creates an alert if the rate of bytes received by a host on all network interfaces exceeds 10 bit/s:
```

docs/en/observability/aggregation-options.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+
+| Min
+| Lowest value of a numeric field.
+
+| Percentile
+| The point at which a specific percentage of observed values occur.
+You can choose the 95th or 99th percentile.
+
+| Rate
+| The rate at which a specific field changes over time. To learn about how the rate is calculated,
```
dedemorton, line 32:
```suggestion
| Percentile
| Point at which a specific percentage of observed values occur.
You can choose the 95th or 99th percentile.

| Rate
| Rate at which a specific field changes over time. To learn about how the rate is calculated,
```

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+
+* `max_value_in_current_bucket` [now-1m → now]: 1
+* `max_value_in_previous_bucket` [now-2m → now-1m]: 0
+* `interval_in_seconds`: 60
+
+The rate calculation would be: `(1 - 0) / 60 = 0.0166666666667`
+
+If you want to alert when the rate of restarts is above 1 within a 1-minute window, you would set the threshold above `0.0166666666667`.
+
+The calculation you need to use depends on the interval that's selected:
```
maryam-saeidi, line 51: Personally, I don't think the information after this line is that useful (I've added them initially more for brainstorming)

@jasonrhodes @dedemorton do you see keeping the rest of the section (from this line onward) useful or shall we remove it? (Especially, the Network rate example would be useful if we add more context, having only an image seems more confusing to me.)

docs/en/observability/aggregation-options.asciidoc
maryam-saeidi: I like the new [structure](https://observability-docs_bk_3676.docs-preview.app.elstc.co/guide/en/observability/master/aggregation-options.html) 👍🏻 

Just an idea:

|Aggregation|Current|Suggestion|
|---|---|---|
|Document count|Total number of documents in a field.|Number of documents in your dataset|

docs/en/observability/aggregation-options.asciidoc
```diff
@@ -0,0 +1,42 @@
...
+
+| Max
+| Highest value of a numeric field.
+
+| Min
+| Lowest value of a numeric field.
+
+| Percentile
+| The point at which a specific percentage of observed values occur.
+You can choose the 95th or 99th percentile.
```
jasonrhodes: Maybe something like this:

nth Percentile
Numeric value which represents the point at which n% of all values in the selected dataset are lower (choices are 95th or 99th)

jasonrhodes (commented):
Thank you @dedemorton for all of the help and back and forth on this, it's looking great imo <3

docs/en/observability/aggregation-options.asciidoc
jasonrhodes: Or "in the selected dataset"

maryam-saeidi: Agree, same for sum: 
> Total of a numeric field in ~~your~~ selected dataset.

docs/en/observability/rate-aggregations.asciidoc
jasonrhodes: I'd opt for removing things we aren't sure of and evolving the docs more as we go. I think what's above this section, and the rest of the changes in this PR, are already a huge improvement, so I don't think we lose much if we pull back a bit on some of the things we haven't had time to fully edit/tweak/etc.

dedemorton: >I don't think the information after this line is that useful

I agree. TBH, I had to read the info in the table a couple times to understand why it was there, and it seems to restate what we've already said, only in a slightly different way.

Regarding the network rate example, I agree that having a screen capture is not quite good enough, but I do think it's useful to show a realistic example. It's especially important to provide more context for accessibility. That's why I've added a long alt text description to the image, but it would be much better to properly introduce the example and provide more context. 

@maryam-saeidi What additional context should we provide here? Should we show an example document? Describe what's shown in the screen capture? What else? I'm happy to fill in missing details if you can provide them, or I can remove the screenshot for now, and we can revisit this later. WDYT?

bmorelli25 (approved):
Pending the last question for @maryam-saeidi, this lgtm!

docs/en/observability/rate-aggregations.asciidoc
```diff
@@ -0,0 +1,49 @@
+[[rate-aggregations]]
+= Rate aggregations
```
maryam-saeidi, line 2: I think aggregations should be singular (In [elasticsearch document rate aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-rate-aggregation.html), it is singular)

maryam-saeidi (dismissed):
LGTM! I've added one comment for `Rate aggregations`.

Thanks for all your effort to improve our documentation ❤️

bmorelli25 (dismissed):

docs/en/observability/rate-aggregations.asciidoc
dedemorton, line 2:
```suggestion
[[rate-aggregation]]
= Rate aggregation
```

docs/en/observability/aggregation-options.asciidoc
```diff
@@ -0,0 +1,39 @@
...
+
+| Min
+| Lowest value of a numeric field.
+
+| Percentile
+| Numeric value which represents the point at which n% of all values in the selected dataset are lower (choices are 95th or 99th).
+
+| Rate
+| Rate at which a specific field changes over time. To learn about how the rate is calculated,
+refer to <<rate-aggregations>>.
```
dedemorton, line 32:
```suggestion
refer to <<rate-aggregation>>.
```

mdbirnstiehl (approved):

Issue #3659, opened 2024-03-08
Add list of valid connectors to the alerting docs


dedemorton:
Replace screen captures in the alerting docs with a list of valid connectors that link to the connectors documentation.

dedemorton:
Closed by https://github.com/elastic/observability-docs/pull/3690. I am not going to port this change to serverless until after I've had a chance to clean up inconsistencies in those docs.

PR #3690, opened 2024-03-20
[Alerting] Add list of valid connectors to alerting rules docs

dedemorton:
Updates stateful docs for https://github.com/elastic/observability-docs/issues/3659.

This change needs to be ported to serverless, too.

TODOs:

 - [x] Add list of supported connectors.
- [x]  Verify that all connectors in the list are supported for each rule.


dedemorton:
@lcawl I need you input on a couple of things:

* What do you think of providing links to the connector docs (as I've done in this PR)? The benefit is that it shows all valid connectors for the observability alerting rules. The downside (as always) is maintenance, but since all the observability alerts seem to support the same set of connectors, it doesn't seem like a big challenge, and it helps users find the info in the Kibana docs.
* It seems like the docs have diverged a lot in how we introduce the concept of connectors (obs guide vs serverless vs Kibana). What do you prefer? I don't have time to do a lot of reconciling/consistifying right now.

lcawl:
> * What do you think of providing links to the connector docs (as I've done in this PR)? The benefit is that it shows all valid connectors for the observability alerting rules. The downside (as always) is maintenance, but since all the observability alerts seem to support the same set of connectors, it doesn't seem like a big challenge, and it helps users find the info in the Kibana docs.

I think like you said the trick will be keeping that list up-to-date. Connectors come into the platform from many directions (e.g. user-contributed or initiated by the responseOps team or by one of the solution teams), so they sometimes sneak up on me. As long as this Observability-specific list is somewhere we know to update every time a new connector is added, it shouldn't be too bad, however.

> * It seems like the docs have diverged a lot in how we introduce the concept of connectors (obs guide vs serverless vs Kibana). What do you prefer? I don't have time to do a lot of reconciling/consistifying right now.

I think it's more divergent than ever now because (at least the last time I checked) in Serverless, each solution/project has a different subset of connectors that it supports. So this will definitely require some thinking once we get our connector pages migrated into docsmobile (i.e. are there some that apply to multiple projects, will this change over time, how does this affect their placement in the navigation?). Connectors are also used by cases, so we don't want to tie connectors exclusively to our rule-related content. 

It used to be that after you'd chosen a connector, the action-related options you'd see would also vary depending on your rule type (e.g. can you do action summaries, can you specify a time filter or additional KQL filters, etc) but that disparity is disappearing as we roll out the "alerts as data" features across all the solutions. I think we still won't be able to cut and paste the same info across all rules (or if we do, we'll still need rule-specific screenshots), however, because the default message that appears in each rule (and the specific action variables it contains) differs for each rule. If you want to chat further about the short-term and long-term ways we can improve on this, I'm game.

dedemorton:
I should have clarified that my main goal here is to remove the screen capture and replace it with a list because the screen capture doesn't show anything unique about the UI. It seems like the main purpose is to identify the list of valid connectors, but that means the info isn't accessible for the visually impaired. So, I've added a list with links and removed the screenshots.

After we have a clear content strategy for shared content, we'll probably need to revisit this decision. I'm going to mark this as ready for review with the understanding that we may take a different tack after we've had a chance to put more thought into how all the Alerting docs should work together and be supported by shared info (like the connector docs).

docs/en/observability/uptime-tls-alert.asciidoc
```diff
@@ -40,14 +47,13 @@ image::images/tls-alert.png[Monitor status rule]
...
 
-You can extend your rules by connecting them to actions that use the following
+Extend your rules by connecting them to actions that use the following
 supported built-in integrations. Actions are {kib} services or integrations with
 third-party systems that run as background tasks on the {kib} server when rule conditions are met.
 
 You can configure action types on the <<configure-uptime-alert-connectors,Settings>> page.
 
-[role="screenshot"]
-image::images/alert-action-types.png[TLS certificate rule connectors]
```
bmorelli25, line 50: Can we delete this image? Or is it used elsewhere?

docs/en/observability/monitor-status-alert.asciidoc
```diff
@@ -4,6 +4,12 @@
 <titleabbrev>Monitor status</titleabbrev>
 ++++
 
+//REVIEWERS:  I haven't updated this topic because I'm not sure if this page should exist.
+//I don't see the Uptime app in the UI anymore, but maybe it's only available if users have
+//installed Heartbeat? If so, users who see this topic may get confused. Should I remove this topic?
+//Looks like the alerting settings for synthetics are covered here: https://www.elastic.co/guide/en/observability/8.13/synthetics-settings.html#synthetics-settings-alerting
+//Also looks like the UI under Settings only covers a subset of the connectors that are available for other types of observability rules. Why?
```
bmorelli25, line 11: > By default, Uptime is no longer shown if they have no Heartbeat data, because the preferred approach is to use the Synthetics Kibana app.  Synthetics provides a better experience for managing monitors than Uptime (both through the [UI](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-ui.html), and/or as [project monitors](https://www.elastic.co/guide/en/observability/current/synthetics-get-started-project.html).

bmorelli25: > All new users  have it turned off by default but it's still there to turn on. Existing users will continue to see Uptime.

docs/en/observability/uptime-tls-alert.asciidoc
dedemorton: I'm pretty sure I searched the obs project to see if it's used elsewhere, but I will double-check. The gist is that this image does not really add value beyond what users can see in the UI with their own eyes.

bmorelli25 (approved):

PR #3691, opened 2024-03-20
Clarify synthetics params / secrets docs

vigneshshanmugam:
+ Clarify and add additional details to section on synthetics security / working with sensitive values. Also document how synthetics app privileges can be used for accessing these values.


dedemorton (approved):
Looks good from an editorial perspective.

Issue #3658, opened 2024-03-08
Add list of valid action variables to the alerting docs


dedemorton:
Update sections about action variables with a list of variables that are valid for the selected rule (similar to what is done [here](https://www.elastic.co/guide/en/kibana/current/rule-type-index-threshold.html#action-variables-index-threshold)). Replace the screen capture.

dedemorton:
This issue is fixed in https://github.com/elastic/observability-docs/pull/3693.

We'll need to make similar changes to serverless, but since the format of the serverless docs is internally inconsistent and also inconsistent with stateful, I'm going to wait until I am able to reconcile the differences and add this info to all of the alerting topics in serverless.

Issue #3622, opened 2024-02-13
[Meta] Observability alerting doc improvements


dedemorton:
This meta issue tracks all the issues and tasks related to improving the Observability alerting documentation.

**The work required to resolve this issue will span multiple sprints.**

- [x] https://github.com/elastic/observability-docs/issues/2605.
- [ ] Additional changes for stateful docs:
  - [x] Update stack rules by removing links to observability rules and replacing them with a single link. Users should be able to find this link easily (shouldn't be buried low on a sub-page). Also the section task about categories that no longer exist. Fixed by https://github.com/elastic/kibana/pull/177525
  - [x] Restructure the content -- separate defining rules from viewing alerts -- similar to what we've done in serverless (see draft TOC). Be as consistent as possible with serverless to make porting the changes back and forth easier. https://github.com/elastic/observability-docs/pull/3640
  - [x] #3659 https://github.com/elastic/observability-docs/pull/3690
  - [x] #3658 https://github.com/elastic/observability-docs/pull/3693
  - [ ] https://github.com/elastic/observability-docs/issues/3735
  - [ ] https://github.com/elastic/observability-docs/issues/4040 (related to https://github.com/elastic/observability-docs/pull/3693). 
  - [ ] https://github.com/elastic/observability-docs/issues/4128
  - [x] ~Consider removing screen captures that don't add value. Resize the ones that we decide to keep because they are ridiculously large.~ Removing this because we can fix the size of images easily after we migrate to docsmobile, so there is no need to do this now
- [ ] Additional changes for serverless docs:
  - [x] Add action variables and list of valid connectors to serverless docs (after we've decided on the ideal structure for alerting topics).
  - [ ] Test and update the [docs about viewing alerts](https://www.elastic.co/guide/en/observability/8.12/view-observability-alerts.html). Looks like there are some options in the actions menu that are not documented. Validate other alerting topics as time allows. #4039
   - [x] Update stale screen captures in serverless
- [x] https://github.com/elastic/observability-docs/issues/3534 Fixed by: https://github.com/elastic/staging-serverless-observability-docs/pull/222
- [ ] #3899

## Open questions

- [x] When syncing the stateful content with serverless, do we want to align with the serverless content as closely as possible? _YAS!_
- [x] Do we like the format of the serverless content and want to follow it? _Maybe - def like the collapsible sections, but want to make sure procedures in serverless do not gloss over details. Also want to remove extraneous screen captures in favor of providing content that adds value._
[...]

PR #3693, opened 2024-03-21
[Alerting] Add list of action variables to Observability alerting docs

dedemorton:
This PR adds a list of valid action variables to the following alerting rules for stateful:

- [x] Custom threshold
- [x] Log threshold
- [x] Inventory threshold
- [x] Metric threshold
- [x] SLO burn rate

Adds changes to the stateful docs for issue https://github.com/elastic/observability-docs/issues/3658. **These updates will need to be ported to the serverless docs, too.**

In a follow-on PR, we should add examples to the variable descriptions, but that's out-of-scope for this PR. Tracked here: https://github.com/elastic/observability-docs/issues/3622

## Notes

The following rules were not updated because they appear in the Uptime app, which is not visible in Kibana by default. Should we update these rules to include action variable lists or are the rules going away?

-  Monitor status
-  TLS certificate
-  Uptime duration anomaly
 
The following rules were not updated for this PR because they appear in other guides (not the Observability Guide). They will need to be updated in follow-on PRs. There is an unverified spreadsheet [here](https://docs.google.com/spreadsheets/d/1XgpCOnVo0rcOKRVFCAwsXkAb7rugfgiutntIPC6cB48/edit?usp=sharing) that shows which variables are valid for each rule type: 

- Anomaly detection ([documented in the Machine Learning Guide](https://www.elastic.co/guide/en/machine-learning/8.13/ml-configuring-alerts.html)
- APM anomaly ([documented under APM in the Kibana Guide](https://www.elastic.co/guide/en/kibana/8.13/apm-alerts.html))
- Error count threshold ([documented under APM in the Kibana Guide](https://www.elastic.co/guide/en/kibana/8.13/apm-alerts.html))
- Failed transaction rate threshold ([documented under APM in the Kibana Guide](https://www.elastic.co/guide/en/kibana/8.13/apm-alerts.html))
- Latency threshold ([documented under APM in the Kibana Guide](https://www.elastic.co/guide/en/kibana/8.13/apm-alerts.html))

For reference, here is the full list of rules that users see when they click **Create rule** in the Alerts UI.

image


docs/en/observability/infrastructure-threshold-alert.asciidoc
```diff
@@ -4,6 +4,8 @@
 <titleabbrev>Infrastructure threshold</titleabbrev>
 ++++
 
+//TODO: Change this to inventory rule.
```
dedemorton, line 7: I will rename this rule later because I'm not sure GitHub will give us a sensible diff after the rename.

docs/en/observability/create-alerts.asciidoc
```diff
@@ -7,6 +7,10 @@ Make sure alerting is already set up in {kib}. For details, see
 {kibana-ref}/alerting-setup.html[Setup and prerequisites].
 ====
 
+//REVIEWERS: What should I do with references to the Uptime app?
+//What about the alerting rules related to Uptime?
+//Uptime no longer appears in the UI (by default, at least).
```
bmorelli25, line 12: Commented in the other PR, but we should leave these. It doesn't show up by default for new users without heartbeat data, but it is still available for existing users with heartbeat data.

bmorelli25 (dismissed):

maryam-saeidi (dismissed):
LGTM!

> Should we update these rules to include action variable lists or are the rules going away?
> - Monitor status
> - TLS certificate
> - Uptime duration anomaly

I don't think so, @paulb-elastic @dominiqueclarke @shahzad31 wdyt?

> The following rules were not updated for this PR because they appear in other guides (not the Observability Guide)

For the APM ones, it is strange that we don't have them in Observability guide 🤔 

For the Anomaly detection and Elasticsearch Query, this is understandable because we recently made them available in Observability.
@vinaychandrasekhar @jasonrhodes What do you think about where the related docs should live? I think it make sense to have one page for these rules but show it in both Observability docs and stack management rule list.

jasonrhodes:
Adding this to the alerting sync agenda so we can discuss.

dedemorton:
@maryam-saeidi Some responses so you understand what I'm thinking:

>For the APM ones, it is strange that we don't have them in Observability guide

We are planning to move these rules to the Obs Guide, just not during this sprint

>What do you think about where the related docs should live? I think it make sense to have one page for these rules but show it in both Observability docs and stack management rule list.

Just to add my 2 cents, I think we need to make all of the rules available in the Obs Guide. I have this as a bullet point in a [meta issue](https://github.com/elastic/observability-docs/issues/3622) I'm keeping of all the changes that I think need to be made over time. Would love to get other folks' input on your question, too, of course.

docs/en/observability/infrastructure-threshold-alert.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[infrastructure-threshold-alert]]
-= Create an infrastructure threshold rule
+= Create an inventory threshold rule
```
dedemorton, line 2: I'm not going to change the ID because it'll break Kibana links.

bmorelli25:
> For the APM ones, it is strange that we don't have them in Observability guide

>> We are planning to move these rules to the Obs Guide, just not during this sprint

Just to add to this — we're prioritizing this in the next sprint so this will be done very soon

bmorelli25 (approved):

PR #3698, opened 2024-03-21
Fix URL for AWS EC2

tanja-milicic:


bmorelli25 (approved):
Thanks, @tanja-milicic!

PR #3674, opened 2024-03-13
Update APM URLs

bmorelli25:
### Summary

This PR updates our APM documentation by prepending the APM documentation paths with `apm-`.

### Task list
- [x] **[Pre-req]** Remove old release note redirects https://github.com/elastic/observability-docs/pull/3688
- [x] **[Pre-req]** Update release note URLs https://github.com/elastic/apm-server/pull/12820
- [x] **[Pre-req]** Create and test redirect list **([WIP Google Sheet](https://docs.google.com/spreadsheets/d/1EXF-ddCijpvJbnG-8OiASImXhp0KeP0o4JDpVvn7UC0/edit#gid=0))**
- [x] **[This PR]** Revert [#3688](#3688) and add new redirects for release notes
- [x] **[This PR]** Update all URL paths to include `apm-`
- [ ] **[after merge]** Request 301 redirects from marketing

### URLs not updated

The following URLs had paths already prepended with `apm-` and thus do not need redirects:

```
/apm.html
/apm-agent-auth.html
/apm-breaking.html
/apm-data-security.html
/apm-data-streams.html
/apm-distributed-tracing.html
/apm-integration-upgrade-steps.html
/apm-integration-upgrade-steps-ess.html
/apm-integration-upgrade-steps.html
/apm-k8s-attacher.html
/apm-mutating-admission-webhook.html
/apm-rum.html
/apm-server-configuration.html
/apm-server-starting.html
```

### Related

For https://github.com/elastic/observability-docs/issues/3573.

### Process

Documenting my process below in case we need to make similar changes in the future. Note that this is absolutely not performant, but it somehow does work! The bash file in step two takes ~3 minutes to run.

1. I replaced all APM headings in VSC:
```txt
Find: \[\[(.+)\]\]
Replace: [[apm-$1]]
Include: docs/en/observability/apm
```

2. I found all headings in the APM documentation that were prepended with `apm-*`. For each heading, this script searches for links missing `apm-*` and automatically adds them.
```bash
grep -Rih "\[\[apm-" /Users/brandon/Documents/GitHub/observability-docs/docs/en/observability/apm > headings.txt
cat headings.txt | while read line
do
  data=$(echo $line | sed -r 's/\[\[apm-(.*)\]\]$/\1/')
  echo $data
  # Update internal links
  sed -i '' "s/\<\<$data/\<\<apm-$data/g" $(find . -type f -name "*.asciidoc" 2>/dev/null)
  # Update external links (mostly in redirects file)
[...]

PR #3713, opened 2024-04-01
Add apm prefixed redirects

colleenmcginnis:
Adds redirects from old Obs URL to new Obs URL for all APM pages updated in https://github.com/elastic/observability-docs/pull/3674.


mdbirnstiehl (approved):
🐦

PR #3723, opened 2024-04-03
Move APM UI documentation to observability-docs with history

bmorelli25:
### Summary

**[Preview this PR here](https://observability-docs_bk_3723.docs-preview.app.elstc.co/guide/en/observability/master/apm.html)**

- [x] Moves our APM UI documentation from the Kibana repo to the Observability docs repo
- [x] Maintains commit history
- [x] Restructures APM UI docs into the preexisting APM content

Screenshot 2024-04-24 at 12 56 05 PM

### Potential future improvements

- Combine API sections
- Combine troubleshooting sections
- Combine settings sections
- Move storage explorer docs into manage storage section


docs/en/observability/apm-ui/spans.asciidoc
```diff
@@ -0,0 +1,72 @@
+[[spans]]
+=== Trace sample timeline
```
mdbirnstiehl, line 2: Is this section kind of redundant with the [Trace sample timeline](https://github.com/elastic/observability-docs/pull/3723/files#diff-3e9e0e1f90bd53a1d780bad2c5c43fa0054b26f6c0bfc9b180b775c199d53ee3R129) section on the Trace samples page? There's definitely a lot more info here, not sure if this info would be to much to move into the Trace samples page. If we want to keep a separate page, should it be under **Transactions**, unless this element exists somewhere outside of the individual transactions pages.

docs/en/observability/apm-ui/index.asciidoc
```diff
@@ -0,0 +1,74 @@
+[[xpack-apm]]
+= Navigate the APM UI
```
mdbirnstiehl, line 2: This section is kind of a beast, but I'm not really sure how I would organize it differently. Seems like you have your top 3 in the Kibana UI with Services, Traces, Dependencies, but most of the headings are service-specific? I wonder if it would make sense to organize the things that are service specific under the services heading to make the initial list under Navigate the APM UI less daunting?

mdbirnstiehl (commented):
I had a few comments. The APM UI is very complicated so it's kind of difficult to organize. Not sure if everything I commented is helpful, but I am open to any brainstorming on organizing this content that might come from this.

docs/en/observability/apm-ui/index.asciidoc
colleenmcginnis: As discussed in our weekly sync, we're just going to "lift and shift" the content from the Kibana guide to here. Then we will revisit the larger information architecture of the APM section in a follow-up PR.

docs/en/observability/apm-ui/spans.asciidoc
colleenmcginnis: As discussed in our weekly sync, we're just going to "lift and shift" the content from the Kibana guide to here so I think we can refine the content in follow-up PR. (This is not the only instance of redundant content.)

docs/en/observability/apm/how-to.asciidoc
```diff
@@ -1,17 +1,47 @@
 [[apm-how-to-guides]]
-== How-to guides
+== APM UI guides
```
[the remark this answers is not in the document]
colleenmcginnis: A quick search in this repo's `docs/en/observability` directory shows 32 uses of `APM UI` and  111 uses of `APM app` (including `{apm-app}`).

@bmorelli25 should we change all the new section titles to use _APM Kibana_ or _APM app_ instead of _APM UI_ (for example, _Navigate the APM app_, _APM app guides_, and _APM app APIs_)?  For some reason I thought _APM UI_ (now _Applications UI_) was more of a serverless thing, but I could be misremembering.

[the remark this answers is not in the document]
bmorelli25: APM app is an outdated term so we shouldn't go with that.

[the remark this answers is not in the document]
bmorelli25: APM Kibana, perhaps?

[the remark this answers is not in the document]
colleenmcginnis: >APM app is an outdated term so we shouldn't go with that.
Should I find-and-replace the 100+ "APM app" references that are in the docs to avoid confusion? (Either here or in a follow-up PR.)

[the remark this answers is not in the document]
dedemorton: I'd prefer to avoid "APM Kibana" because it will be an extra cleanup step when we port to serverless where the concept of Kibana doesn't exist. We are using "APM UI" in the serverless docs currently. I don't have a strong preference really, but would prefer to avoid diverging serverless from stateful when possible.

docs/en/observability/.DS_Store
colleenmcginnis, line 1: Note: Ideally we should delete this file, but I can't get this file to show up locally (even when unhiding hidden files) and I can't delete in the GitHub UI. 🤷

bmorelli25: Might be worth giving this a quick try? https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository

mdbirnstiehl (approved):
🐦 <(LGTM!)

dedemorton (approved):
LGTM  👀

PR #3792, opened 2024-04-16
Add a detailed explanation about the different field definition fields.

Oddly:
Added a description of the different files in the `fields` directory of the integration:
- agent.yml
- base-fields.yml
- ecs.yml
- fields.yml


Oddly:
After yesterday, I realized I could tell a lot more about these files, so I added this to the page.
Please take a critical look at this addition, I've gathered this information from personal experience and sometimes inferring from comments on commits. Always open for suggestions and corrections.

bmorelli25:
run docs-build

bmorelli25:
@elastic/fleet can someone from the UI and integrations team please review this PR? This is an enhancement to the integrations developer guide.

docs/en/integrations/build-integration.asciidoc
```diff
@@ -208,20 +208,55 @@ apache
...
 │   │   │      default.yml
 │   │   └───fields
-│   │          agent.yml <1>
-│   │          base-fields.yml <2>
-│   │          ecs.yml <3>
-│   │          fields.yml <4>
+│   │          agent.yml
+│   │          base-fields.yml
+│   │          ecs.yml
+│   │          fields.yml
```
jsoriano, line 214: We should mention somewhere that these names are a convention, we don't really do any special handling depending on their names and any file name with `.yml` extension would work. One could define all their fields without any order in `myfields.yml` and it would work.

docs/en/integrations/build-integration.asciidoc
```diff
@@ -208,20 +208,55 @@ apache
...
+[source,yaml]
+----
+- external: ecs
+  name: client.ip
+- external: ecs
+  name: destination.domain
+----
+
+The `external: ecs` definition instructs the `elastic-package` binary to refer to an external ECS reference to resolve these fields. By default it looks at the ECS https://raw.githubusercontent.com/elastic/ecs/v8.5.1/generated/ecs/ecs_nested.yml[reference] file on Github, following the Git reference in the `_dev/build/build.yml` file located at the root of the package directory.
+By changing the reference to the ECS reference file in `_dev/build/build.yml` from `Git@..` to `file:///file/path/ecs_nested.yml`, it is possible to look for this file locally.
```
jsoriano, line 242: Two things to add here. We support now two ways to import the whole ECS:
* Adding `import_mappings: true` into the `ecs` section in `_dev/build/build.yml` adds a dynamic mapping with most of the ECS definitions ([this one](https://github.com/elastic/elastic-package/blob/f439b96a74c27c5adfc3e7810ad584204bfaf85d/internal/builder/_static/ecs_mappings.yaml)).
* Starting on 8.13.0, Fleet pre-installs the [`ecs@mappings`](https://github.com/elastic/elasticsearch/blob/c2a3ec42632b0339387121efdef13f52c6c66848/x-pack/plugin/core/template-resources/src/main/resources/ecs%40mappings.json) component template into all integrations. This template also includes dynamic mappings for most of ECS fields.

In both cases, definitions in the fields files in packages take precedence, so normal definitions, or definitions with `external` can be used to override the dynamic templates.

jsoriano (commented):
Thanks for improving these docs!

docs/en/integrations/build-integration.asciidoc
Oddly: Good points!
It seems there is a lot more to explain in the other sections as well, bit a how-to doesn’t seem like the place for it. Is there some reference documentation where this can be explained in more length?
For some parts I had to dig through source code to be able to grok things to make my own integration. Would be nice if other people don’t have to.
Besides that, I really love the way integrations centralize everything.

jsoriano: Yeah, sadly we have documentation spread along several places. You can find more docs at least in:
* https://github.com/elastic/elastic-package/tree/main/docs/howto
* https://github.com/elastic/integrations/tree/main/docs

+1 to centralize everything in a single place. Probably also taking into account development of inputs in Beats or other collectors.

Oddly: I'd be willing to contribute to the documentation (instead of having to write internal docs), but I'd need someone to discuss changes with in addition to a landing place for these centralized docs.

@bmorelli25 : What is your take on this?

bmorelli25: This is a topic that @nimarezainia, @kilfoyle, and I discussed recently. Based on Slack messages today, it sounds like we're all in favor of moving forward with the integrations developer guide (these docs) as our source of truth for all things developer-related.

@kilfoyle is going to be moving the new content in https://github.com/elastic/integrations/tree/main/docs to this guide. Beyond that, we're still formulating a plan.

We'd love any contribution you're willing to make to the docs. I'm happy to help organize and structure any content you want to contribute.

Is that helpful?

Oddly: Thanks for the clarification, that's definitely helpful.
So, to sum it up:
- I'll review @jsoriano above suggestions, make changes and (hopefully) get this MR merged.
- Eventually we can also put this information in the integration develop guide.
- I'll keep an eye on the integration develop docs to contribute.

Is there a semi real-time way we can use to communicate?

bmorelli25: That's mostly correct. The only clarification I'll add is:

> Eventually we can also put this information in the integration develop guide.

This PR will add this content to the integration developer guide. That's what we call the documentation set that you're adding this content too.

> Is there a semi real-time way we can use to communicate?

Hmmm. Unfortunately, not really. The only thing I can think of is the [Elastic Stack Community Slack](elasticstack.slack.com). My username in that Slack group is the same as it is on GitHub.

jsoriano: > @jsoriano, can you confirm that in practice this means that ECS fields and types don't need to be specified in the integration, because they are dynamically picked up at build time (import_mappings option) or dynamically picked up in the component template in Elastic?

Correct. But only for packages with `import_mappings: true`, or that only support versions of the stack starting on 8.13.0.

Oddly:
Final version of the docs, please do a thorough review to make sure I didn't miss anything.

docs/en/integrations/build-integration.asciidoc
```diff
@@ -208,20 +208,88 @@ apache
...
+The `build.yml` file should look like the following:
++
+[source,yaml]
+----
+dependencies:
+  ecs:
+    reference: git@v8.6.0
+----
+
+In addition to this approach, we support three more methods to import the ECS, described below.
```
agithomas, line 251: @jsoriano , would it be good to suggest a preferred method here?

jsoriano: We could provide kind of a tree of decision:
* If the package only supports versions starting on 8.13.0, it doesn't need to import or define any ECS field, they are going to be provided by the stack.
* If the package supports older versions, and uses ECS mappings, it needs to import them, using `import_mappings: true` would be enough.
* If the package needs to overwrite some ECS mapping, these fields can  be imported with `external`. For the rest of fields the package would also use any of the above methods depending on the versions of the stack supported.

When does a package need to overwrite ECS mappings with `external: ecs` fields? Some examples:
* In cases where the provided dynamic mappings are not specific enough and there are mapping issues (not known example now, but could happen in theory).
* In cases where we want to provide values for `constant_keyword`s.
* In other cases where we would want to do fine-grained optimizations, like selecting between different text or numeric types.
* ...

docs/en/integrations/build-integration.asciidoc
```diff
@@ -238,14 +238,47 @@ To avoid redefining every ECS field in every integration, we can specify an exis
...
+Starting with version 8.13.0, Fleet automatically pre-installs the `ecs@mappings` component template containing the ECS definitions into all integrations. This template is similar to the dynamic mapping method, apart from the fact that it does so on ingestion instead of including the mappings into the package.
+
+. Local ECS reference file (air-gapped setup)
+By changing the Git reference to the path of the downloaded https://raw.githubusercontent.com/elastic/ecs/v8.6.0/generated/ecs/ecs_nested.yml[ECS reference] file in `_dev/build/build.yml`, it is possible for the `elastic-package` command line tool to look for this file locally. Note that the path should be the full path.
+Doing this, our `build.yml` file looks like:
++
+----
+dependencies:
+  ecs:
+    reference: file:///home/user/integrations/packages/nginx/ecs_nested.yml
```
jsoriano, line 275: I didn't know this works :slightly_smiling_face:

docs/en/integrations/build-integration.asciidoc
```diff
@@ -208,20 +208,88 @@ apache
...
+[source,yaml]
+----
+dependencies:
+  ecs:
+    reference: git@v8.6.0
+    import_mappings: true
+----
+
+. Fleet pre-installed template (from version 8.13.0)
+Starting with version 8.13.0, Fleet automatically pre-installs the `ecs@mappings` component template containing the ECS definitions into all integrations. This template is similar to the dynamic mapping method, apart from the fact that it does so on ingestion instead of including the mappings into the package.
```
jsoriano, line 266: Maybe we should put this point first, mentioning that if the package supports only versions starting on 8.13.0, it doesn't require to use any other method, except for corner cases. More details in https://github.com/elastic/observability-docs/pull/3792/files#r1588831495

docs/en/integrations/build-integration.asciidoc
agithomas: This information is super useful for an integration developer.

jsoriano (approved):
LGTM, it may be good to follow Agi's [comment,](https://github.com/elastic/observability-docs/pull/3792/files#r1588831495) and mention that for packages supporting only versions starting on 8.13.0, it is not required to define any ECS fields, and for packages using older versions, it is recommended to use `import_mappings`, unless some corner case needs to be handled.

agithomas:
@Oddly , can you please include the `decision logic` mentioned [here](https://github.com/elastic/observability-docs/pull/3792#discussion_r1588935545) to the documentation. 


++ @lalit-satapathy

Oddly:
Hi Agithomas, will do!

docs/en/integrations/build-integration.asciidoc
Oddly: One more question about this for now; If developers go the `ecs@mapping` route, do they still need to define fields that have historically been defined in agent.yml?
My gut feeling says no, but I want confirmation before I'm going to put it down in writing.

jsoriano: > One more question about this for now; If developers go the `ecs@mapping` route, do they still need to define fields that have historically been defined in agent.yml?

It will depend on whether the field is part of ECS or not. Many fields will be, but some of them are not. For example the apache package defines `input.type` in the `agent.yml` file, but this field is not part of ECS.

jsoriano: > Another question popped up, do we need to put the following in the `manifest.yml` file of the dataset in the integration to make dynamic mappings work?

It should not be needed to explicitly enable dynamic mappings, I think that default behavior is fine for the mappings provided for ECS.

Oddly:
This has become a very exhaustive explanation of field mapping methods in integrations.
While good that this is documented, I hope the size of it is not in too stark contrast with the rest of the guide.

docs/en/integrations/build-integration.asciidoc
```diff
@@ -306,20 +374,58 @@ apache
...
++
+[source,yaml]
+----
+- external: ecs
+  name: client.ip
+- external: ecs
+  name: destination.domain
+----
+
+=== fields.yml
```
agithomas, line 415: Can we provide a "Learn more" link to  https://github.com/elastic/integrations/blob/main/docs/generic_guidelines.md#document-all-field ?

agithomas (commented):
Small suggestion:

Oddly:
There is some really good information in there!
To keep overview, I think we need to do two things:
1) We provide a general link to those generic guidelines at the beginning of the docs
2) We link to the relevant part of the guidelines in each section (fields -> guidelines/fields, dataset -> guidelines/dataset, etc).

Since there's a lot on that generic guidelines page that we can not (yet) explicitly link to a subject on this page, a link to it at the beginning is also something that's needed I think.

@bmorelli25 
Some ideas for the long term regarding this doc:
I think it would be best to gather all (reference) information about a subject in one place. That way the user has overview and the chances of outdated documentation is smaller. I like to use the [diataxis](https://diataxis.fr/) way of documenting.

Therefore, it may be better to convert this how-to guide into a reference guide containing all information about integrations/packages (into multiple pages?) and move the how-to information to the new [quick-start guide](https://github.com/elastic/observability-docs/pull/3845#issuecomment-2096718151).

The Filebeat documentation structure would be (IMHO) a good starting point to use.

docs/en/integrations/build-integration.asciidoc
bmorelli25: As of https://github.com/elastic/observability-docs/pull/3836, this content has moved to the integrations developer guide!

bmorelli25: Here's a link: https://www.elastic.co/guide/en/integrations-developer/current/general-guidelines.html#_document_all_fields

bmorelli25:
@Oddly thank you so much for your continued help with these docs.

I'm on board with that plan. The ultimate goal here is to have the integration developer guide be the only place for dev docs on this topic. I have a PR open (https://github.com/elastic/integrations/pull/9796) that will remove the md files that currently exist in elastic/integrations.

> Therefore, it may be better to convert this how-to guide into a reference guide containing all information about integrations/packages (into multiple pages?) and move the how-to information to the new https://github.com/elastic/observability-docs/pull/3845#issuecomment-2096718151.

That sounds like a good plan. Should we merge this PR and do that in a follow-up or rework this PR?

bmorelli25:
run docs-build

Oddly:
I’ll try to keep working on these docs as I build integrations myself.
Lets merge this PR and open a new one for rework.

bmorelli25 (approved):
🎉

jsoriano (approved):
Thanks!

Oddly:
@jsoriano or @bmorelli25 
Can you give another approval? Added the link to the article, it's ready to merge now.

bmorelli25:
run docs-build

bmorelli25 (approved):

PR #3805, opened 2024-04-18
github-action(release automation): reviewers and use codeowners insteads

v1v:
For some reason `gh` cannot find it

> could not request reviewer: 'elastic/obs-docs' not found

And the CODEOWNERS honour the reviewers regardless.

Hence no need to use it explicitly in the script.


bmorelli25 (approved):
👍 Thanks as always, Victor!

PR #3867, opened 2024-05-07
Fix formatting in topic about logs explorer

dedemorton:
An empty line is required or the bulleted list does not render as expected:

image


bmorelli25 (approved):

mdbirnstiehl (approved):
Thanks for the fix!

PR #3877, opened 2024-05-09
Document the lack of RUM global labels support in aggregations

carsonip:
Aggregations no longer support RUM global labels since https://github.com/elastic/apm-server/pull/10624

### DIFF
--- a/docs/en/observability/apm/data-model.asciidoc
+++ b/docs/en/observability/apm/data-model.asciidoc
@@ -357,9 +357,9 @@ You can filter and group by these dimensions (some of which are optional, for ex
 * `host.name`: The user-defined name of the host or the detected hostname of the service that served the transaction
 * `host.os.platform`: The platform name of the service that served the transaction, for example `linux`
 * `kubernetes.pod.name`: The name of the Kubernetes pod running the service that served the transaction
-* `labels`: Key-value object containing string labels set globally by the APM agents.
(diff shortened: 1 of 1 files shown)


docs/en/observability/apm/data-model.asciidoc
```diff
@@ -357,9 +357,9 @@ You can filter and group by these dimensions (some of which are optional, for ex
 * `host.name`: The user-defined name of the host or the detected hostname of the service that served the transaction
 * `host.os.platform`: The platform name of the service that served the transaction, for example `linux`
 * `kubernetes.pod.name`: The name of the Kubernetes pod running the service that served the transaction
-* `labels`: Key-value object containing string labels set globally by the APM agents.
+* `labels`: Key-value object containing string labels set globally by the APM agents except RUM agents.
```
inge4pres, line 360: How can we make more evident to the reader that these data sent by RUM agents are discarded?
Is this the right place to add such note?

dedemorton: Would it make sense to add another sentence, something like: "String labels set globally by RUM agents are discarded." (APM isn't my area, so I am just spitballing in an attempt to be helpful.)

carsonip: I gave it a go. Let's see if it looks better.

bmorelli25 (approved):

PR #3889, opened 2024-05-14
Add Backport automation for separately versioned content

bmorelli25:
**This is a WIP. Feedback is very welcome.**

### Summary

This PR adds two new automations to the observability-docs repo:

- `add-backport-labels.yml`
- `clean-backports.yml`

#### Contributor flow

This PR proposes the following contributor flow:

1. A contributor opens a PR that adds content to this repository. When opening the PR, the contributor will add one or more deployment-type labels: `serverless`, `stateful`, or  `cloud`. Note that `cloud` is not currently relevant to this repo.
2. After the PR is reviewed and merged, the `add-backport-labels.yml` automation will run. This automation uses the `label-mapping.yml` file to map the deployment-type labels to the relevant backport branches. For example, if the PR is labeled with `stateful`, the `stateful` label would be removed and replaced with the `backport-8.13` and `backport-8.12` labels.
3. Mergify automatically opens PRs to the correct branches based on the newly added `backport-*` labels.
4. On these backport PRs, a new automation, `clean-backports.yml` will run. This automation uses regex to match the base branch of the backport PR with any block-listed directories. In this PR, `backport-mapping.yml` block-lists the `serverless` dir for branches that match the following regex pattern: `[7,8].*`. This means if our original PR had both serverless and stateful content, and has a backport that is targeting 8.13, an automation will run that `rm -rf serverless` on the backport PR. This ensures serverless content is not backported to branches other than `main` even when it exists in a stateful PR.
5. The backport PR passes checks and can be merged.

#### Problems solved

- Contributors don't need to know relevant backport branches. All they need to know is if the content applies to serverless, stateful, cloud, or apm-agents.
- Contributors don't have to open separate PRs for separately-versioned content. Any content can go into any PR and will make it to the appropriate backport branches.

PR #3916, opened 2024-05-22
Delete serverless directory on backport PRs

bmorelli25:
Simplified version of https://github.com/elastic/observability-docs/pull/3889 that ~doesn't work for some reason. Introduces conflicts.~ works after @colleenmcginnis fixed it.


.github/workflows/delete-serverless-dir-on-backports.yml
```diff
@@ -0,0 +1,38 @@
+name: Delete serverless directory in backports
+
+on:
+  pull_request:
+    branches:
+      - '7.*'
+      - '8.*'
+    branches-ignore:
+      - 'main'
```
bmorelli25, line 9:
```suggestion
```

colleenmcginnis (commented):
This looks good to me. Can you add a note to [the README](https://github.com/elastic/observability-docs?tab=readme-ov-file#backporting) about this? It would be good to have a place to point contributors if they notice this behavior and have questions about this.

bmorelli25:
Per Joe's testing, this needs to be backported to any branch where it will be used.

colleenmcginnis (approved):
🐦

PR #3922, opened 2024-05-24
Revert "Add docs builder"

glitteringkatie:
Reverts elastic/observability-docs#3920

_A note from @colleenmcginnis: We're removing this workflow because it's currently publishing all images from this repo to docs.elastic.co. We only want to publish the images in `docs/en/serverless` to avoid build performance problems and general messiness. We will add this workflow back once that is possible._


colleenmcginnis (approved):
🥚

PR #3927, opened 2024-05-24
Remove builder again until we can troubleshoot

colleenmcginnis:
We're continuing to have trouble getting only a subset of docs to publish to docs.elastic.co. Removing this in an abundance of caution until we can further troubleshoot this issue.


scottybollinger (approved):

colleenmcginnis:
Stale.

PR #3940, opened 2024-05-29
Enable automatic elastic.co doc builds for internal users

bmorelli25:


.github/workflows/co-docs-builder.yml
colleenmcginnis, line 1: I think we also would want to delete `labeled` from line 19 in this file. Otherwise a new build will happen every time we add a label to a PR.

bmorelli25: Thanks!

colleenmcginnis (approved):
Looks good to me. @elastic/docs-engineering any concerns?

scottybollinger:
> Looks good to me. @elastic/docs-engineering any concerns? 

None here. I echo your concerns raised elsewhere about implementing this in other repos we can't control, but this PR is fine.

scottybollinger (approved):

colleenmcginnis:
@bmorelli25 can you add a description to this PR so we're not confused what it is later?

bmorelli25:
Done! Thank you both!

PR #3991, opened 2024-06-11
Fix AWS Firehose documentation

kaiyan-sheng:
This PR is to fix several things:
1. the destination deployment  must be on ESS AWS regions and govcloud regions
2. Remove Firehose integration is in Beta note
3. Use CloudTrail logs as an example instead of VPCflow logs because you can actually configure VPC flow logs to send to Firehose directly


alaudazzi (approved):
LGTM

Issue #3981, opened 2024-06-09
[Request]: Remove ndjson.org


WilliamDEdwards:
### What documentation page is affected

Several.

### What change would you like to see?

Remove link to `ndjson.org`.

See https://github.com/ndjson/ndjson.github.io/issues/24

PR #3997, opened 2024-06-12
fix: replace ndjson.org with ndjson spec

kruskall:
the original website expired and it's currently serving malicious content.

Closes https://github.com/elastic/observability-docs/issues/3981


bmorelli25 (approved):
wow good spot

Issue #4089, opened 2024-07-24
Add otel distro links to APM book


bmorelli25:

colleenmcginnis:
Some thoughts on integrating mentions of the Elastic distros into this guide:

### Compare data ingestion options

Somewhere in the APM docs, should we compare the APM agents vs. Elastic Distros for OpenTelemetry vs. OpenTelemetry integration? Whenever there's more than one way to do something, there are likely to be questions about which should be used in different contexts/use cases/preferences. How will users know which to choose if they haven't come in with an existing opinion (or are we assuming they will)?

### Consistently present Elastic distros as an alternative to APM agents

Throughout the observability guide, we tend to say "APM agent" to mean "the thing you use to ingest data". For example:

* Getting started content instructing users to [_install APM agents_](https://www.elastic.co/guide/en/observability/current/_step_3_install_apm_agents.html)
* [Diagrams used to explain how APM works](https://www.elastic.co/guide/en/observability/8.14/apm-getting-started-apm-server.html)

Where we use "APM agent(s)" to mean "the thing you use to ingest data" (and not specifically one of the Elastic APM agents) should we acknowledge that we mean "Elastic APM agents and/or Elastic distros"? Or is "APM agent" meant to be a broad term that encompasses both? If it's not, we'd probably want to establish a single term that encompasses both.

<img width="1245" alt="Screenshot 2024-07-24 at 3 26 48 PM" src="https://github.com/user-attachments/assets/9bb0227c-4c98-420d-91ac-c13e4190c597">

colleenmcginnis:
We're holding on this until after 9.0. cc @akhileshpok 

When we're ready, these are all the pages we'll have to evaluate (these might look a bit different post-9.0 when we're using the new docs system):

* [API keys](https://www.elastic.co/guide/en/observability/current/apm-api-key.html)
* [APM agent central configuration](https://www.elastic.co/guide/en/observability/current/apm-agent-configuration.html)
* [APM agent log sending](https://www.elastic.co/guide/en/observability/current/logs-send-application.html)
* [APM agent TLS communication](https://www.elastic.co/guide/en/observability/current/apm-agent-tls.html)
* [APM agents](https://www.elastic.co/guide/en/observability/current/apm-agents.html)
* [Built-in data filters](https://www.elastic.co/guide/en/observability/current/apm-filtering.html)
* [Common problems](https://www.elastic.co/guide/en/observability/current/apm-common-problems.html)
[...]

PR #4096, opened 2024-07-25
[proof of concept] Restructure APM content related to APM agents and OpenTelemetry?

colleenmcginnis:
⚠️  This is just to illustrate a potential way to address the thoughts in https://github.com/elastic/observability-docs/issues/4089#issuecomment-2248888103 (please read that comment for context)

@elastic/obs-docs I'm curious what you think about reorganizing the APM content related to APM agents, Elastic Distros for OpenTelemetry, and the existing OpenTelemetry integration to allow readers to compare/contrast these different options and have a place to put content that is relevant to _all_ APM agents and/or Elastic Distros in one place? One potential problem with this approach is that maybe content is too deeply nested.


colleenmcginnis:
We will revisit this later.

PR #3687, opened 2024-03-19
[DOCS] Clarify AI Assistant will spawn ML nodes

lucabelluccini:
Clarify AI Assistant will spawn ML nodes

To the doc writers:
- please change wording
- backport to all versions back to AI Assistant pages got created
- add an additional paragraph if the use of AI Assistant / Knowledge Base implies ML Nodes will no more be scaled down back to 0

PR #4109, opened 2024-08-02
Update .mergify.yml

bmorelli25:
- [x] Organizes file
- [x] Removes base=main requirement from some backport rules enabling us to backport from any branch. For example, notice how this PR didn't have all backports created: https://github.com/elastic/observability-docs/pull/3687. It will after this is merged.
- [x] Add `backport-main` label so we can backport from other branches to main when PRs are opened against version branches.


dedemorton (approved):
LGTM

Issue #4087, opened 2024-07-19
[Request]: Add gemini to list of supported connectors for Observability AI assistant


dedemorton:
### Description

We need to add the gemini connector to the list of supported connectors [here](https://www.elastic.co/guide/en/observability/master/obs-ai-assistant.html).

![image](https://github.com/user-attachments/assets/1cadbc54-4567-41d7-9fb6-45c95e61908a)

We also need to align our docs better with other solutions (like security) that provide similar functionality.  But for this issue, let's just get the connector and any config details added to our existing page.

### Resources

This feature was implemented in https://github.com/elastic/kibana/pull/188002

### Which documentation set does this change impact?

Stateful and Serverless

### Feature differences

No differences

### What release is this request related to?

8.15

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @grabowskit 

**Stakeholders:** n/a

dedemorton:
This is mostly implemented in https://github.com/elastic/observability-docs/pull/4143 (just needs to be finalized and merged).

dedemorton:
Closed by https://github.com/elastic/observability-docs/pull/4143

PR #4143, opened 2024-08-12
O11y AI Assistant -  Google Gemini support

lucabelluccini:
We document Google Gemini in the connectors https://www.elastic.co/guide/en/kibana/current/gemini-action-type.html and in our release highlights https://www.elastic.co/guide/en/observability/current/whats-new.html, but it's missing from the o11y docs.

Add this mention to any place we list the supported connectors with the O11y AI Assistant.

TODO after merging: 
- [ ] Port to serverless

### DIFF
--- a/docs/en/observability/observability-ai-assistant.asciidoc
+++ b/docs/en/observability/observability-ai-assistant.asciidoc
@@ -36,11 +36,11 @@ The AI assistant requires the following:
 
 * {stack} version 8.9 and later.
 * An https://www.elastic.co/pricing[Enterprise subscription].
-* An account with a third-party generative AI provider that supports function calling. The Observability AI Assistant supports the following providers:
-** OpenAI `gpt-4`+.
-** Azure OpenAI Service `gpt-4`(0613) or `gpt-4-32k`(0613) with API version `2023-07-01-preview` or more recent.
-** AWS Bedrock, specifically the Anthropic Claude models.
-** Google Gemini `gemini-1.5-pro-preview-0409` or `gemini-1.5-flash-001`, with defaulting to `gemini-1.5-pro-001`.
+* An account with a third-party generative AI provider that preferably supports function calling.
+If your AI provider does not support function calling, you can configure AI Assistant settings under **Stack Management** to simulate function calling, but this might affect performance.
++
+Refer to the {kibana-ref}/action-types.html[connector documentation] for your provider to learn about supported and default models.
+
 * The knowledge base requires a 4 GB {ml} node.
 
 [IMPORTANT]
--- a/docs/en/observability/observability-ai-assistant.asciidoc
+++ b/docs/en/observability/observability-ai-assistant.asciidoc
@@ -40,7 +40,7 @@ The AI assistant requires the following:
 ** OpenAI `gpt-4`+.
 ** Azure OpenAI Service `gpt-4`(0613) or `gpt-4-32k`(0613) with API version `2023-07-01-preview` or more recent.
 ** AWS Bedrock, specifically the Anthropic Claude models.
-** Google Gemini, with defaulting to `gemini-1.5-pro-001`.
+** Google Gemini `gemini-1.5-pro-preview-0409` or `gemini-1.5-flash-001`, with defaulting to `gemini-1.5-pro-001`.
 * The knowledge base requires a 4 GB {ml} node.
 
 [IMPORTANT]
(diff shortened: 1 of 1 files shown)


dedemorton:
@lucabelluccini There's one more place where we mention supported connectors. Do you want me to take over this PR and make the required changes?

lucabelluccini:
Hello @dedemorton - if you can, it would be ideal.
In particular, I would suggest to check with @grabowskit to verify if we need to mention only specific models are compatible with O11y AI Assistant.

vinaychandrasekhar:
We would like to specify supported versions within the model (i.e., the Gemini model in this case). 

Checking with @emma-raffenne and team on the min supported version etc.

emma-raffenne:
@lucabelluccini and @dedemorton 

We have tested the following versions, using simulated function calling for both:
- gemini-1.5-pro-preview-0409
- gemini-1.5-flash-001

Given we haven't tested the `gemini-1.5-pro-001` version, I think it's better not to support it until we can confirm the results of its evaluation.

cc @vinaychandrasekhar

docs/en/observability/observability-ai-assistant.asciidoc
```diff
@@ -39,6 +40,7 @@ The AI assistant requires the following:
 ** OpenAI `gpt-4`+.
 ** Azure OpenAI Service `gpt-4`(0613) or `gpt-4-32k`(0613) with API version `2023-07-01-preview` or more recent.
 ** AWS Bedrock, specifically the Anthropic Claude models.
+** Google Gemini, with defaulting to `gemini-1.5-pro-001`.
```
dedemorton, line 43:
```suggestion
** Google Gemini `gemini-1.5-pro-preview-0409` or `gemini-1.5-flash-001`, with defaulting to `gemini-1.5-pro-001`.
```

gemini-1.5-pro-preview-0409
gemini-1.5-flash-001

dedemorton:
This should be ready for final review as soon as the build passes.

lucabelluccini:
Hello! Just FYI, our own docs (linked above) sets `gemini-1.5-pro-001` as default - see [here](https://www.elastic.co/guide/en/kibana/current/gemini-action-type.html#:~:text=The%20GAI%20model%20for%20Google%20Gemini%20to%20use.%20Current%20support%20is%20for%20the%20Google%20Gemini%20models%2C%20defaulting%20to%20gemini%2D1.5%2Dpro%2D001.%20The%20model%20can%20be%20set%20on%20a%20per%20request%20basis%20by%20including%20a%20%22model%22%20parameter%20alongside%20the%20request%20body.)

dedemorton:
>Just FYI, our own docs (linked above) sets gemini-1.5-pro-001

Good point. My change contradicts itself. 😸 

@emma-raffenne Let me know what I should say.

emma-raffenne:
@dedemorton @lucabelluccini 
After reviewing this part:

Screenshot 2024-08-21 at 12 19 11

I thinks we need to make a couple of changes:
- the observability AI assistant can use GenAI services that do not support function calling. In that case, the "[Simulate function calling](https://edge-lite-oblt.kb.us-west2.gcp.elastic-cloud.com/app/management/kibana/observabilityAiAssistantManagement)" setting needs to be enabled.
- the obs AI Assistant supports OpenAI, Amazon Bedrock, and Google Gemini _connectors_. I would recommend to link to each connector page of the [Kibana Stack Management docs](https://www.elastic.co/guide/en/kibana/current/action-types.html) to rely on what's indicated there as per default model. If the default model needs to be updated, it should be at the connector page.

Additionally we could add the list of the models we have tested with but I would need to confirm the full list. Let's not include them here for now.

Would that work?

dedemorton:
@emma-raffenne Is [this change](https://github.com/elastic/observability-docs/pull/4143/commits/3c7d515ef2d115d11c93ba8baf178c9c77b810ae) what you had in mind, or do you still want me to list supported versions, but point to the connector docs for defaults? (I am using a generic link to the connector docs because we already show links elsewhere.) IMO it would be great if the connector docs were a single source of truth wrt what is supported and what the defaults are, but right now, those docs don't contain all those details. And I'm not sure if our AI Assistant has special requirements that apply only to the Obs AI Assistant. Let me know what you think.

lucabelluccini:
As the AI Connector of Platform can be used by both O11y & Security, we should try to converge what's common in the connectors docs and recommendations for O11y in the O11y AI docs, notably:
- we do not recommend BYO-LLM (Security seems to "allow" it)
- you need a non-free tier or otherwise you'll hit rate limiting immediately - this might be applicable to all connectors(?)
- function calling and switch to simulate function calling - this might be applicable to all connectors(?)
- O11y might be tested only with specific 3rd party vendors and models and maybe in the future the platform offering will diverge vs O11y or Security specific recommendations

dedemorton:
@lucabelluccini Please accept my apologies for taking so long to get this merged. I've been working on a totally different project. In the interest of getting this merged, I'm going to merge this PR now and create a follow-up issue to address the concerns you expressed in [your comment.](https://github.com/elastic/observability-docs/pull/4143#issuecomment-2326147889) (I think more discussion might be required to figure out the right way to document the recommendations.)

dedemorton:
Closes https://github.com/elastic/observability-docs/issues/4087.

PR #4147, opened 2024-08-13
Add tags to fix chunking in obs guide

dedemorton:
Clean up float tags in advance of bumpting the chunk level to 4.

required before merging https://github.com/elastic/docs/pull/3045.

### DIFF
--- a/docs/en/observability/apm-ui/api.asciidoc
+++ b/docs/en/observability/apm-ui/api.asciidoc
@@ -108,11 +108,13 @@ include::api.asciidoc[tag=using-the-APIs]
 [[apm-update-config]]
 ==== Create or update configuration
 
+[float]
 [[apm-update-config-req]]
 ===== Request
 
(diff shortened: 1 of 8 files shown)


colleenmcginnis:
This is pushing some headings to `h7`, which the /docs repo is set up to handle (they just appear as regular text) because [I don't think `h7` is _technically_ valid HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements). I'm testing some options for making this work with the updated chunk level.

colleenmcginnis (approved):
I pushed a commit to adjust a couple leveloffsets based on what I was seeing in the preview and locally with https://github.com/elastic/docs/pull/3045. I also pushed a commit to https://github.com/elastic/docs/pull/3045 to handle `h7`s.

mdbirnstiehl (approved):

PR #4162, opened 2024-08-15
Update profiling upgrade hint

rockdaboot:
The hint is relevant only for version 8.9 and below, but doesn't mention the version.

### DIFF
--- a/docs/en/observability/profiling-get-started.asciidoc
+++ b/docs/en/observability/profiling-get-started.asciidoc
@@ -81,10 +81,6 @@ NOTE: To configure data ingestion, you need elevated privileges, typically the `
 
 If you're upgrading from a previous version with Universal Profiling enabled, see the <<profiling-upgrade,upgrade guide>>.
 
-IMPORTANT: When upgrading from version 8.9.x or earlier, you must remove all existing profiling data.
-
-We still recommend upgrading, as newer versions contain improvements and new features.
-
 [discrete]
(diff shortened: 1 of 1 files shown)


docs/en/observability/profiling-get-started.asciidoc
```diff
@@ -81,8 +81,8 @@ NOTE: To configure data ingestion, you need elevated privileges, typically the `
 
 If you're upgrading from a previous version with Universal Profiling enabled, see the <<profiling-upgrade,upgrade guide>>.
 
-IMPORTANT: When upgrading, you must remove all existing profiling data.
-We still recommend upgrading as the latest version contains several improvements and new features.
+IMPORTANT: When upgrading from version 8.9.x or below, you must remove all existing profiling data.
```
mdbirnstiehl, line 84:
```suggestion
IMPORTANT: When upgrading from version 8.9.x or earlier, you must remove all existing profiling data.

```

mdbirnstiehl (commented):
one small suggestion.

mdbirnstiehl (approved):

inge4pres:
Note: we have this distinction of pre/post 8.10 already declared here
https://www.elastic.co/guide/en/observability/current/profiling-upgrade.html#profiling-upgrade-process

Should we remove entirely that section from the get started page? Or maybe link to the upgrade page?

mdbirnstiehl (approved):

Issue #3419, opened 2023-12-11
[Request]: Fix APM API key creation command


bmorelli25:
### Description

https://www.elastic.co/guide/en/kibana/current/agent-key-api.html#apm-create-agent-key-privileges is incorrect.

@joeafari was able to successfully create a key with the following privs:

```
{
  "apm_agent_key_user": {
    "cluster": [
      "manage_own_api_key"
    ],
    "indices": [],
    "applications": [
      {
        "application": "kibana-.kibana",
        "privileges": [
          "feature_apm.all"
        ],
        "resources": [
          "*"
        ]
      },
      {
        "application": "apm",
        "privileges": [
          "event:write",
          "config_agent:read"
        ],
        "resources": [
          "*"
        ]
      }
    ],
    "run_as": [],
    "metadata": {},
    "transient_metadata": {
      "enabled": true
    }
  }
}
```

### Resources

https://elastic.slack.com/archives/CJYH0RWR3/p1701780969214649

### Which documentation set does this change impact?

Stateful only

### Feature differences

n/a

### What release is this request related to?

N/A

### Collaboration model

The documentation team

### Point of contact.

None

PR #4165, opened 2024-08-15
[apm] Update agent key privileges to include access to Kibana

colleenmcginnis:
Fixes https://github.com/elastic/observability-docs/issues/3419

Updates the privileges in the [Create agent key](https://www.elastic.co/guide/en/observability/current/apm-create-agent-key.html) section's example role to include access to Kibana. 

cc @joeafari


bmorelli25 (approved):
changes match what I see in the issue and in slack

PR #4169, opened 2024-08-16
Fix broken image links in data quality topic

dedemorton:
Fixes a few broken image links that I noticed while perusing the docs.

I didn't add an image for the degraded document icon because it doesn't exist already in our images or images/icons directory.


mdbirnstiehl (approved):
Thanks, DeDe!

PR #4183, opened 2024-08-21
Add topic about missing fields

### LINKED ISSUE #4091: [Request]: 'Missing Field' explanation for Host documentation
### Description

We're going to be update the 'missing field' UX that shows in lens when a field is missing from the index to be less 'harsh' ('error') and be softer ('information') due to complaints from users that the experience seems 'broken' when it's not:

_'Scary' error message_
image

Our solution will present something much softer like this:

_Softer messages are more appropriate in the context of Hosts_
image

As part of this, we're going to be linking through to Host-specific documentation to explain what missing fields means and why it could happen (via the 'learn more' link).

### Resources

Here is the issue:

- [x] https://github.com/elastic/observability-dev/issues/3089

The documentation should explain the main use cases why fields may be missing:

**1. The chart refers to a field which is not relevant to the hosts they are observing**
This is probably the most common issue. Specifically, if you are only observing windows hosts, 'load' is not collected at all (because 'load' is not a windows concept at all). We should use the load metric not being collected for windows as being a common example of why the fields are missing.

**2. They may not be collecting metrics as we recommend**
This could be because either:

- The system integration is not installed
- They aren't collecting metrics through OpenTelemetry (OTel metrics are not automatically mapped to ECS so these should now work)

_Note : By the time this documentation goes out, the [infrastructure onboarding journey](https://edge-lite-oblt.kb.us-west2.gcp.elastic-cloud.com/app/observabilityOnboarding/?category=infra) will be [updated](https://github.com/elastic/observability-dev/issues/3763) to include the simplest options for onboarding hosts - I'm not sure if you can allude to this or not in the docs...it might become to hard to maintain this_

**3. They have explicitly chosen not to send these metrics**
Some customers choose to limit the metrics emitted to Elastic to save on space/improve cluster performance. For example, the systems integration has options to easily choose which metrics you wish to emit:

image

Stateful and Serverless

### Feature differences

Identical for stateful/serverless

8.16 (but we aim to have this deployed in serverless by 26th August so would like this live for at least serverless by then)

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @roshan-elastic 

**Stakeholders:** @crespocarlos

dedemorton:
Closes #4091 

Preview link: https://observability-docs_bk_4183.docs-preview.app.elstc.co/guide/en/observability/master/handle-no-results-found-message.html

TODO after merging:

- [ ] Port to serverless


dedemorton:
@roshan-elastic This PR is ready for review as soon as the doc build completes. Let me know what you think of the title. I wanted to make it easy for users to map what they see in the UI to the topic title here, but not sure if it works completely. I also wanted to provide some actionable guidance rather than just explaining the situations where this message appears. Let me know what you think!

roshan-elastic:
Hey @dedemorton thanks for writing this up. Here are my thoughts:

>  Let me know what you think of the title

I hear your point on actionable guidance. I think the tricky thing is that action may/may not be needed.

Maybe we could be a bit softer and make it more informative? I like the way you provide action points where it might be more something you want to fix (e.g. you've chosen not to collect it):

**Understanding 'No results found' message**

**Docs hierarchy**

image

It feels like this might be a bit too prevalent and I'm wondering whether it makes sense to nest this under some kind of a 'Troubleshooting' section?

image

Not sure this is the best solution but it feels a bit granular to be at the same level as the other major topics in the nav.

Any thoughts?


_Note : We can easily redirect the UI link to the docs via the short-link we're using (https://ela.st/hosts-missing-fields)_

dedemorton:
>Not sure this is the best solution but it feels a bit granular to be at the same level as the other major topics in the nav.

I totally agree. The topic is a weird outlier in the hierarchy. In general, I hesitate to add a container that only has a single topic, but I think I can get add some useful info to the container topic that makes it worthwhile, so I'll give it a shot.

dedemorton:
@roshan-elastic I've added the troubleshooting section as you requested. I reused some text that we have elsewhere to put some useful detail in the topic. I'm assuming someone must have worked with support on the wording for that page, but I'm not sure. Check out the preview as soon as the build is done and let me know what you think. Thanks!

roshan-elastic (approved):
LGTM! Thanks for doing this!

roshan-elastic:
Oh actually @dedemorton - just checked out the actual [troubleshooting page](https://observability-docs_bk_4183.docs-preview.app.elstc.co/guide/en/observability/master/troubleshooting-infrastructure-monitoring.html).

I think this is OK to go live - I'm wondering if it's worth putting in links to the 'child' docs on the page too (so you don't have to look at the LHN to see the child docs)?

dedemorton:
>I think this is OK to go live - I'm wondering if it's worth putting in links to the 'child' docs on the page too (so you don't have to look at the LHN to see the child docs)?

I think it's a good idea to do this. It's always a little weird to add a hierarchy or bulleted list that has one item. Typically I would put a singleton topic as a section in the main troubleshooting topic until we have more content, but since we are pointing to this page directly from the UI, I think it's better to have a dedicated page.

docs/en/observability/handle-no-results-found-message.asciidoc
```diff
@@ -0,0 +1,37 @@
...
+
+This could be for any of these reasons:
+
+* The integration that collects the missing metrics is not installed.
+For example, to collect metrics from your host system, you can use the {integrations-docs}/system[System integration].
+To fix the problem, install the integration and configure it to send the missing metrics.
++
+TIP: Follow one of our quickstarts under **Observability** → **Add data** → **Collect and analyze logs** to make sure the correct integrations are installed and all required metrics are collected.
+
+* You are not using the Elastic distribution of the OpenTelemetry Collector, which automatically maps data to the Elastic Common Schema (ECS) fields expected by the visualization.
```
mdbirnstiehl, line 25:
```suggestion
* You are not using the Elastic Distribution of the OpenTelemetry Collector, which automatically maps data to the Elastic Common Schema (ECS) fields expected by the visualization.
```

This aligns with the OTel docs and the OTel language distros.

mdbirnstiehl (commented):
One small change for consistency with our OTel docs.

mdbirnstiehl (approved):
🚀

roshan-elastic:
Cheers @dedemorton @mdbirnstiehl,

This is awesome - users now understand why there are missing fields in our UI :)

FYI @teresaalvarezsoler - thought you might like this!

missing fields

PR #4192, opened 2024-08-23
[apm/known_issues] Add known issues for ILM to DSL switch

lahsivjar:
APM-Server switched from Index Lifecycle Management(ILM) to Datastream Lifecycle (DSL) in v8.15.0. This switch was done when we moved from APM integration (which used ILM) to APM-data plugin (which uses DSL) in ES for managing APM datastreams. As a result of the switch, any old datastreams created before the switch would be Unmanaged because the datastream will never be updated with the DSL lifecycle -- this has to be done manually by using the [PUT API](https://www.elastic.co/guide/en/elasticsearch/reference/current/tutorial-manage-existing-data-stream.html).

Related issue: https://github.com/elastic/apm-server/issues/13898

### DIFF
--- a/docs/en/observability/apm/known-issues.asciidoc
+++ b/docs/en/observability/apm/known-issues.asciidoc
@@ -29,19 +29,20 @@ The issue does _not_ occur when creating a _new_ cluster using 8.15.0.
 The issue also does not occur if a custom ILM policy is configured using a custom component template.
 
 // Describe why it happens
-In 8.15.0, APM Server switched to use Datastream Lifecycle(DSL) to manage data retention
+In 8.15.0, APM Server switched to use Data stream Lifecycle to manage data retention
 for APM indices for new deployments as well as for upgraded deployments with default lifecycle
-configurations. Unfortunately, since any datastream created before 8.15.0 does not have DSL
-configuration, such existing datastreams become unmanaged for default lifecycle configurations.
+configurations. Unfortunately, since any data stream created before 8.15.0 does not have data
+stream lifecycle configuration, such existing data streams become unmanaged for default
+lifecycle configurations.
 
 // How to fix it
-Upgrading to 8.15.1 should fix any new indices created for the datastream. However,
+Upgrading to 8.15.1 should fix any new indices created for the data stream. However,
 indices created in version 8.15.0 would remain unmanaged if a custom ILM policy is not
 used. One of the following two approaches can be adopted to fix the unmanaged indices:
 1. Manually delete the indices when they are no longer needed.
-2. Explicitly configure APM datastreams with the default DSL config, note that using
-this approach would migrate all datastreams to DSL which should be equivalent to the
-default ILM policies:
+2. Explicitly configure APM data streams with the default data stream lifecycle config,
+note that using this approach would migrate all data streams to data stream lifecycle
+which should be equivalent to the default ILM policies:
 
 [source,txt]
 ----
--- a/docs/en/observability/apm/known-issues.asciidoc
+++ b/docs/en/observability/apm/known-issues.asciidoc
@@ -19,6 +19,80 @@ If applicable, exact error messages linked to this issue so users searching for 
 Link to fix
 ////
 
+*Upgrading to v8.15.0 may cause APM indices to become unmanaged w.r.t. lifecycle policy* +
+_Elastic Stack versions: 8.15.0_ +
+_Fixed in Elastic Stack version 8.15.1_
+
+// The conditions in which this issue occurs
+The issue occurs when Elastic Stack is upgraded to 8.15.0.
+The issue will NOT occur on a new cluster creation with version 8.15.0.
+The issue occurs because APM Server switched to Elasticsearch APM-data plugin for managing index templates.
+
+The switch also migrated us from Index Lifecycle Management (ILM) to Datastream Lifecycle (DSL) for managing
+data retention for APM indices. Due to this migration, any datastream created before 8.15.0 would be without
+datastream lifecycle configuration and, when migrated to 8.15.0, would become unmanaged (without any data 
+retention configuration). This can be fixed by explicitly configuring APM datastreams with the default
+lifecyle config:
+
+[source,txt]
+----
+
+Thanks for adding more details, we have identified a bug which is causing this issue. In 8.15.0 we migrated from using ILM to DLM (Datastream Lifecycle) and on update, the existing datastreams are not updated with the DLM settings causing Unmanaged indexes.
+
+I have created a bug report in APM-Server and we are working on a fix. For short-term mitigation, we would need to update all the APM datastream manually with the default data retention:
+
+PUT _data_stream/traces-apm-*/_lifecycle
+{
+  "data_retention": "10d"
+}
+
+PUT _data_stream/traces-apm.rum*/_lifecycle
+{
+  "data_retention": "90d"
+}
+
+PUT _data_stream/traces-apm.sampled*/_lifecycle
+{
+  "data_retention": "1h"
+}
+
+PUT _data_stream/metrics-apm.*.1m-*/_lifecycle
+{
+  "data_retention": "90d"
+}
+
+PUT _data_stream/metrics-apm.*.10m-*/_lifecycle
+{
+  "data_retention": "180d"
+}
+
+PUT _data_stream/metrics-apm.*.60m-*/_lifecycle
+{
+  "data_retention": "390d"
+}
+
+PUT _data_stream/metrics-apm.internal-*/_lifecycle
+{
+  "data_retention": "90d"
+}
+
+PUT _data_stream/metrics-apm.app.*/_lifecycle
+{
+  "data_retention": "90d"
+}
+
+PUT _data_stream/logs-apm.*/_lifecycle
+{
+  "data_retention": "10d"
+}
+----
+
+NOTE that while this issue is fixed in 8.15.1, any indices created in 8.15.0 due
+to rollovers would still be unmanaged even after upgrading. The only way to fix
+these is by configuring the datastream lifecycle explicitly as described above.
+Explicitly configuring datastream lifecycle has an added benefit of migrating
+to datastream lifecycle for all APM datastreams.
+
 [[broken-apm-anomaly-rule]]
 *Upgrading to v8.13.0 to v8.13.2 breaks APM anomaly rules* +
 _Elastic Stack versions: 8.13.0, 8.13.1, 8.13.2_ +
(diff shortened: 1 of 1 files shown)


lahsivjar:
[**For reviewers**] This doc update should be valid for both `8.15.1` and `8.16.0`, however, the fix for `8.16.0` will probably diverge and is not in yet. I am not sure if we should only merge this for 8.15 now or for both (given that we MUST fix it for `8.16.0`. CC: @simitt @endorama @inge4pres (8.15 release manager)

simitt:
I would say merge for `8.15.1` now, and for `8.16.0` later when it is actually fixed

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,80 @@ If applicable, exact error messages linked to this issue so users searching for
...
+{
+  "data_retention": "10d"
+}
+----
+
+NOTE that while this issue is fixed in 8.15.1, any indices created in 8.15.0 due
+to rollovers would still be unmanaged even after upgrading. The only way to fix
+these is by configuring the datastream lifecycle explicitly as described above.
+Explicitly configuring datastream lifecycle has an added benefit of migrating
+to datastream lifecycle for all APM datastreams.
```
colleenmcginnis, line 94: Is the linked PR the one that fixed the issue?

```suggestion
While this issue is fixed in 8.15.1 (https://github.com/elastic/elasticsearch/pull/112028[elastic/elasticsearch#112028]), any indices created in 8.15.0 due
to rollovers would still be unmanaged even after upgrading. The only way to fix
these is by configuring the data stream lifecycle explicitly as described above.
Explicitly configuring data stream lifecycle has an added benefit of migrating
to data stream lifecycle for all APM data streams.
```

colleenmcginnis (commented):
I left some suggestions below. Let me know if my suggestions affect accuracy. 

Also, if there are any specific error messages or behaviors that would signal to the user that they are experiencing this issue, please add them here. That can help get users to this info when they search Google etc. for a specific error message or description of a particular behavior they are seeing.

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,80 @@ If applicable, exact error messages linked to this issue so users searching for
 Link to fix
 ////
 
+*Upgrading to v8.15.0 may cause APM indices to become unmanaged w.r.t. lifecycle policy* +
+_Elastic Stack versions: 8.15.0_ +
+_Fixed in Elastic Stack version 8.15.1_
```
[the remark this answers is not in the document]
lahsivjar: It does!

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,80 @@ If applicable, exact error messages linked to this issue so users searching for
...
+// The conditions in which this issue occurs
+The issue occurs when Elastic Stack is upgraded to 8.15.0.
+The issue will NOT occur on a new cluster creation with version 8.15.0.
+The issue occurs because APM Server switched to Elasticsearch APM-data plugin for managing index templates.
+
+The switch also migrated us from Index Lifecycle Management (ILM) to Datastream Lifecycle (DSL) for managing
+data retention for APM indices. Due to this migration, any datastream created before 8.15.0 would be without
+datastream lifecycle configuration and, when migrated to 8.15.0, would become unmanaged (without any data 
+retention configuration). This can be fixed by explicitly configuring APM datastreams with the default
+lifecyle config:
```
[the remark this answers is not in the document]
lahsivjar: Sounds better, thanks for the suggestions!

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,80 @@ If applicable, exact error messages linked to this issue so users searching for
...
+
+[source,txt]
+----
+
+Thanks for adding more details, we have identified a bug which is causing this issue. In 8.15.0 we migrated from using ILM to DLM (Datastream Lifecycle) and on update, the existing datastreams are not updated with the DLM settings causing Unmanaged indexes.
+
+I have created a bug report in APM-Server and we are working on a fix. For short-term mitigation, we would need to update all the APM datastream manually with the default data retention:
+
```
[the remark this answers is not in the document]
lahsivjar: This is a copy-paste mistake 😮‍💨 , thanks for catching it.

docs/en/observability/apm/known-issues.asciidoc
lahsivjar: Yes, the link is correct, thanks for adding that.

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,83 @@ If applicable, exact error messages linked to this issue so users searching for
...
+The issue does _not_ occur when creating a _new_ cluster using 8.15.0.
+The issue also does not occur if a custom ILM policy is configured using a custom component template.
+
+// Describe why it happens
+This happens because APM Server now uses the Elasticsearch apm-data plugin to manage
+index templates, and as a result APM Server now uses datastream lifecycle (DSL) instead
+of index lifecycle management (ILM) to manage data retention for APM indices.
+This means that any datastream created before 8.15.0 does not have DSL configuration
+and, when upgraded to 8.15.0, the datastream becomes unmanaged because it has
+no DSL configuration.
```
simitt, line 37: Is it important to call out the switch to apm-data plugin? I wonder what the user benefit for this level of detail is, and whether we should simply focus on the switch to DSL; something like:

> in 8.15.0, APM Server switched to use datastream lifecycle (DSL) to manage data retention for APM indices for new deployments as well as for upgraded deployments with default lifecycle configurations. Unfortunately, since any datastream created before `8.15.0` does not have DSL configuration, such existing datastreams become unmanaged.

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,83 @@ If applicable, exact error messages linked to this issue so users searching for
...
+index templates, and as a result APM Server now uses datastream lifecycle (DSL) instead
+of index lifecycle management (ILM) to manage data retention for APM indices.
+This means that any datastream created before 8.15.0 does not have DSL configuration
+and, when upgraded to 8.15.0, the datastream becomes unmanaged because it has
+no DSL configuration. 
+
+// How to fix it
+Upgrading to 8.15.1 should fix any new indices created for the datastream. However,
+indices created in version 8.15.0 would remain unmanaged if a custom ILM policy is not
+used. One of the following two approaches can be adopted to fix the unmanaged indices:
```
simitt, line 42: However, indices created in version 8.15.0 would remain unmanaged **_for default lifecycle configurations_**.

simitt (commented):
left some small comments

docs/en/observability/apm/known-issues.asciidoc
lahsivjar: Sounds good! Updated!

simitt (approved):

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,82 @@ If applicable, exact error messages linked to this issue so users searching for
...
+_Elastic Stack versions: 8.15.0_ +
+_Fixed in Elastic Stack version 8.15.1_
+
+// The conditions in which this issue occurs
+The issue only occurs when _upgrading_ the {stack} to 8.15.0.
+The issue does _not_ occur when creating a _new_ cluster using 8.15.0.
+The issue also does not occur if a custom ILM policy is configured using a custom component template.
+
+// Describe why it happens
+In 8.15.0, APM Server switched to use Data stream Lifecycle to manage data retention
```
colleenmcginnis, line 32: Based on the [Elasticsearch docs that cover this topic](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-stream-lifecycle.html), I think all three words should be lowercase: data stream lifecycle.

```suggestion
In 8.15.0, APM Server switched to use data stream lifecycle to manage data retention
```

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,82 @@ If applicable, exact error messages linked to this issue so users searching for
...
+for APM indices for new deployments as well as for upgraded deployments with default lifecycle
+configurations. Unfortunately, since any data stream created before 8.15.0 does not have data
+stream lifecycle configuration, such existing data streams become unmanaged for default
```
colleenmcginnis, line 35:
```suggestion
configurations. Unfortunately, since any data stream created before 8.15.0 does not have a data
stream lifecycle configuration, such existing data streams become unmanaged for default
```

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,82 @@ If applicable, exact error messages linked to this issue so users searching for
...
+lifecycle configurations.
+
+// How to fix it
+Upgrading to 8.15.1 should fix any new indices created for the data stream. However,
+indices created in version 8.15.0 would remain unmanaged if a custom ILM policy is not
+used. One of the following two approaches can be adopted to fix the unmanaged indices:
```
colleenmcginnis, line 41: This feels a bit easier to understand: 

```suggestion
Upgrading to 8.15.1 should fix any new indices created for the data stream. However,
indices created in version 8.15.0 would remain unmanaged if the default ILM policy is
used. One of the following approaches can be adopted to fix the unmanaged indices:
```

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -19,6 +19,82 @@ If applicable, exact error messages linked to this issue so users searching for
...
+1. Manually delete the indices when they are no longer needed.
+2. Explicitly configure APM data streams with the default data stream lifecycle config,
+note that using this approach would migrate all data streams to data stream lifecycle
+which should be equivalent to the default ILM policies:
```
colleenmcginnis, line 45:
```suggestion
2. Explicitly configure APM data streams with the default data stream lifecycle config.
Using this approach would migrate all data streams to use data stream lifecycles,
which should be equivalent to the default ILM policies:
```

colleenmcginnis (dismissed):
I just suggested some minor copy edits below. 🙂

inge4pres:
> I would say merge for `8.15.1` now, and for `8.16.0` later when it is actually fixed

IIANM we cannot do that, but we can merge this as-is for 8.16, backport to 8.15 and then once the fix is completed for 8.16 update again the main branch to amend the docs for 8.16

simitt (approved):

simitt:
>>I would say merge for 8.15.1 now, and for 8.16.0 later when it is actually fixed

>IIANM we cannot do that, but we can merge this as-is for 8.16, backport to 8.15 and then once the fix is completed for 8.16 update again the main branch to amend the docs for 8.16

SGTM!

colleenmcginnis (approved):

simitt:
@colleenmcginnis are we ok with merging this in already and also backport before the `8.15.1` release is out?

colleenmcginnis:
Yep! I can merge now and make sure the backport goes through successfully.

PR #4083, opened 2024-07-18
Fix apm agentcfg api key privileges

carsonip:
Align with feature-roles.asciidoc

PR #4193, opened 2024-08-23
[apm] Update APM feature roles docs

### LINKED ISSUE #3980: [Request]: `apm_writer` user does not have enough permissions
### What documentation page is affected

When setting up APM server, it needs to connect to Elasticsearch. 

Following the principle of least privilege, https://www.elastic.co/guide/en/observability/8.14/apm-privileges-to-publish-events.html guides users in using a dedicated `apm_writer` user for this purpose, which is then specified in `/etc/apm-server/apm-server.yml`.

However, the specified permissions do not suffice. 

On startup, APM server logs:

```
Jun 09 10:47:47 elasticsearch-test.cyberfusion.cloud apm-server[170123]: {"log.level":"error","@timestamp":"2024-06-09T10:47:47.193+0200","log.logger":"beater","log.origin":{"function":"github.com/elastic/apm-server/internal/beater.waitReady","file.name":"beater/waitready.go","file.line":64},"message":"precondition failed: error querying cluster_uuid: status_code=403","service.name":"apm-server","ecs.version":"1.6.0"}
```

This can be fixed by adding the `monitor` cluster privilege. This is definitely a documentation issue. 

However, even then, the following is logged also:

```
Jun 09 10:48:17 elasticsearch-test.cyberfusion.cloud apm-server[170123]: {"log.level":"error","@timestamp":"2024-06-09T10:48:17.201+0200","log.logger":"agentcfg","log.origin":{"function":"github.com/elastic/apm-server/internal/agentcfg.(*ElasticsearchFetcher).Run.func1","file.name":"agentcfg/elasticsearch.go","file.line":150},"message":"refresh cache error: refresh cache elasticsearch returned status 403","service.name":"apm-server","ecs.version":"1.6.0"}
```

This cannot be resolved by adding the `all` cluster and indice privilege. The error does not occur when using the built-in `elastic` user, though, so perhaps that user has a special state? I can't tell if this is a bug, or a documentation issue. 

### What change would you like to see?

See above. 

### Additional info

_No response_

colleenmcginnis:
In https://github.com/elastic/observability-docs/issues/3980#issuecomment-2298759455, @endorama suggested that we clarify the docs on APM feature roles. Let me know what you think of this approach. I attempted to clarify how feature roles work and which privileges are required for different use cases by:

* Reframing the intro to feature roles to emphasize that the what we're recommending is to create several feature-related _roles_ and then assign one or more of these roles to each _user_ (or group) based on which features the user needs to access.
* Providing an example of when a user might need to be assigned multiple roles from this list of feature roles (this needs work).
* Emphasizing that the _Central configuration management role_ is required in most cases (unless central configuration management is explicitly disabled by the org/user).
* Putting all roles on a single page to increase the likelihood that the reader will see the context provided in the intro paragraphs.
* Adding a step at the end of the instructions for creating each role to remind users that they can assign other APM feature roles to the users as needed.

Here's what hasn't been addressed yet:

* [x] Verify that the privileges listed in each role are correct and reflect the latest version of APM / Elastic as a whole. (I'm not sure how to go about confirming or testing these.)
* [x] Are these the _right_ feature roles? Are any missing? Should any be removed?

Here's the PR preview: https://observability-docs_bk_4193.docs-preview.app.elstc.co/guide/en/observability/master/apm-feature-roles.html

@endorama let me know what you think of this approach. If we're on the right track, we can iterate in this PR.

cc @simitt @carsonip


docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -29,23 +29,43 @@ APM Server exposes an HTTP endpoint, and as with anything that opens ports on yo
...
-* <<apm-privileges-api-key,API key role>>: To create and manage API keys.
-* <<apm-privileges-agent-central-config,Central configuration management role>>: To view
-APM Agent central configurations.
-* <<apm-privileges-rum-source-mapping,RUM source mapping role>>: To read RUM source maps.
+Below are some common tasks that users in your organization might need to do and links to more information on creating roles that provides access to the right resources:
+
+* <<apm-privileges-agent-central-config,*Central configuration management role*>>:
+Allows a user to view APM Agent central configurations, which is *required* when
+<<apm-agent-configuration,central configuration management>> is enabled (it is enabled by default).
+* <<apm-privileges-to-publish-events,*Writer role*>>:
```
carsonip, line 51: The writer role is actually the most important role, which allows apm-server to write to ES (although users may configure apm-server to write to something else e.g. logstash). I wonder if we should move it to the top and mark it as required as well.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -29,23 +29,43 @@ APM Server exposes an HTTP endpoint, and as with anything that opens ports on yo
...
+Allows a user to publish events collected by APM Server.
+* <<apm-privileges-to-publish-monitoring,*Monitoring role*>>: Allows a user to publish monitoring data,
+view monitoring data, or both.
+* <<apm-privileges-api-key,*API key role*>>: Allow a user to create and manage API keys.
```
carsonip, line 55: This is deprecated

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -29,23 +29,43 @@ APM Server exposes an HTTP endpoint, and as with anything that opens ports on yo
...
+* <<apm-privileges-rum-source-mapping,*RUM source mapping role*>>: Allows a user to read RUM source maps.
+
+// Is "deployment" the right word?
```
carsonip, line 58: Probably not the most accurate word. Also, I'd just say that if a user uses agent central configuration to manage APM agents, as well as using RUM source mapping, then they need to assign 3 roles to apm-server.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -93,20 +155,22 @@ that has the following privileges:
 |====
```
carsonip, line 155: @endorama to confirm the required privileges to query cluster uuid.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -164,6 +236,8 @@ If you don't use the +apm_system+ user:
...
 +
 . Assign the *monitoring role* to users who need to write monitoring data to {es}.
+. Assign <<apm-feature-roles,additional APM feature roles>> to users as needed
+(including the _Central configuration management role_, which is <<apm-central-config-role-note,required in most cases>>).
```
carsonip, line 240: As mentioned above, most of the time it needs apm_writer as well. I wonder if we could capture that without cluttering this note.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -236,20 +322,20 @@ need to view monitoring data for APM Server:
...
 ////
 
+[float]
 [[apm-privileges-api-key]]
-=== Grant privileges and roles needed for API key management
-
-++++
-<titleabbrev>Create an _API key_ user</titleabbrev>
-++++
+=== Create an _API key_ role
```
carsonip, line 335: @endorama to confirm if this should be mark as deprecated

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -62,18 +82,60 @@ In general, there are three types of privileges you'll work with:
...
+***********************************  ***********************************
+***********************************  ***********************************
+////
 
 [float]
-==== General writer role
+[[apm-privileges-to-publish-events]]
+=== Create a _writer_ role
 
+APM users that publish events to {es} need privileges to write to APM data streams.
```
carsonip, line 138: Add a note that this is not needed if apm-server doesn't write to ES directly, although that is _very_ rare. Also, in those cases, the other configured ES roles will need to be passed in another section e.g. `apm-server.agent.config.elasticsearch` instead of `output.elasticsearch`.

carsonip (commented):
> Verify that the privileges listed in each role are correct and reflect the latest version of APM / Elastic as a whole. (I'm not sure how to go about confirming or testing these.)
Are these the right feature roles? Are any missing? Should any be removed?

Our team should confirm the required privileges. IIUC points to update:
1. missing privilege to query cluster uuid
2. mark api key role as deprecated

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: Do you know in which version it was deprecated?

docs/en/observability/apm/feature-roles.asciidoc
endorama, line 1: I'm somewhat confused reading the page: is it about how clients write to APM Server or how APM Server writes to Elasticsearch? The navigation suggests the latter ("Secure communication with the Elastic Stack") but the enphasys on "users" (plural) makes me wonder (es "It’s recommended that you only grant users the minimum privileges", "Below are some common tasks that users in your organization might need to do").

If the page is about how APM Server writes to Elasticsearch I would replace mentions of "users" with "the APM Server user", so it's clear that is is a "machine user".

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -62,17 +80,20 @@ In general, there are three types of privileges you'll work with:
...
 
-[float]
-==== General writer role
+// Not sure if I captured this accurately...
+[NOTE]
+====
+This is not needed if APM Server doesn't write to {es} directly.
+Instead, another configured {es} role will need to be passed to
+<<apm-agent-config-elasticsearch,`apm-server.agent.config.elasticsearch`>>.
+====
```
endorama, line 96: This doesn't sound right to me, but I may be missing some use case. I have 2 points:
1. APM Server would always write to ES directly :thinking: What use cases would not see APM Server using the Elasticsearch output but not writing directly to Elasticsearch?
2. Documentation for `output.elasticsearch`, which `apm-server.agent.config.elasticsearch` link to, does not mention roles at all, so I think this ends up being too confusing.

Can we clarify (1)? I think (2) will be more easy to explain after clarifying that.

endorama (commented):
Left some comment but overall is already a great improvement :star_struck:

endorama:
I'll take on me to:

- [x] Verify that the privileges listed in each role are correct and reflect the latest version of APM / Elastic as a whole. (I'm not sure how to go about confirming or testing these.)
- [x] Are these the right feature roles? Are any missing? Should any be removed?
- [x] confirm the required privileges to query cluster uuid.
- [x] confirm if _API key_ role should be mark as deprecated (and if yes when deprecation happened)

docs/en/observability/apm/feature-roles.asciidoc
carsonip: > APM Server would always write to ES directly

APM server can be configured to write to logstash, kafka etc, all the outputs supported by libbeat.

endorama:
TL;DR: querying cluster uuid requires `monitor` permission if Elasticsearch security features are enabled or none if not.

Today I researched how to query cluster UUID. My understanding is that `apm-server` queries Elasticsearch on the `/` path to obtain it. From the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-api-root.html)

> f the Elasticsearch security features are enabled, you must have the monitor, manage, or all cluster privilege to use this API.

endorama:
Regarding the Api Key role, my understanding is that is only needed for the `apikey` command in `apm-server`. This command has never been labelled GA, is deprecated since 8.6.0 and [will be removed](https://github.com/elastic/apm-server/issues/14127). For this command to work priviliges `manage_own_api_key` are required. The users should leverage Kibana and Elasticsearch APIs instead.

docs/en/observability/apm/feature-roles.asciidoc
endorama: Yes, it should be. Full context in  https://github.com/elastic/observability-docs/pull/4193#issuecomment-2363155509

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: Closing the loop via https://github.com/elastic/observability-docs/pull/4193#issuecomment-2363155509: 

>deprecated since 8.6.0

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: I attempted to add more context related to (1) in 571a58ca34b28e8339ba532ebc45fbd57d2ceb97, but it's not clear to me what needs to be done related to (2). @endorama @carsonip can you be more specific about what should be changed about the note as it exists in 571a58ca34b28e8339ba532ebc45fbd57d2ceb97?

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: Moving all related discussion to https://github.com/elastic/observability-docs/pull/4193#discussion_r1747383918.

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: How does this look c8fb50d5ae36acc0630b7509b28f3afb687b17c1 ?

endorama:
I've completed checking for roles/permissions.

* 🟩 `source map` role correctly covers using APM source maps in RUM
* 🟩 privileges checks are allowed without authentication in APM use case (checking own permissions)
* 🟨 cluster UUID checks are always performed as part of APM server startup preconditions; this check may require `monitor` permissions if Elasticsearch Security is enabled
* 🟨 if Tail Based Sampling is enabled a license check is performed; this check requires `monitor` permissions
* 🟩 `writer` role correctly covers ingesting data from APM Server to Elasticsearch
* 🟥 `central config` role *should* correctly cover using Central Config, but I was not able to test it. The current docs mention setting `allow_restricted_indices` to `true` but I was not able to understand how to do that from the Roles UI when creating a role. @colleenmcginnis if you have any clarification/input here I'd like to test this one too.
<details><summary>Details</summary>
<p>


- `SEARCH .apm-source-map` [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L913)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L913): **covered** by `source map` role
	- used [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/sourcemap/metadata_fetcher.go?plain=1#L266)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/sourcemap/metadata_fetcher.go?plain=1#L266)
- `HasPrivileges` request [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/auth/apikey.go?plain=1#L138)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/auth/apikey.go?plain=1#L138): does not require privileges
	- [[HasPrivileges API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-has-privileges.html)](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-has-privileges.html)
		- **All** users can use this API, but **only** to determine their own privileges.
		  To check the privileges of other users, you must use the run as feature.  
- `GET /` [here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L986):
	- check is performed as part of every run when checking preconditions [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L642)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L642)
- `GET _license` [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L625)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L625)
	- only requested if tail sampling is enabled [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L619)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/beater.go?plain=1#L619)
	- [[Get license API](https://www.elastic.co/guide/en/elasticsearch/reference/8.15/get-license.html)](https://www.elastic.co/guide/en/elasticsearch/reference/8.15/get-license.html) requires `monitor` permissions
- `Bulk` [[here](https://github.com/elastic/go-docappender/blob/b90c4d9e901e97e75279b3a8ce9a2eb2129d8166/bulk_indexer.go?plain=1#L375)](https://github.com/elastic/go-docappender/blob/b90c4d9e901e97e75279b3a8ce9a2eb2129d8166/bulk_indexer.go?plain=1#L375) (in `go-docappender`):
	- Bulk is performed as part of a `docappender`
		- `BulkIndexerItem`s are added to the configured appedender [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/processors.go?plain=1#L108)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/beater/processors.go?plain=1#L108)
- `SEARCH .apm-agent-configuration` [[here](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/agentcfg/elasticsearch.go?plain=1#L263)](https://github.com/elastic/apm-server/blob/184702854b8af4a992a237ac056ad1f1632d09fd/internal/agentcfg/elasticsearch.go?plain=1#L263): **covered** by `central config` role but I wasn't able to test it because in the UI there seems to be no way to set `allow_restricted_indices=true`


</p>
</details>

colleenmcginnis:
>The current docs mention setting `allow_restricted_indices` to `true` but I was not able to understand how to do that from the Roles UI when creating a role. @colleenmcginnis if you have any clarification/input here I'd like to test this one too.

It looks like that setting was added about 7 months ago by @carsonip in https://github.com/elastic/observability-docs/commit/4834fce9107c9bb55bbf1baa8d23228b1bd93811. Maybe Carson can provide context? (The related [PR](https://github.com/elastic/observability-docs/pull/3607) and [issue](https://github.com/elastic/apm-server/issues/12378) don't provide much information.)

colleenmcginnis:
Also @endorama ...

>cluster UUID checks are always performed as part of APM server startup preconditions; this check may require monitor permissions if Elasticsearch Security is enabled

... can you confirm if this just for the writer role or if it is necessary for all roles? The phrase _APM server startup preconditions_ makes me think it's necessary to do anything, but I wanted to make sure.

endorama:
@colleenmcginnis Sorry for the confusion! `monitor` refers to the cluster privilege. Being it a cluster privilege it's not necessary on all roles but at least 1 needs to have it if:
- security is enabled
- tail based sampling is used

I think adding it to the `writer` role make sense as the checks are performed as part of the usual APM server startup process.

carsonip:
> The current docs mention setting allow_restricted_indices to true but I was not able to understand how to do that from the Roles UI when creating a role

Apparently there is no option to set allow_restricted_indices in roles UI. Roles created in this UI will have allow_restricted_indices default to false. e.g.
```
{
  "agentcfg_1": {
    "cluster": [],
    "indices": [
      {
        "names": [
          ".apm-agent-configuration"
        ],
        "privileges": [
          "read"
        ],
        "field_security": {
          "grant": [
            "*"
          ],
          "except": []
        },
        "allow_restricted_indices": false
      }
    ],
    "applications": [],
    "run_as": [],
    "metadata": {},
    "transient_metadata": {
      "enabled": true
    }
  }
}
```

However, if you create the API key from API Key UI, under "Control security privileges", the example config in the textbox will have `allow_restricted_indices` configurable. The last time I tested this, agentcfg requires allow_restricted_indices to be true. @endorama would be good if you confirm that again.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -9,8 +9,29 @@ NOTE: This documentation only applies to the APM Server binary.
...
+=== Role-based access
+
+{es-security-features} provides {ref}/built-in-roles.html[built-in roles] that grant a
+subset of the privileges needed by users.
+When possible, assign users the built-in roles to minimize the affect of future changes on your security strategy.
+
+If no built-in role is available, you can assign users the privileges needed to accomplish a specific task by
+creating and assigning feature roles to users.
+
+<<apm-feature-roles,*Read more in Use feature roles →*>>
```
colleenmcginnis, line 23: Does this framing make sense?

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -21,60 +42,82 @@ For secure communication between APM Server and APM Agents, see <<apm-secure-age
...
+.Example: Assigning multiple roles to an APM Server user
+*****
+If you use agent central configuration to manage APM agents, you use RUM source mapping,
+and you want to create an APM Server user who can <do something>,
+you would need to assign these three roles to the user:
+
+* <<apm-privileges-to-publish-events,Writer role>>
+* <<apm-privileges-agent-central-config,Central configuration management role>>
+* <<apm-privileges-to-publish-events,RUM source mapping role>>
+*****
```
colleenmcginnis, line 96: What is an example of a task that an APM Server user with these roles might need to complete? Feel free to suggest a different set of roles as an example if this one doesn't make sense.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -21,60 +42,82 @@ For secure communication between APM Server and APM Agents, see <<apm-secure-age
...
-==== General writer role
+// Not sure if I captured this accurately...
+[NOTE]
+====
+This is not needed when APM Server doesn't write to {es} directly.
+For example, in some cases you may configure APM Server to write to another
+output like Logstash, Kafka, or any another output supported by libbeat.
+In these cases, another configured {es} role will need to be passed to
+<<apm-agent-config-elasticsearch,`apm-server.agent.config.elasticsearch`>>.
+====
```
colleenmcginnis, line 118: From @endorama:

>Documentation for `output.elasticsearch`, which `apm-server.agent.config.elasticsearch` link to, does not mention roles at all, so I think this ends up being too confusing.

My question:

>@endorama @carsonip can you be more specific about what should be changed about the note as it exists in the [latest commit]?

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -90,23 +133,88 @@ that has the following privileges:
...
 |Write events into {es}
+
+|Cluster
+|`monitor`
+a|* Allows cluster UUID checks, which are performed as part of APM server startup preconditions
+if {ref}/security-settings.html[Elasticsearch security] is enabled (it is enabled by default).
+* Allows a license check, which is required if <<apm-tail-based-sampling,tail-based sampling>> is enabled.
+
+NOTE: If you have explicitly disabled Elastic security _and_ you are _not_ using tail-based sampling,
+this privilege may not be necessary.
```
colleenmcginnis, line 144: This row is new and needs review.

docs/en/observability/apm/feature-roles.asciidoc
```diff
@@ -90,23 +133,88 @@ that has the following privileges:
...
+To grant an APM Server user with the required privileges for managing central configuration in {es} without {kib},
+assign the user the following privileges:
+
+[options="header"]
+|====
+|Type | Privilege | Purpose
+
+| Index
+|`read` on `.apm-agent-configuration` index, `allow_restricted_indices: true`
+|Allow APM Server to manage central configurations in {es}
```
colleenmcginnis, line 184: @endorama to confirm what @carsonip said in https://github.com/elastic/observability-docs/pull/4193#issuecomment-2374558159. Let me know if this needs to change or if we need to provide more information about how to configure this.

colleenmcginnis (commented):
Thanks for all your help on this @endorama and @carsonip! I think we're getting close now. 😅  I'm going to take this PR out of draft, and I left some specific comments below that still need to be addressed.

endorama:
@carsonip I can confirm it's broken with `allow_restricted_indices: false`. I'll test it with the API key as described.

endorama:
@carsonip Tested and it's working! While testing I also went down a rabbit hole for the API Key format that I documented in https://github.com/elastic/observability-docs/issues/4312 so we can address it separately.

docs/en/observability/apm/feature-roles.asciidoc
endorama: Permissions are correct, but we need to mention that this cannot be achieved with the UI. Linking to https://www.elastic.co/guide/en/observability/current/apm-beats-api-keys.html would provide the appropriate additional information. Incidentally there is an open PR related to that page, as currently does not show the correct information https://github.com/elastic/observability-docs/pull/4083

docs/en/observability/apm/feature-roles.asciidoc
endorama: It does to me. I'm confused about how to use built-in roles though. We provide
`apm_system` and `apm_user` but I was not able to understand which permissions those would apply and if they were relevant compared to what is written under Use feature roles. I'll do a bit of research there.

docs/en/observability/apm/feature-roles.asciidoc
endorama: This looks correct to me.

docs/en/observability/apm/feature-roles.asciidoc
endorama: I would update it as follow:
```suggestion
// Not sure if I captured this accurately...
[NOTE]
====
This is not needed when APM Server doesn't write to {es} directly.
For example, in some cases you may configure APM Server to write to another
output like Logstash, Kafka, or any another output supported by libbeat.
In these cases, different authentication credentials will need to be passed to
<<apm-agent-config-elasticsearch,`apm-server.agent.config.elasticsearch`>>.
====
```

The idea is that you don't pass a role to that configuration but either a username/password pair, an API key or TLS certificates. Removing the mention to roles I think removes the confusion.

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: How does this look?

```suggestion
.Example: Assigning multiple roles to an APM Server user
*****
If you want to create an APM Server user who can use the Elastic APM Real User Monitoring (RUM)
JavaScript Agent to ingest data from a frontend application and you use central configuration
to manage APM agents, you would need to assign these three roles to the user:

* <<apm-privileges-to-publish-events,Writer role>>
* <<apm-privileges-agent-central-config,Central configuration management role>>
* <<apm-privileges-to-publish-events,RUM source mapping role>>
*****
```

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: I understand that we can't configure this role in the UI, but I'm confused about why we would point to the [Grant access using API keys](https://observability-docs_bk_4083.docs-preview.app.elstc.co/guide/en/observability/master/apm-beats-api-keys.html) page. Unless I'm misunderstanding, _Grant access using API keys_ describes an _alternative_ to creating roles, not how to use an API to create/update roles. Is that correct? Are we saying that the user shouldn't use a role to give an APM Server user this kind of access? Or do we want to instruct the user to use an API ([Create or update roles API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role.html)?) to create the _central config_ role?

colleenmcginnis: Would something like this be appropriate? 

>To create a role with the required privileges for managing central configuration in Elasticsearch without Kibana, you must to use the [Roles API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role.html) (the required privileges can't be assigned to a role in Kibana):
>
> ```
>POST /_security/role/apm_agentcfg
>{
>  "description": "Allow APM Server to manage central configurations in Elasticsearch.",
>  "indices": [
>    {
>      "names": [".apm-agent-configuration"],
>      "privileges": ["read"],
>      "allow_restricted_indices": true
>    }
>  ]
>}
>```

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: >Elasticsearch security features provides [built-in roles](https://www.elastic.co/guide/en/elasticsearch/reference/8.15/built-in-roles.html) that grant a subset of the privileges needed by APM users. When possible, assign users the built-in roles to minimize the affect of future changes on your security strategy.

This is in the [current docs](https://www.elastic.co/guide/en/observability/current/apm-secure-comms-stack.html#apm-feature-roles), but if this is no longer relevant we can remove that idea in this PR &mdash; just let me know!

endorama: Yes, I would remove it. I checked with the team also and those roles are not relevant for the APM Server use cases.

colleenmcginnis:
```suggestion
[discrete]
[[apm-secure-comms-stack-role-based]]
=== Role-based access

Manage access on a feature-by-feature basis by creating several custom feature-related roles and assigning
one or more of these roles to each APM Server user based on which features they need to access.

<<apm-feature-roles,*Read more in Use feature roles →*>>
```

docs/en/observability/apm/feature-roles.asciidoc
endorama: Looks good :+1: ! Just one small detail, on line 94 I see `<<apm-privileges-to-publish-events,RUM source mapping role>>`, is `apm-privileges-to-publish-events` the correct link path?

docs/en/observability/apm/feature-roles.asciidoc
endorama: > I understand that we can't configure this role in the UI, but I'm confused about why we would point to the [Grant access using API keys](https://observability-docs_bk_4193.docs-preview.app.elstc.co/guide/en/observability/master/apm-beats-api-keys.html) page. Unless I'm misunderstanding, Grant access using API keys describes an alternative to creating roles, not how to use an API to create/update roles. 

Yes this is correct and yes also to your confusion! 

I think your proposal to use the Roles API is better in this context! I haven't tested it yet but looking at the permissions we grant to the API key and the Roles API docs everything seems correct. I'll test it tomorrow.

docs/en/observability/apm/feature-roles.asciidoc
colleenmcginnis: Good catch!

docs/en/observability/apm/feature-roles.asciidoc
endorama: tested and it's working correctly! The user with that role assigned was correctly able to read the APM agent configurations, while removing the role produced an error. :sparkles:

endorama (dismissed):
🤩 looks great! Thank you!

mdbirnstiehl (commented):
Some little things and some questions about heading hierarchy.

colleenmcginnis:
>Some little things and some questions about heading hierarchy.

Thank you! I messed up the headings when I caught up to `main` today. 🙈

mdbirnstiehl (approved):
🦭 👍

PR #4200, opened 2024-08-26
[APM] Add note about `@custom` component templates

colleenmcginnis:
Fixes https://github.com/elastic/apm-server/issues/11651

👋 @marclop it's not clear in the [issue](https://github.com/elastic/apm-server/issues/11651) if this applies only to `[data_stream][dataset]` or also to `type` and `namespace`, too.


colleenmcginnis:
👋 @marclop can you take a look at this PR sometime this week?

docs/en/observability/apm/configure/outputs/logstash.asciidoc
```diff
@@ -129,7 +129,14 @@ In addition to `@metadata`, APM Server provides other potentially useful fields,
 `data_stream` field, which can be used to conditionally operate on
 {observability-guide}/apm-data-model.html[event types], namespaces, or datasets.
 
-As an example, you might want to use {ls} to route all `metric` events to the same custom metrics data stream,
+// TO DO:
+// It's unclear in the issue whether it's _only_ for datasets or if it's
+// event types, namespaces, and datasets.
```
marclop, line 134: As you aptly inferred, it applies to all the data_stream fields:

- data_stream.dataset
- data_stream.namespace
- data_stream.type

There are more fields that are `constant_keyword` and these may change over time. I think we should instruct customers to review the default templates depending on the routing they wish to use. @axw do you have any input on how to make this the least painful?

mdbirnstiehl (approved):
🐦

docs/en/observability/apm/configure/outputs/logstash.asciidoc
colleenmcginnis: 👋  @axw can you please take a look at @marclop's question above?

axw: Sorry, notification got drowned out in my inbox :(

Is overriding `constant_keyword` to `keyword` something we want to encourage? It works today, but I think it goes against data stream naming. We generally assume that the data stream name matches the fields `{data_stream.type}-{data_stream.dataset}-{data_stream.namespace}`.

Could we alternatively recommend that users rename the `data_stream.dataset` field, and then rely on Logstash's [data stream routing](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data-streams) functionality?

colleenmcginnis: >Could we alternatively recommend that users rename the data_stream.dataset field, and then rely on Logstash's [data stream routing](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data-streams) functionality?

How would the user rename the `data_stream.dataset` field? Would the same process work for all other fields that are `constant_keyword` that @marclop mentioned above?

axw: @colleenmcginnis I should have said "rewrite" rather than "rename". The user would add a [`mutate` filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-mutate.html), which rewrites the `data_stream.dataset` according to whatever logic they choose. I don't have an exact Logstash configuration handy though.

axw: > Would the same process work for all other fields that are constant_keyword that @marclop mentioned above?

Yes, see https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data_stream_auto_routing:

> Automatically routes events by deriving the data stream name using specific event fields with the %{[data_stream][type]}-%{[data_stream][dataset]}-%{[data_stream][namespace]} format.

So the user could mutate any of those fields, and the output would route accordingly.

docs/en/observability/apm/configure/outputs/logstash.asciidoc
```diff
@@ -129,23 +129,40 @@ In addition to `@metadata`, APM Server provides other potentially useful fields,
...
 
-As an example, you might want to use {ls} to route all `metric` events to the same custom metrics data stream,
-rather than to service-specific data streams:
+As an example, you might want to use {ls} to route all `metrics` events to the same custom metrics data stream,
+rather than to service-specific data streams.
+
+However, if when you combine all `metrics` events there are events that have the `data_stream.dataset` field
+set to different values, indexing will fail with a message stating that the field does not accept any other values.
+This is because the `data_stream.dataset` field's mapping is set to `constant_keyword`,
+which expects all values of the fields in the index to be the same.
```
colleenmcginnis, line 138: Is this accurate?

docs/en/observability/apm/configure/outputs/logstash.asciidoc
```diff
@@ -129,23 +129,40 @@ In addition to `@metadata`, APM Server provides other potentially useful fields,
...
+
+To prevent losing data due to failed indexing, you should either:
+
+* Create a `@custom` component template to override the fields' mapping from `constant_keyword` to `keyword`.
+Read more in <<apm-custom-index-template>>.
+* Add a {logstash-ref}/plugins-filters-mutate.html[Logstash mutate filter] to update the value of `data_stream.dataset`.
+This option is shown in the example below.
```
colleenmcginnis, line 145: @marclop do you think we should keep creating a custom component template as an option or only suggest the latter?

docs/en/observability/apm/configure/outputs/logstash.asciidoc
```diff
@@ -154,11 +171,12 @@ output {
 }
 ----
```
colleenmcginnis, line 172: Does this look accurate? Specifically:

* Are the placement _and_ syntax of lines 155-157 correct?
* Is line 158 still necessary or would the correct index be assumed based on [`data_stream_auto_routing`](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data_stream_auto_routing)?

colleenmcginnis (commented):
Thanks for your input @axw @marclop! I attempted to capture your feedback in the latest commit, and I left some specific questions below.

docs/en/observability/apm/configure/outputs/logstash.asciidoc
axw: > Are the placement and syntax of lines 155-157 correct?

No, the `mutate` block needs to be in a separate `filter` block, not under `output` - see below.

> Is line 158 still necessary or would the correct index be assumed based on [data_stream_auto_routing](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-data_stream_auto_routing)?

It's not necessary when configuring `data_stream => "true"` in the Elasticsearch output. I have not tested this, but I think this is correct:

```
filter {
  if [@metadata][beat] == "apm-server" {
    if [data_stream][type] == "metrics" {
      mutate {
        update => { "[data_stream][dataset]" => "custom" }
      }
    }
  }
}
output {
  elasticsearch {
    data_stream => "true"
    cloud_id => "${CLOUD_ID}"
    cloud_auth => "${CLOUD_AUTH}"
  }
}
```

docs/en/observability/apm/configure/outputs/logstash.asciidoc
axw: Yes, this is accurate.

> Is there a specific error/warning message we can include in case any users are searching for information about it?

For example, try running the following two requests:

```
POST /metrics-custom-default/_doc
{
  "@timestamp": "2024-10-08",
  "data_stream": {"dataset": "custom"}
}

POST /metrics-custom-default/_doc
{
  "@timestamp": "2024-10-08",
  "data_stream": {"dataset": "standard"}
}
```

The first one will auto-create the data stream, and the initial value of `data_stream.dataset: custom` will be used for the `constant_keyword` value. Any future values must match this. The second one does not match, and will fail with an error like:

```
{
  "error": {
    "root_cause": [
      {
        "type": "document_parsing_exception",
        "reason": "[1:60] failed to parse field [data_stream.dataset] of type [constant_keyword] in document with id '2z29apIBiisszFNu21Dk'. Preview of field's value: 'standard'"
      }
    ],
    "type": "document_parsing_exception",
    "reason": "[1:60] failed to parse field [data_stream.dataset] of type [constant_keyword] in document with id '2z29apIBiisszFNu21Dk'. Preview of field's value: 'standard'",
    "caused_by": {
      "type": "illegal_argument_exception",
      "reason": "[constant_keyword] field [data_stream.dataset] only accepts values that are equal to the value defined in the mappings [custom], but got [standard]"
    }
  },
  "status": 400
}
```

So the error messages to look for would be:
 - `failed to parse field [data_stream.dataset] of type [constant_keyword]`
 - `[constant_keyword] field [data_stream.dataset] only accepts values that are equal to the value defined in the mappings`

docs/en/observability/apm/configure/outputs/logstash.asciidoc
axw: Let's just keep the latter. Although you _can_ override the type. It's not necessary to change the type when automatically routing based on the `data_stream.dataset` field value.

axw (approved):
Thanks @colleenmcginnis!

mdbirnstiehl (approved):
🦭 👍

PR #4218, opened 2024-09-04
[apm] Add headings to APM known issues page

colleenmcginnis:
Closes #4212 

Adds headings to the APM known issues page to make it easier to read. The heading text is longer than we prefer, but I think it's ok for this page only (@elastic/obs-docs let me know if you disagree).


mdbirnstiehl (approved):
🐦

PR #4113, opened 2024-08-03
Port EEM and new services experience docs to stateful

dedemorton:
Ports #4062 to serverless.  Also makes some minor tweaks to serverless docs to mirror structural changes needed for stateful docs.

Previews:
* [New services experience](https://observability-docs_bk_4113.docs-preview.app.elstc.co/guide/en/observability/master/new-experience-services.html) 
* [Elastic Entity Model](https://observability-docs_bk_4113.docs-preview.app.elstc.co/guide/en/observability/master/elastic-entity-model.html)

TODO:

- [x] Add topics to the correct place in the nav after #3723 is merged.
- [x] Finish porting all changes
- [x] Update link when topic about defining service.name has been ported to stateful (requires https://github.com/elastic/observability-docs/pull/4116)

PR #4236, opened 2024-09-10
Revert "[8.15](backport #4113) Port EEM and new services experience docs to stateful"

dedemorton:
Reverts elastic/observability-docs#4132

It sounds like this was not meant to go out until 8.16.

@mdbirnstiehl Reverting this might affect some of the docs you added..... Let's see what happens with the build and we can decide to go from there.

See related discussion [here](https://github.com/elastic/kibana/issues/192318#issuecomment-2342196141).


mdbirnstiehl (approved):
I think it should be ok as I never linked directly to the new experience, but I did mention it, meaning to link once it went like. Maybe I should remove that mention until 8.16.

dedemorton:
@mdbirnstiehl Might be a good idea to remove it.

PR #4238, opened 2024-09-12
Update docs after migrating from apm integration package to ES apm-data plugin in 8.15

carsonip:
Remove mentions of integration packages as apm-server moved away from it in 8.15


carsonip:
Superseded by #4239

PR #4239, opened 2024-09-12
Update docs after migrating from apm integration package to ES apm-data plugin in 8.15

carsonip:
## Description

Remove mentions of integration packages as apm-server moved away from it in 8.15

_Check all that apply._

- [ ] None of the above

### Related issue
Closes # 

## Checklist

- [x] Product/Engineering Review
- [ ] Writer Review

### Follow-up tasks

_Select one._

### DIFF
--- a/docs/en/observability/apm/data-streams.asciidoc
+++ b/docs/en/observability/apm/data-streams.asciidoc
@@ -109,7 +109,7 @@ For example, consider traces that would originally be indexed to `traces-apm-def
 ]
 ----
 
-For more about other ingest pipelines called by default, e.g. `traces-apm@custom`, see {fleet-guide}/data-streams.html#data-streams-pipelines[integration data streams ingest pipelines].
+For more about other ingest pipelines called by default, e.g. `traces-apm@custom`, see definition of `\*-apm*@default-pipeline` ingest pipelines from {es} apm-data plugin.
 
 For more custom APM ingest pipeline guides, see <<apm-ingest-pipelines,parse data using ingest pipelines>>.
 
--- a/docs/en/observability/apm/data-streams.asciidoc
+++ b/docs/en/observability/apm/data-streams.asciidoc
@@ -109,7 +109,7 @@ For example, consider traces that would originally be indexed to `traces-apm-def
 ]
 ----
 
-To find other ingest pipelines from the {es} apm-data plugin that are called by default, go to *Stack management* → *Ingest pipelines* {kibana-ref}/management.html[[in Kibana] and search for `apm`.
+To find other ingest pipelines from the {es} apm-data plugin that are called by default, go to *Stack management* → *Ingest pipelines* {kibana-ref}/management.html[in Kibana] and search for `apm`.
 Default APM ingest pipelines will follow the pattern `*-apm*@default-pipeline`.
 
 For more custom APM ingest pipeline guides, see <<apm-ingest-pipelines,parse data using ingest pipelines>>.
--- a/docs/en/observability/apm/data-streams.asciidoc
+++ b/docs/en/observability/apm/data-streams.asciidoc
@@ -109,7 +109,8 @@ For example, consider traces that would originally be indexed to `traces-apm-def
 ]
 ----
 
-For more about other ingest pipelines called by default, e.g. `traces-apm@custom`, see definition of `\*-apm*@default-pipeline` ingest pipelines from {es} apm-data plugin.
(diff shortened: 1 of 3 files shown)


docs/en/observability/apm/ingest-pipelines.asciidoc
```diff
@@ -65,6 +65,5 @@ If you prefer more guidance, see one of these tutorials:
 
 * <<apm-filters-ingest-pipeline>> — Learn how to obfuscate passwords stored in the `http.request.body.original` field.
 * <<apm-data-stream-rerouting>> — Learn how to reroute APM data to user-defined APM data streams.
-* {fleet-guide}/data-streams-pipeline-tutorial.html[Transform data with custom ingest pipelines] — A basic Elastic integration tutorial where you learn how to add a custom field to incoming data.
```
carsonip, line 68: This is a shame, as the tutorial is good, but some instructions do not apply to ES plugin

docs/en/observability/apm/data-streams.asciidoc
```diff
@@ -109,7 +109,7 @@ For example, consider traces that would originally be indexed to `traces-apm-def
 ]
 ----
 
-For more about other ingest pipelines called by default, e.g. `traces-apm@custom`, see {fleet-guide}/data-streams.html#data-streams-pipelines[integration data streams ingest pipelines].
+For more about other ingest pipelines called by default, e.g. `traces-apm@custom`, see definition of `\*-apm*@default-pipeline` ingest pipelines from {es} apm-data plugin.
```
colleenmcginnis, line 112: Should this link somewhere? Usually when we say "see {something}" the {something} is a linked resource.

colleenmcginnis (commented):
One small question.

docs/en/observability/apm/data-streams.asciidoc
carsonip: There are no docs around it, not like integration data streams ingest pipelines. What I was trying to convey was that users should just go to ES and check out the relevant pipelines themselves. I don't feel comfortable linking to the [actual source code](https://github.com/elastic/elasticsearch/tree/main/x-pack/plugin/apm-data), as pinning source code version to doc version is not feasible. Any good idea how to rewrite this?

docs/en/observability/apm/ingest-pipelines.asciidoc
colleenmcginnis: Tracking next steps for that tutorial in https://github.com/elastic/observability-docs/issues/4314.

docs/en/observability/apm/data-streams.asciidoc
colleenmcginnis: >go to ES and check out the relevant pipelines themselves

Ok. This makes sense to me. Where would they go or what would they do to find the relevant pipelines? Stack management in Kibana? Using an API? Something else?

carsonip: I'm thinking about stack management, but if a power user likes API, they can do so too.

colleenmcginnis: >I'm thinking about stack management, but if a power user likes API, they can do so too.

We can keep it simple with just Stack management.

What about something like this? Feel free to adjust as needed.

```suggestion
To find other ingest pipelines from the {es} apm-data plugin that are called by default, go to {kibana-ref}/management.html[*Stack management*] → *Ingest pipelines* and search for `apm`.
Default APM ingest pipelines will follow the pattern `*-apm*@default-pipeline`.
```

carsonip: Looks perfect, thanks!

docs/en/observability/apm/data-streams.asciidoc
```diff
@@ -109,7 +109,8 @@ For example, consider traces that would originally be indexed to `traces-apm-def
 ]
 ----
 
-For more about other ingest pipelines called by default, e.g. `traces-apm@custom`, see {fleet-guide}/data-streams.html#data-streams-pipelines[integration data streams ingest pipelines].
+To find other ingest pipelines from the {es} apm-data plugin that are called by default, go to *Stack management* → *Ingest pipelines* {kibana-ref}/management.html[[in Kibana] and search for `apm`.
```
colleenmcginnis, line 112: My mistake! 

```suggestion
To find other ingest pipelines from the {es} apm-data plugin that are called by default, go to *Stack management* → *Ingest pipelines* {kibana-ref}/management.html[in Kibana] and search for `apm`.
```

colleenmcginnis (approved):

mdbirnstiehl (approved):

PR #4242, opened 2024-09-12
[APM] Remove limitation about host metrics in otel docs

colleenmcginnis:
Per @smith @AlexanderWert this limitation should be removed.


colleenmcginnis:
>Does the limitations page direct users to use our otel collector distro?

No. Last I heard we were waiting to integrate information on the Elastic Distribution of OpenTelemetry Collector until it was out of preview. cc @mlunadia

AlexanderWert:
> This is still a limitation if you're sending data collected by the hostmetrics receiver directly to APM server without them being sent through the elasticinframetrics processor in the otel collector.

@smith Since 8.15, the APM Server is also executing parts of the elasticinframetrics processor. So, it should work there, too.

bmorelli25 (approved):

PR #4246, opened 2024-09-12
[Observability docs] Change 'beats' namespace to 'elastic-agent'

kilfoyle:
Elastic Agent Docker images are currently available under both a `beats` and `elastic-agent` namespace. For 9.0 only the latter will be available, so this will update the current (8.15) docs.

Rel: https://github.com/elastic/ingest-docs/issues/1314


jlind23 (approved):

alaudazzi (approved):
LGTM

PR #4388, opened 2024-10-17
fix(new minor automation) change the location for `mergify` and use `8.x` instead of `main`

v1v:
This should avoid https://github.com/elastic/observability-docs/pull/4387 and creating a new minor branch from `main`... it should be `8.x`

@elastic/obs-docs , I ran https://github.com/elastic/observability-docs/actions/runs/11383719000 but I didn't realise the branch off will be from `main` instead of `8.x`.

I need your super-power to delete the `8.16` branch 🙏


bmorelli25 (approved):
🙇

PR #4418, opened 2024-10-22
Document no data settings for custom threshold rule

### LINKED ISSUE #4068: [Request]: Documenting no-data settings for the custom threshold rule
### Description

In the custom threshold rule, we have 2 settings:

1. Alert on no data (`alertOnNoData`)
2. Alert on missing group (`alertOnGroupDisappear`)

At each given time, only one of these settings can be used depending on whether there is a group by field for the rule or not. In this [PR](https://github.com/elastic/kibana/pull/188300), we tried to improve the UX for using these settings by only showing one setting in the UI and changing the related underlying setting depending on whether there is a group by field or not.

Here is a description related to these settings that we would like to include in this document:
In the API, we have 2 settings related to no data:
1. `alertOnNoData`: this is used when we don't have any group by and will trigger a no alert data if the condition doesn’t report any data over the expected time period or if the rule fails to query Elasticsearch. (Essentially, it means something is wrong, and we don't have data to evaluate the related threshold)
2. `alertOnGroupDisappeer`: When we have group by fields and after seeing a group, if that group does not report any data, we trigger a missing group no data alert. 
Here is an example scenario when we have `host.name` as the group by field for CPU usage above 80%:
    -  During the first rule execution, we have 2 hosts that are reporting data: `host-1` and `host-2`
    - In the next rule execution, `host-1` does not report any data, so we trigger a no-data alert for `host-1`
    - In the next execution, if `host-1` starts reporting data again, we will have 2 scenarios:
        - If the `host-1` reports data for CPU usage and it is above the threshold of 80%, then we don't trigger a new alert, but we change the existing alert from no-data to a triggered alert that breaches the threshold. It is important to keep in mind that we don't send any notifications in this case because there is still an ongoing issue.
        

        - If the `host-1` reports CPU usage below the threshold of 80%, then we change the alert status to recovered.

It would be great to also include information about how users can untrack a group that is decommissioned (related [ticket](https://github.com/elastic/kibana/issues/161621)). For example, maybe in the above scenario, `host-1` is decommissioned so we don't need the related alert anymore. In that case, user can select that alert in the alert table and use `mark as untrack` action:

In the UI, we are now using only one setting, and we enable the related API setting according to the group by field. Here, you can see that we also adjust the related tooltip based on whether a group by field is selected or not.

image

### Resources

PR: https://github.com/elastic/kibana/pull/188300

Related issues:
- https://github.com/elastic/kibana/issues/188229
- https://github.com/elastic/kibana/issues/161621

Stateful and Serverless

### Feature differences

This feature is identical in both environments.

8.16

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @maryam-saeidi 

**Stakeholders:**
- @maciejforcone 
- @jasonrhodes 
- @vinaychandrasekhar

dedemorton:
## Description
Adds documentation that explain new setting "no data" setting for the custom threshold rule.

Link to preview: https://observability-docs_bk_4418.docs-preview.app.elstc.co/guide/en/observability/master/custom-threshold-alert.html#trigger-alert-when-no-data

_Check all that apply._

- [ ] None of the above

### Related issue
Closes #4068 

## Checklist

- [ ] Writer Review

### Follow-up tasks

_Select one._

  - [x] Port to serverless docs: ADD ISSUE LINK BEFORE MERGING

### DIFF
--- a/docs/en/observability/threshold-alert.asciidoc
+++ b/docs/en/observability/threshold-alert.asciidoc
@@ -107,7 +107,40 @@ If the `Host A, Architecture A` group matches the rule conditions, but the `Host
 If you select one field—for example, `host.name`—and `Host A` matches the conditions but `Host B` doesn't, one alert is triggered for `Host A`.
 If both groups match the conditions, alerts are triggered for both groups.
 
-When you select *Alert me if a group stops reporting data*, the rule is triggered if a group that previously reported metrics does not report them again over the expected time period.
+[discrete]
+[[trigger-alert-when-no-data]]
+== Trigger "no data" alerts (optional)
+
+Optionally configure the rule to trigger an alert when:
+
+* there is no data, or
+* a group that was previously detected stops reporting data.
+
+To do this, select **Alert me if there's no data**.
+
+The behavior of the alert depends on whether any **group alerts by** fields are specified:
+
+* **No "group alerts by" fields**: (Default) A "no data" alert is triggered if the condition fails to report data over the expected time period, or the rule fails to query {es}. This alert means that something is wrong and there is not enough data to evaluate the related threshold.
+
+* **Has "group alerts by" fields**: If a previously detected group stops reporting data, a "no data" alert is triggered for the missing group.
++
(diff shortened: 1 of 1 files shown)


bmorelli25:
run docs-build

dedemorton:
>Does it make sense also to mention that "no data" has its own action group? Is this already covered somewhere?

That's covered where actions are described, so I don't think I would mention it here.

eedugon (approved):
Looks great to me ;)

PR #4501, opened 2024-11-08
Add reference to prefer-ilm setting for apm lifeycle known bug

lahsivjar:
## Description

Clarify that one of the suggested fix for the unmanaged indices bug introduced in 8.15.0 and fixed in 8.15.1 should only be applied for datatreams with default policies (as they are the only ones affected). If the suggested fix was applied to a datastream using custom ILM policy then the `prefer_ilm` setting needs to be added to the custom ILM policy.

_Check all that apply._

- [ ] None of the above

### Related issue
Closes # 

## Checklist

- [ ] Writer Review

### Follow-up tasks

_Select one._

### DIFF
--- a/docs/en/observability/apm/known-issues.asciidoc
+++ b/docs/en/observability/apm/known-issues.asciidoc
@@ -24,7 +24,7 @@ _Versions: XX.XX.XX, YY.YY.YY, ZZ.ZZ.ZZ_
 [discrete]
 == Upgrading to v8.15.x may cause ingestion to fail
 
-_Elastic Stack versions: 8.15.0+_
+_Elastic Stack versions: 8.15.0, 8.15.1, 8.15.2, 8.15.3_ +
 _Fixed in Elastic Stack version 8.15.4_
     
 // The conditions in which this issue occurs
--- a/docs/en/observability/apm/known-issues.asciidoc
+++ b/docs/en/observability/apm/known-issues.asciidoc
@@ -93,7 +93,7 @@ To fix these unmanaged indices, consider one of the following approaches:
 . Explicitly configure APM data streams to use the default data stream lifecycle configuration.
 This approach migrates all affected data streams to use data stream lifecycles,
 maintaining behavior equivalent to the default ILM policies.
-Apply this fix only to data streams that are currently unmanaged due to default ILM policies.
+Apply this fix only to data streams that have unmanaged indices due to missing default ILM policies.
 +
 [source,txt]
 ----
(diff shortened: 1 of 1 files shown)


docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -91,7 +91,10 @@ used. One of the following approaches can be adopted to fix the unmanaged indice
 . Manually delete the indices when they are no longer needed.
 . Explicitly configure APM data streams with the default data stream lifecycle config.
 Using this approach would migrate all data streams to use data stream lifecycles,
-which should be equivalent to the default ILM policies:
+which should be equivalent to the default ILM policies. Note that if you have a custom
+lifecycle policy in place for any of the datastreams then you don't need to do anything
+for those datastreams as they are not affected by this bug. Going forward, for using
+custom ILM policies with new datastreams, please consult the {ref}/apm-ilm-how-to.html[updated guide].
```
simitt, line 97: Why do you think this belongs here? If I have a custom ILM policy I don't even read this line due to 
>The issue also does not occur if a custom ILM policy is configured using a custom component template.

lahsivjar: Took a stab at updating this in a different way to avoid confusion. PTAL.

simitt: This part looks good to me now - are you planning on also adding the `prefer-ilm` section to this PR or a separate one? (The PR description is not updated)

bmorelli25:
Added two commits from an open PR (https://github.com/elastic/observability-docs/pull/4506) that also makes updates to this file.

bmorelli25:
I made a few small changes to the text. Please take a look and let me know if this still looks okay. If so, we'll get this merged. Thanks!

docs/en/observability/apm/known-issues.asciidoc
```diff
@@ -84,63 +85,26 @@ stream lifecycle configuration, such existing data streams become unmanaged for
...
-which should be equivalent to the default ILM policies:
+Upgrading to 8.15.1 resolves the lifecycle issue for any new indices created for APM data streams.
+However, indices created in version 8.15.0 will remain unmanaged if the default ILM policy is in place.
+To fix these unmanaged indices, consider one of the following approaches:
+
+. Manually delete the unmanaged indices when they are no longer needed.
+. Explicitly configure APM data streams to use the default data stream lifecycle configuration.
+This approach migrates all affected data streams to use data stream lifecycles,
+maintaining behavior equivalent to the default ILM policies.
+Apply this fix only to data streams that are currently unmanaged due to default ILM policies.
```
lahsivjar, line 96:
```suggestion
Apply this fix only to data streams that have unmanaged indices due to missing default ILM policies.
```

The datastream won't be unmanaged currently so a slight change of wordings.

lahsivjar:
Thanks @bmorelli25 , a minor comment - rest LGTM!

docs/en/observability/apm/known-issues.asciidoc
bmorelli25: Ah! Good catch.

bmorelli25 (approved):

colleenmcginnis (approved):
🦭

PR #4502, opened 2024-11-08
Update search connectors links

leemthompo:
There was a couple of stragglers here, they aren't broken links but they are soft redirects as of 8.16.0, so best to replace them. :)


eedugon (approved):
LGTM! thanks!

PR #4391, opened 2024-10-17
ci: fix branch creation and mergify file

v1v:
I fixed it partially, but missed that I need to set the branches correctly after. moving to 8.x

PR #4554, opened 2024-11-20
fix: force checkout from main

v1v:
For some reason, `git checkout main` didn't run and created a PR with the wrong base branch, see https://github.com/elastic/observability-docs/pull/4553

https://github.com/elastic/observability-docs/pull/4391/files#diff-5712d072e70274b06faa283d7da2be2aeb554b214de3ba70c148664bd9c523b5R71 is the explicit checkout but it was not honoured :/


colleenmcginnis (approved):
✨

PR #4599, opened 2024-11-27
Clean up APM getting started intro page

### LINKED ISSUE #4336: [Request]: Review APM getting started landing page
### Description

In https://github.com/elastic/observability-docs/pull/4230, @bmorelli25 brought up a few issues with the [APM getting started landing page](https://www.elastic.co/guide/en/observability/current/apm-getting-started-apm-server.html) where we compare/contrast the two approaches: [APM Server binary](https://www.elastic.co/guide/en/observability/current/apm-getting-started-apm-server.html#apm-setup-apm-server-binary) and [Fleet-managed APM Server](https://www.elastic.co/guide/en/observability/current/apm-getting-started-apm-server.html#apm-setup-fleet-managed-apm). We should review this whole page to make sure it's still accurate.

Specific feedback from https://github.com/elastic/observability-docs/pull/4230:

* At the top of the page we say _The easiest way to get started with Elastic APM is by using our hosted Elasticsearch Service on Elastic Cloud_, but in the APM Server binary's list of pros we say _Simplest self-managed option_.
  
  From @bmorelli25 via [comment](https://github.com/elastic/observability-docs/pull/4230#discussion_r1752867997):
  > At the top of the page we link to the fleet server method and say that's the easiest way to get started. Here we say the opposite.

* We have a note that says _Fleet-managed APM Server does not have full feature parity with the APM Server binary method of running Elastic APM._

  From @bmorelli25 via [comment](https://github.com/elastic/observability-docs/pull/4230#discussion_r1752875441):
  >Is this still the case? In looking through Slack, @simitt recently mentioned that other than fewer outputs, there is general feature parity. This might be stale.

* In the list of required components for the Fleet-managed option, we refer to APM Server, Elastic Agent, and Fleet Server individually.

  From @bmorelli25 via [comment](https://github.com/elastic/observability-docs/pull/4230#discussion_r1752877780):
  >I don't think this is fair. Elastic agent gives you Fleet Server and APM Server.

### Resources

Initial comments in https://github.com/elastic/observability-docs/pull/4230

Stateful only

### Feature differences

Stateful only

N/A

### Collaboration model

The documentation team

### Point of contact.

**Main contact:** @colleenmcginnis 

**Stakeholders:** @bmorelli25 @simitt

colleenmcginnis:
## Description

Cleans up the [APM getting started intro page](https://observability-docs_bk_4599.docs-preview.app.elstc.co/guide/en/observability/master/apm-getting-started-apm-server.html) as discussed in #4336 including:

* **Which option is the "easiest"**: @simitt @bmorelli25 what do you think about just removing the note at the top and letting users determine which is the easiest method for their situation? If we can't say that one is easier than the other without explaining the context, then I'm not sure it makes sense to have a note at the top. 
* **What's preventing feature parity**: I moved this note to the part of the table that discusses supported outputs.
*  **Required components**: I rephrased the Elastic Agent list item (I was going to use a footnote but I was having trouble using it inside a list that's inside a table).
* I also deleted a big block of commented out content since we can get it from the git history if needed, but let me know if you think we should keep it. 

_Check all that apply._

- [ ] None of the above

### Related issue
Closes #4336

## Checklist

- [x] Product/Engineering Review
- [ ] Writer Review

### Follow-up tasks

_Select one._

  - [x] The concepts in this PR only apply to one doc set (serverless _or_ stateful)

### DIFF
--- a/docs/en/observability/apm/getting-started-apm/index.asciidoc
+++ b/docs/en/observability/apm/getting-started-apm/index.asciidoc
@@ -5,14 +5,7 @@
 <titleabbrev>Get started</titleabbrev>
 ++++
 
-****
-The easiest way to get started with Elastic APM is by using our
-{ess-product}[hosted {es} Service] on {ecloud}.
-The {es} Service is available on AWS, GCP, and Azure.
-*To get started in minutes, follow the steps in <<get-started-with-fleet-apm-server>>.*
-****
-
-IMPORTANT: Starting in version 8.15.0, the {es} apm-data plugin manages APM index templates,
+NOTE: Starting in version 8.15.0, the {es} apm-data plugin manages APM index templates,
 lifecycle policies, and ingest pipelines.
 
 The APM Server receives performance data from your APM agents,
@@ -32,8 +25,6 @@ There are two options, and the components required are different for each:
 Fleet is a web-based UI in {kib} that is used to centrally manage {agent}s.
 In this deployment model, use {agent} to spin up APM Server instances that can be centrally-managed in a custom-curated user interface.
 
-NOTE: Fleet-managed APM Server does not have full feature parity with the APM Server binary method of running Elastic APM.
-
 image::./images/fm-ov.png[APM Server fleet overview]
 
 // (outputs, stable APIs)
@@ -49,12 +40,11 @@ integrations from one central {fleet} UI.
 a| * {es}
 * {ess}
 
+NOTE: Fleet-managed APM Server does support all the outputs that are supported by the APM Server binary method of running Elastic APM.
+
 | *Required{nbsp}components*
 a| * APM agents
-* APM Server
-* {agent}
-* Fleet Server
-* {stack}
+* {agent} (which runs multiple subprocesses including APM Server, Fleet Server, and {stack})
 
 | *Configuration method*
 | {kib} UI
@@ -92,49 +82,6 @@ a| * APM agents
 | *Configuration method*
 a| YAML
 |===
-
-// [float]
-// [[apm-setup-apm-server-ea]]
-// === Standalone Elastic Agent-managed APM Server
-// // I really don't know how to sell this option
-// Instead of installing and configuring the APM Server binary, let {agent} orchestrate it for you.
-// Install {agent} and manually configure the agent locally on the system where it's installed.
-// You are responsible for managing and upgrading {agent}. This approach is recommended for advanced users only.
-
-// **Pros**:
-
-// - Easily add integrations for other data sources
-// useful if EA already in place for other integrations, and customers want to customize setup rather than using Fleet for configuration
-// // TODO:
-// // maybe get some more hints on this one from the EA team to align with highlighting the same pros & cons.
-
-// **Available on Elastic Cloud**: ❌
-
-// This supports all of the same outputs as binary
-// see https://github.com/elastic/apm-server/issues/10467
-// **Supported outputs**:
-
-// **Configuration method**: YAML
-
-// image::./images/ea-ov.png[APM Server ea overview]
-
-// @simitt's notes for how to include EA-managed in the decision tree:
-// ****
-// If we generally describe Standalone Elastic Agent managed APM Server then we should also add it to this diagram:
-// Do you want to use other integrations?
-// -> yes: Would you like to use the comfort of Fleet UI based management? -> yes: Fleet managed APM Server; -> no: Standalone Elastic Agent managed APM Server
-// -> no: What is your prefered way of configuration? -> yaml: APM Server binary; -> Kibana UI: Fleet managed APM Server
-// ****
-
-// Components required:
-
-// [options="header"]
-// |====
-// | Installation method                         | APM Server | Elastic Agent | Fleet Server
-// | APM Server binary                           | ✔️          |               |
-// // | Standalone Elastic Agent-managed APM Server | ✔️          | ✔️             |
-// | Fleet-managed APM Server                    | ✔️          | ✔️             | ✔️
-// |====
 
 [float]
 == Help me decide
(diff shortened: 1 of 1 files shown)


bmorelli25 (approved):
I'm good with these changes. The [Get Started page](https://observability-docs_bk_4599.docs-preview.app.elstc.co/guide/en/observability/master/apm-getting-started-apm-server.html) is now a quick overview, a brief pro/con list of each option, and the flowchart to help people decide. I think there's enough context on the page for users to decide which option is easiest for their use case.

eedugon:
Great doc!! I have a question about this box:
image

`Do you want to use other integrations` sounds too generic when it probably wants to ask `Do you want to use other integrations together with the APM Server`?

eedugon:
I have a couple of extra questions, just in case you consider them useful, feel free to ignore them if wanted ;)

image

What does it mean `centrally` and `on the edge` in the context of the question? I haven't seen any explanation about that in the rest of the doc. It feels like `on the edge` means `where the apps are running, far away from Elastic Stack`, and `centrally` might mean then `closer to the elastic stack`, but I feel a bit lost with both answers.

image

Is it mandatory to run the APM integration in the SAME Agent that acts as Fleet server? If so I think we should clearly state this, because a Fleet user with existing Fleet Servers might think that the APM Server integration could run directly in a normal Elastic Agent managed by fleet without being also a Fleet Server.

That's also not described / explained in the [Fleet-managed APM Server](https://observability-docs_bk_4599.docs-preview.app.elstc.co/guide/en/observability/master/get-started-with-fleet-apm-server.html) guide.

About the `Step 1` of that guide: https://observability-docs_bk_4599.docs-preview.app.elstc.co/guide/en/observability/master/get-started-with-fleet-apm-server.html#_step_1_set_up_fleet (Setup Fleet)

Shouldn't we rely and refer to the Fleet documentation for the Fleet Setup (if it's not already done)? Our content could get outdated as the Fleet Server setup is part of the ingest / fleet documentation and there are multiple `Deployment modes` available.

Of course this question is related to the previous one about if a `fleet server together with the apm server` is mandatory in the fleet managed mode.

In my opinion, and if the Fleet Server in the same agent as the APM Server is **not mandatory**, we should state that in the docs, also explaining that:
- If the user doesn't have a Fleet Setup then the suggestion is to colocate together the Fleet Server and the APM Server (as it's quicker and simpler).
- If the user already has a Fleet Setup (they could even have multiple Fleet Servers in their environment), they can do any of:
  - Add an agent with the `APM Server` to act as APM Server
  - Add the `APM Server` integration to any of their existing fleet servers so they would act as Fleet Servers + APM Servers.

Sorry for the noise!

docs/en/observability/apm/getting-started-apm/index.asciidoc
```diff
@@ -49,12 +40,11 @@ integrations from one central {fleet} UI.
 a| * {es}
 * {ess}
 
+NOTE: Fleet-managed APM Server does support all the outputs that are supported by the APM Server binary method of running Elastic APM.
```
simitt, line 43: This is not true - Fleet managed APM Server does **only** support Elasticsearch output. 

>The missing support for outputs different than Elasticsearch is the only thing that comes to mind ad-hoc, but I also haven't spent much time on this recently. We should still be calling out that we only support ES as output when running under Fleet.

simitt (changes requested):

simitt:
>What does it mean centrally and on the edge in the context of the question? I haven't seen any explanation about that in the rest of the doc. It feels like on the edge means where the apps are running, far away from Elastic Stack, and centrally might mean then closer to the elastic stack, but I feel a bit lost with both answers.

IMO the terms `central` and `on the edge` are fairly standard terms, and yes they explain that for RUM one needs a central apm server, as the requests are coming from real users browsers, where no APM Server is running. I don't think we should go into this level of detail in such an overview, but rather in the RUM docs. 

>Is it mandatory to run the APM integration in the SAME Agent that acts as Fleet server?
No. It is just the simplest setup with Fleet/EA, as the users then only have to run one Elastic Agent. 

>About the Step 1 of that guide: https://observability-docs_bk_4599.docs-preview.app.elstc.co/guide/en/observability/master/get-started-with-fleet-apm-server.html#_step_1_set_up_fleet (Setup Fleet):
>Shouldn't we rely and refer to the Fleet documentation for the Fleet Setup (if it's not already done)? Our content could get outdated as the Fleet Server setup is part of the ingest / fleet documentation and there are multiple Deployment models available.

No concerns with this, if detailed Fleet docs are available.

eedugon:
Ok, I see that `centrally` and `on the edge machines` concepts are kind of explained here: https://www.elastic.co/guide/en/observability/current/apm-upgrade-to-apm-integration.html#apm-arch-upgrade

In my opinion the `on the edge` on the screenshots we are sharing is not intuitive at all, but it could be me.

I think **on the edge** actually implies `on the applications edge machines without direct access from web browsers` (that's why RUM doesn't support that architecture), and centrally means somewhere with direct access from all apps and web browsers.

But to be honest not sure how to improve it :(

docs/en/observability/apm/getting-started-apm/index.asciidoc
colleenmcginnis:
```suggestion
NOTE: Fleet-managed APM Server does _not_ support all the outputs that are supported by the APM Server binary method of running Elastic APM.
```

colleenmcginnis:
@eedugon do you mind if we move your comments to a separate issue? I think updating the diagram and the other subpages in the getting started section are out of scope for this PR, which aims to address the issues listed in https://github.com/elastic/observability-docs/issues/4336.

colleenmcginnis:
Moved @eedugon's comments to https://github.com/elastic/observability-docs/issues/4668 and https://github.com/elastic/observability-docs/issues/4669. Hiding them in this PR.

simitt (approved):

bmorelli25 (approved):

Issue #4229, opened 2024-09-06
[Request]: [SLOs] Update SLO privilege model docs


dominiqueclarke:
### What documentation page is affected

https://www.elastic.co/guide/en/observability/current/slo-privileges.html

### What change would you like to see?

The SLO privilege model was updated for 8.16.0.

In particular, there is no longer a requirement for SLO editor users to have cluster privileges `manage_transform` and `manage_ingest_pipelines`. 

Additionally, the correct index privileges were added to the Observability `editor` role. Therefore, users can now choose to either use the `editor` role or create a custom role to grant index privileges to SLO users.

The required index privileges are as follows

Editor user
```
.slo-observability-*: `read`, `view_index_metadata`, `write`, `manage`
```
Viewer user
```
.slo-observability-*: `read`, `view_index_metadata`

### Additional info

_No response_

eedugon:
Proposed changes to the doc (already working on them):

- Change the names `SLO All` and `SLO Read` for the roles to become `SLO Editor` and `SLO Viewer`.
- Update the permissions needed for both types of roles to match the new values.
- Update screenshots.

Add the information about the `editor` **built-in role** being another option to grant SLO Editor permissions to the users.

Any thoughts?

eedugon:
PR draft started here: https://github.com/elastic/observability-docs/pull/4300

eedugon:
@dominiqueclarke , would we need to update anything in the `serverless` documentation related with this issue?

Based on the doc https://www.elastic.co/docs/current/serverless/observability/create-an-slo it's mentioned that `The Admin role or higher is required to create SLOs.`, so I would say we don't need to change anything at Serverless documentation side.

By the way, is the previous statement accurate for serverless? Shouldn't `Editor` be enough based on the explanations from https://www.elastic.co/docs/current/serverless/general/assign-user-roles#observability ? (sorry for the offtopic).

dominiqueclarke:
@eedugon 

> is the previous statement accurate for serverless? Shouldn't Editor be enough based on the explanations from

You are correct. We should update the serverless docs to mention `The Editor role or higher is required to create SLOs`, rather than `Admin`.

eedugon:
thanks @dominiqueclarke : I have updated also the serverless doc in the mentioned PR

PR #4643, opened 2024-12-09
Update slo-privileges.asciidoc

AkashPrakashShinde:

_Check all that apply._

- [ ] Stateful (`docs/en/observability/*`)
- [ ] None of the above

### Related issue
Closes # 

## Checklist

- [ ] Writer Review

### Follow-up tasks

_Select one._


docs/en/observability/slo-privileges.asciidoc
```diff
@@ -35,11 +35,12 @@ To create a role:
 Set the following privileges for the SLO Editor role:
 
 . Under *Index privileges* in the *Elasticsearch* section, add `.slo-observability-*` to the *Indices* field and `read`, `view_index_metadata`, `write`, and `manage` to the *Privileges* field.
+. Under *Cluster privileges* add `manage_transform` as SLO most likely to create the transform respect to your SLO.
```
colleenmcginnis, line 38: I'm not sure what you're trying to say here. Is there an issue or something else that explains why this privilege is needed?

docs/en/observability/images/SLO_permissions.png
colleenmcginnis, line 1: The text in this image is small and pretty difficult to read and there's a lot of whitespace. Can you reduce the width of your browser window before taking a screenshot? You can refer to the [existing image](https://www.elastic.co/guide/en/observability/current/slo-privileges.html#slo-all-access) for guidance on an appropriate width.

docs/en/observability/slo-privileges.asciidoc
```diff
@@ -35,11 +35,12 @@ To create a role:
...
 . Click *Add index privilege*.
 . In the *Indices* field, add the indices for which you plan to create SLOs. Then, add `read` and `view_index_metadata` to the *Privileges* field. The following example shows `logs-*`, but you can specify any indices.
 +
 [role="screenshot"]
-image::images/slo-es-priv-editor.png[Cluster and index privileges for SLO Editor role]
```
colleenmcginnis, line 42: If this image is no longer needed, you should delete it. Alternatively you could rename your new image `slo-es-priv-editor.png` so it replaces the existing image.

docs/en/observability/slo-privileges.asciidoc
eedugon: The reason of needing the privilege is that SLOs create transforms under the hood, so I would suggest something like:

> Under *Cluster privileges* add `manage_transform` as SLOs rely on Elasticsearch transforms.

Or just the need without any explanation (we are not offering explanations in the other privileges required):

> Under *Cluster privileges* add `manage_transform`.

We have recently added a new doc section called `Understanding SLO internals` with an introduction explaining the relation between SLOs and transforms. We could link to that document to provide background (https://www.elastic.co/guide/en/observability/current/slo-troubleshoot-slos.html#slo-understanding-slos).

eedugon:
After a conversation with @AkashPrakashShinde he has also realized that `manage_pipeline` privilege is also needed. Let's check this out with devs before moving forward.

@dominiqueclarke , if you remember we updated the privileges needed for SLOs a few months ago under this issue: https://github.com/elastic/observability-docs/issues/4229

It looks we might have missed cluster privileges `manage_transform` and `manage_pipeline` for the SLOs to work. This has been detected by @AkashPrakashShinde while working in support.

Would you please take a look and confirm if the privileges are needed? I'm pretty sure the `manage_transform` cluster privilege is needed, but @AkashPrakashShinde mentions something about `manage_pipeline` being optional and probably needed too (which makes sense considering SLOs create transforms and ingest pipelines under the hood).

As soon as you confirm we will fix the docs through this PR.

eedugon:
@dominiqueclarke : @AkashPrakashShinde also found another issue in the docs.
It looks for `index privileges` we are suggesting in the docs `.slo-observability-*` while SLOs code and the logic is checking for `.slo-observability.*` (notice the difference).

Without fixing the privileges the user gets this type of error when accessing SLOs view:
Screenshot 2024-12-11 at 09 22 13

In summary, we need to clarify if we have to fix these items in the doc:
- Change `.slo-observability-*` by `.slo-observability.*`.
- Add `manage_transform` cluster privilege as needed.
- Add `manage_pipeline` cluster privilege as needed.

If you prefer we can create an issue instead of discussing the problem in this PR.

docs/en/observability/slo-privileges.asciidoc
dominiqueclarke: Users should not need `manage_transform` privileges in 8.16. We removed this requirement in an earlier version. Are you receiving reports of SLOs not working without `manage_transforms`? IF so, what version?

PR #4691, opened 2024-12-13
apm-ilm: update apm-ilm-how-to expand dsl migration note

1pkg:
## Description

This PR updates `apm-ilm-how-to` page to expand DSL to ILM migration note. 

_Check all that apply._

- [ ] None of the above

### Related issue
Closes # 

Needs changelog updates https://github.com/elastic/apm-server/pull/14909 to be merged first.  

## Checklist

- [ ] Writer Review

### Follow-up tasks

_Select one._

  - [x] The concepts in this PR only apply to one doc set (serverless _or_ stateful)

### DIFF
--- a/docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
+++ b/docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
@@ -22,7 +22,8 @@ Indices created in 8.15.x and 8.16.x might be managed by {ref}/data-stream-lifec
 More details are available in <<apm-release-notes-8.15>>.
 By default, new indices created in 8.17 are managed by ILM.
 More details are available in <<apm-release-notes-8.17>>.
-If you have indices managed by DSL note that your custom DSL settings will not automatically apply to the new indices managed by ILM.
+If you have indices managed by DSL, any custom DSL settings that you specified will _not_ automatically apply to the new indices managed by ILM.
+Instead, you can replicate custom DSL settings in ILM using this guide.
 ====
 
 [discrete]
--- a/docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
+++ b/docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
@@ -18,9 +18,9 @@ customization for lifecycle.
 
 [NOTE]
 ====
-The indices that were created in 8.15.x and 8.16.x might be managed by {ref}/data-stream-lifecycle.html[data stream lifecycle (DSL)].
+Indices created in 8.15.x and 8.16.x might be managed by {ref}/data-stream-lifecycle.html[data stream lifecycle (DSL)].
 More details are available in <<apm-release-notes-8.15>>.
-In 8.17 by default the new indices will be managed by ILM again.
+By default, new indices created in 8.17 are managed by ILM.
 More details are available in <<apm-release-notes-8.17>>.
 If you have indices managed by DSL note that your custom DSL settings will not automatically apply to the new indices managed by ILM.
 ====
--- a/docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
+++ b/docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
@@ -18,8 +18,11 @@ customization for lifecycle.
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
```diff
@@ -18,8 +18,11 @@ customization for lifecycle.
 
 [NOTE]
 ====
-Some older indices might be managed by {ref}/data-stream-lifecycle.html[data stream lifecycle (DSL)] if a cluster is upgraded to 8.17.
+The indices that were created in 8.15.x and 8.16.x might be managed by {ref}/data-stream-lifecycle.html[data stream lifecycle (DSL)].
+More details are available in <<apm-release-notes-8.15>>.
+In 8.17 by default the new indices will be managed by ILM again.
```
colleenmcginnis, line 23: I removed "again" because we shouldn't assume that someone using 8.17 also used Elastic APM before 8.15 (so this might be the first time their indices are managed by ILM).

```suggestion
Indices created in 8.15.x and 8.16.x might be managed by {ref}/data-stream-lifecycle.html[data stream lifecycle (DSL)].
More details are available in <<apm-release-notes-8.15>>.
By default, new indices created in 8.17 are managed by ILM.
```

docs/en/observability/apm/manage-storage/ilm-how-to.asciidoc
```diff
@@ -18,8 +18,11 @@ customization for lifecycle.
...
 More details are available in <<apm-release-notes-8.17>>.
+If you have indices managed by DSL note that your custom DSL settings will not automatically apply to the new indices managed by ILM.
```
colleenmcginnis, line 25: >your custom DSL settings

I'm confused about this. I thought that customization was always done using ILM policies. At least that's what we wrote for 8.15:

>As of 8.15, new clusters use [data stream lifecycle (DSL)](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/data-stream-lifecycle.html) to provide default data retention settings, but the customization for lifecycle is still performed using ILM policies, which are now configured to override the default DSL polices.
>
>_From https://www.elastic.co/guide/en/observability/8.16/apm-ilm-how-to.html_

If a user did use custom settings in 8.15/8.16 but those settings aren't automatically applied to 8.17 indices is there something they can do to manually apply their custom settings in 8.17?

1pkg: @colleenmcginnis some level of customization is applicable to DSL see this [reference](https://elastic.co/guide/en/elasticsearch/reference/current/tutorial-manage-data-stream-retention.html#retention-configuration). If a user applied this customization in 8.15/8.16 it won't directly automatically transfer to ILM  in 8.17 thus this note and also a mention in 8.17.0 release notes.

colleenmcginnis: I see. Is there something they can do to manually apply the equivalent of any DSL custom settings in ILM in 8.17?

1pkg: To clarify even more DSL customization is less common and is not a part of UI. That's why we never advertised it in the APM docs. Nevertheless, it still possible that it was applied by users therefore we want a clarification note on this subject.

1pkg: Users will still be able to replicate the same DSL customization in ILM in 8.17 manually yes. Following our guides in https://www.elastic.co/guide/en/observability/current/apm-ilm-how-to.html and https://www.elastic.co/guide/en/elasticsearch/reference/current/example-using-index-lifecycle-policy.html#ilm-ex-modify-policy

colleenmcginnis: Does this make sense? 

```suggestion
If you have indices managed by DSL, any custom DSL settings that you specified will _not_ automatically apply to the new indices managed by ILM.
Instead, you can replicate custom DSL settings in ILM using this guide.
```

1pkg: I think it looks great, thank you for refining it!

colleenmcginnis (approved):
✨

PR #4768, opened 2025-01-14
Test docs-freeze.yml

bmorelli25:
## Description

Test docs freeze action


eedugon:
@bmorelli25 : I think the file added in this PR will be needed now that we are in the frozen period, right?

Shouldn't we merge this so it starts affecting all new PRs or have we already added it in another PR? In such case we should probably close this.

PR #4810, opened 2025-01-30
mergify: support 8.18 backports

v1v:
does what it says in the tin

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -352,3 +352,16 @@ pull_request_rules:
         branches:
           - "8.17"
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
+  - name: backport patches to 8.18 branch
+    conditions:
+      - merged
+      - label=backport-8.18
+    actions:
+      backport:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -352,3 +352,29 @@ pull_request_rules:
...
+      - label=backport-9.0
+    actions:
+      backport:
+        assignees:
+          - "{{ author }}"
+        labels:
+          - "backport"
+        branches:
+          - "9.0"
+        title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
```
v1v, line 380:
```suggestion
```

lahsivjar (approved):
Thanks!

bmorelli25 (approved):
ty!

lahsivjar:
@v1v should this be backported to 8.x branch too?

v1v:
> Looks like it is not needed as I can see backports to 8.18 being created from 8.x branch
Changes in the mergify file should be on in the default branch,.Mergify does not need

PR #4887, opened 2025-04-22
New way of backporting

ev1yehor:
## Proposed commit message

New way of backporting to active branches using new github action.

## Why

No need to support the hardcoded branches in the `.mergify.yml` file anymore.

We use a dynamic approach by fetching https://storage.googleapis.com/artifacts-api/snapshots/branches.json and filtering the branches.

## Issues

Already enabled in Elastic Agent:
- https://github.com/elastic/elastic-agent/actions/workflows/backport-active.yml

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,3 +1,17 @@
+commands_restrictions:
+  backport:
+    conditions:
+      - or:
+        - sender-permission>=write
+        - sender=github-actions[bot]
+defaults:
+  actions:
+    backport:
+      title: "[{{ destination_branch }}] (backport #{{ number }}) {{ title }}"
+      assignees:
+        - "{{ author }}"
+      labels:
+        - "backport"
 pull_request_rules:
   - name: ask to resolve conflict
     conditions:
--- a/.github/workflows/backport-active.yml
+++ b/.github/workflows/backport-active.yml
@@ -1,0 +1,22 @@
+name: Backport to active branches
+
+on:
+  pull_request_target:
+    types: [closed]
+    branches:
+      - main
+
+permissions:
+  pull-requests: write
+  contents: read
+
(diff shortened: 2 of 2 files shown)


karenzone (approved):
Cool workflow, especially now that we're actively supporting multiple branches. 
If I'm reading this correctly, we're adding the ability to use 'backport-active-', and yet still retaining the option to backport only to specific branches.  If that's the case, then LGTM unless somebody else thinks of an edge case that I haven't considered.

v1v (approved):

.mergify.yml
```diff
@@ -1,3 +1,17 @@
+commands_restrictions:
+  backport:
+    conditions:
+      - or:
+        - sender-permission>=write
+        - sender=github-actions[bot]
+defaults:
+  actions:
+    backport:
+      title: "[{{ destination_branch }}] (backport #{{ number }}) {{ title }}"
```
v1v, line 10:
```suggestion
      title: "[{{ destination_branch }}] {{ title }} (backport #{{ number }})"
```

v1v (approved):

ev1yehor:
@karenzone You are right. We can use backport-active-* labels to backport changes to active branches while you can still use old labels to specific branches. And If you want to backport changes to specific branch in future, you can always use `@mergify backport $branch` command.

ev1yehor:
@karenzone
 ` "backport-active-all" is the label that automatically backports to all 8.* and 9.* active branches.
 "backport-active-8" is the label that automatically backports to all active minor branches for the 8 major.
 "backport-active-9" is the label that automatically backports to all active minor branches for the 9 major.`
 List of active branches taken from json: https://storage.googleapis.com/artifacts-api/snapshots/branches.json

PR #4926, opened 2025-08-06
[Observability Docs] Replace placeholder URLs

mdbirnstiehl:
Contributes to https://github.com/elastic/docs-content/issues/1806 and updates example links to avoid security threat.

### DIFF
--- a/docs/en/observability/apm/shared-ssl-logstash-config.asciidoc
+++ b/docs/en/observability/apm/shared-ssl-logstash-config.asciidoc
@@ -100,16 +100,16 @@ The following example uses the IP address rather than the hostname to validate t
 
 [source,shell]
 ------------------------------------------------------------------------------
-curl -v --cacert ca.crt <YOUR_IP_ADDRESS>:5044
+curl -v --cacert ca.crt https://192.168.99.100:5044
 ------------------------------------------------------------------------------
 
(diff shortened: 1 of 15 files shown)


bmorelli25 (approved):

docs/en/observability/apm/shared-ssl-logstash-config.asciidoc
```diff
@@ -73,43 +73,43 @@ Before running APM Server, you should validate the {ls} server's certificate. Yo
...
+* Connection #0 to host <YOUR_URL> left intact
 curl: (52) Empty reply from server
 ------------------------------------------------------------------------------
 
 The following example uses the IP address rather than the hostname to validate the certificate:
 
 [source,shell]
 ------------------------------------------------------------------------------
-curl -v --cacert ca.crt https://192.168.99.100:5044
+curl -v --cacert ca.crt <YOUR_IP_ADDRESS>:5044
```
florent-leborgne: For IP addresses we have https://github.com/elastic/docs-content-internal/issues/152 that will get us covered. No need to use placeholders for these examples, we can instead use IP addresses designed for docs examples

florent-leborgne (approved):
Thanks for updating those; see my note about IP addresses, but that is fine in the meantime

PR #4961, opened 2025-10-28
[Internal][Obs UX Management][8.x]: Note filter limitation for metric threshold rules

nastasha-solomon:

_Check all that apply._

- [ ] None of the above

### Related issue
Contributes to https://github.com/elastic/docs-content/issues/3701

## Checklist

- [ ] Writer Review

### Follow-up tasks

_Select one._

  - [x] Port to stateful docs: https://github.com/elastic/docs-content/pull/3702
  - [x] Port to serverless docs: https://github.com/elastic/docs-content/pull/3702

### DIFF
--- a/docs/en/observability/metrics-threshold-alert.asciidoc
+++ b/docs/en/observability/metrics-threshold-alert.asciidoc
@@ -45,7 +45,10 @@ image::images/metrics-alert-filters-and-group.png[Metric threshold filter and gr
 
 The *Filters* control the scope of the rule. If used, the rule will only evaluate metric data that matches the query in this field. In this example, the rule will only alert on metrics reported from a Cloud region called `us-east`.
 
-NOTE: Filters that you've added to the rule using the (https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id)[create rule API] won't display in the UI when you're editing a rule. If you want to modify them, you must manually re-add the filters by entering a KQL query in the rule's **Filter** field.
+NOTE: If you've made a rule with the (https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id)[create rule API] and added Query DSL filters using the `filterQuery` parameter, the filters won't appear in the UI for editing a rule. As a workaround, manually re-add the filters through the UI and save the rule. As you're modifying the rule's filters from the UI, be mindful of the following:
+
+ - The **Filter** field only accepts KQL syntax, meaning you may need to manually convert your Query DSL filters to KQL. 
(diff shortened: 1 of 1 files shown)


docs/en/observability/metrics-threshold-alert.asciidoc
```diff
@@ -45,6 +45,8 @@ image::images/metrics-alert-filters-and-group.png[Metric threshold filter and gr
 
 The *Filters* control the scope of the rule. If used, the rule will only evaluate metric data that matches the query in this field. In this example, the rule will only alert on metrics reported from a Cloud region called `us-east`.
 
+NOTE: Filters that you've added to the rule using the (https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id)[create rule API] won't display in the UI when you're editing a rule. If you want to modify them, you must manually re-add the filters by entering a KQL query in the rule's **Filter** field.
```
nastasha-solomon:
```suggestion
If you've made a rule with the (https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id)[create rule API] and added Query DSL filters using the `filterQuery` parameter, the filters won't appear in the UI. To modify the filters from the UI, first convert the Query DSL filters to KQL, then manually re-add them to rule's **Filter** field.
```

justinkambic: ++ - it's also possible to work around this issue by editing the rule in the UI and saving KQL there. The KQL the user adds will get converted to a query AST that will get populated into the `filterQuery` field as well, so edits via the UI will override the filter originally included when the create API request was made.

If you are able to distill that rambling explanation into something useful to append to what we already have here, that might fully clarify the issue/workaround for users.

justinkambic (approved):
I had one small suggestion to help further clarify the UI workaround for the user, but it might not be essential. I am also fine with this as it is.

docs/en/observability/metrics-threshold-alert.asciidoc
nastasha-solomon: Let me know if this works or needs more changes:

```suggestion
NOTE: If you've made a rule with the (https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id)[create rule API] and added Query DSL filters using the `filterQuery` parameter, the filters won't appear in the UI for editing a rule. As a workaround, manually re-add the filters through the UI and save the rule. As you're modifying the rule's filters from the UI, be mindful of the following:

 - The **Filter** field only accepts KQL syntax, meaning you may need to manually convert your Query DSL filters to KQL. 
 - After you save the rule, filters you've added to the **Filter** field are converted appropriately and specified in the rule's `filterQuery` parameter.  
```

mdbirnstiehl (approved):

